PR #2, opened 2022-04-20
Fix qm9 discrete energies

### LINKED ISSUE #1: Discrete pIC50 predictions for QM9-pretrained SchNetBind model
When using the QM9-pretrained SchNetBind model, the output predictions seem to become discrete after any additional training. The figures below show predicted values vs target values after 1 epoch of training and after 1000 epochs.
1_epoch 1000_epoch
With no additional training, the predictions are continuous as expected.
1000_epoch

kaminow:
## Description
Remove atomref embedding from QM9-pretrained SchNetBind models.
Fixes #1

### DIFF
--- a/covid_moonshot_ml/train_dd.py
+++ b/covid_moonshot_ml/train_dd.py
@@ -77,7 +77,7 @@ def build_model_e3nn(n_atom_types, num_neighbors, num_nodes, node_attr=False,
         model = Network(**model_kwargs)
     return(model)
 
-def build_model_schnet(qm9=None, dg=False, qm9_target=10):
+def build_model_schnet(qm9=None, dg=False, qm9_target=10, remove_atomref=False):
     ## Load pretrained model if requested, otherwise create a new SchNet
     if qm9 is None:
         if dg:
@@ -90,15 +90,25 @@ def build_model_schnet(qm9=None, dg=False, qm9_target=10):
         # target=10 is free energy (eV)
         model_qm9, _ = SchNet.from_qm9_pretrained(qm9, qm9_dataset, qm9_target)
 
-        if dg:
-            model = SchNetBind(model_qm9.hidden_channels, model_qm9.num_filters,
-                model_qm9.num_interactions, model_qm9.num_gaussians,
-                model_qm9.cutoff, model_qm9.max_num_neighbors, model_qm9.readout,
-                model_qm9.dipole, model_qm9.mean, model_qm9.std,
-                model_qm9.atomref.weight.detach().clone())
-            model.load_state_dict(model_qm9.state_dict())
+        if remove_atomref:
+            atomref = None
+            ## Get rid of entries in state_dict that correspond to atomref
+            wts = {k: v for k,v in model_qm9.state_dict().items() \
+                if 'atomref' not in k}
         else:
-            model = model_qm9
+            atomref = model_qm9.atomref.weight.detach().clone()
+            wts = model_qm9.state_dict()
+
+        model_params = (model_qm9.hidden_channels, model_qm9.num_filters,
+            model_qm9.num_interactions, model_qm9.num_gaussians,
+            model_qm9.cutoff, model_qm9.max_num_neighbors,model_qm9.readout,
+            model_qm9.dipole, model_qm9.mean, model_qm9.std, atomref)
+
+        if dg:
+            model = SchNetBind(*model_params)
+        else:
+            model = SchNet(*model_params)
+        model.load_state_dict(wts)
 
     ## Set interatomic cutoff to 3.5A (default of 10) to make the graph smaller
     model.cutoff = 3.5
@@ -132,6 +142,8 @@ def get_args():
         help='Whether to treat the ligand and protein atoms separately.')
     parser.add_argument('-dg', action='store_true',
         help='Whether to predict pIC50 directly or via dG prediction.')
+    parser.add_argument('-rm_atomref', action='store_true',
+        help='Remove atomref embedding in QM9 pretrained SchNet.')
 
     ## Training arguments
     parser.add_argument('-n_epochs', type=int, default=1000,
@@ -202,7 +214,8 @@ def init(args, rank=False):
         for k,v in ds_train[0][1].items():
             print(k, v.shape, flush=True)
     elif args.model == 'schnet':
-        model = build_model_schnet(args.qm9, args.dg, args.qm9_target)
+        model = build_model_schnet(args.qm9, args.dg, args.qm9_target,
+            args.rm_atomref)
         if args.dg:
             model_call = lambda model, d: model(d['z'], d['pos'], d['lig'])
         else:
--- a/covid_moonshot_ml/pred_comp.py
+++ b/covid_moonshot_ml/pred_comp.py
@@ -35,6 +35,8 @@ def get_args():
     parser.add_argument('-wts', help='Model weights file.')
     parser.add_argument('-model_params', help='e3nn model parameters.')
     parser.add_argument('-qm9', help='QM9 directory for pretrained model.')
+    parser.add_argument('-qm9_target', type=int, default=10,
+        help='QM9 pretrained target.')
 
     ## Output arguments
     parser.add_argument('-o', required=True, help='Output file basename.')
@@ -46,6 +48,8 @@ def get_args():
         help='Whether to treat the ligand and protein atoms separately.')
     parser.add_argument('-dg', action='store_true',
         help='Whether to predict pIC50 directly or via dG prediction.')
+    parser.add_argument('-rm_atomref', action='store_true',
+        help='Remove atomref embedding in QM9 pretrained SchNet.')
 
     ## Eval arguments
     parser.add_argument('-n', type=int, default=12,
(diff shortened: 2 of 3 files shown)


yuanqing-wang:
can we figure out first whether this is an architecture problem or a parameter problem? if you do 
```python
( layer.reset_parameter() for layer in model.childern() ) 
```

to make all parameters white noise, it should give you noisy predictions but continuous ones nonetheless

kaminow:
@yuanqing-wang resetting all parameters gives the following plots (with and without the `atomref` encoding)
with_atomref no_atomref
they are indeed continuous

yuanqing-wang:
Hmm could you `print(model)` and paste its architecture here?

kaminow:
The SchNet class overloads `__repr__` so it's not super helpful: `SchNetBind(hidden_channels=128, num_filters=128, num_interactions=6, num_gaussians=50, cutoff=3.5)`. However, I realized that the SchNet class includes its own `reset_parameters` function, and after using that I get the following results:
pred_comp_schnet_qm9_rm_atomref_noise
pred_comp_schnet_qm9_noise
continuous without `atomref`, but discrete with it

yuanqing-wang:
Ah OK I see to understand it now: the `atomref` should be some crazy huge number. so basically you would be counting the number of atoms of various elements if you turn that on and train on data small in energy. could you also make sure that the units are consistent? like what's the unit for the energy you insert as training data

kaminow:
`atomref` is calibrated to each individual target in the QM9 dataset, so the units should be the same as the prediction (meV in this case). also, because QM9 only contains molecules with HCNOF, those are the only non-zero entries in `atomref` (values are `[  -13.9030, -1030.2589, -1485.7117, -2043.0182, -2713.8879]` respectively). what's confusing me is that the `atomref` encoding is the same for both forward passes through the `SchNet` model for a given pass through the `SchNetBind` model, so their contributions should cancel out and not contribute to the error at all.

jeherr:
SchNet's own reset parameter function resets the atomrefs to the initial atomrefs so it's not surprising that it would give the same discrete behavior as before.

https://github.com/pyg-team/pytorch_geometric/blob/6a2b04741a0d511e8599bd648291e659179da5d9/torch_geometric/nn/models/schnet.py#L140

jeherr:
[Am I mistaken or is this line missing a negative sign?](https://github.com/choderalab/covid-moonshot-ml/blob/a2097a9f68626eddb8df4cbb68d5a4ac7a731e0a/covid_moonshot_ml/SchNetBind.py#L23)

yuanqing-wang:
@jeherr I think `\Delta G` here is association whereas `K_d` is dissociation, implying another negative sign that canceled the one in the equation.

kaminow:
> @jeherr I think `\Delta G` here is association whereas `K_d` is dissociation, implying another negative sign that canceled the one in the equation.

this is correct. the equation is `dG = -kt*ln(K_eq)`, and `K_d = 1/K_eq`

kaminow:
> SchNet's own reset parameter function resets the atomrefs to the initial atomrefs so it's not surprising that it would give the same discrete behavior as before.
> 
> https://github.com/pyg-team/pytorch_geometric/blob/6a2b04741a0d511e8599bd648291e659179da5d9/torch_geometric/nn/models/schnet.py#L140

this is a good point. I re-ran with calling the `SchNet` `model.reset_parameters()` and then manually calling `model.atomref.reset_parameters()`, which gives the following (continuous) results
pred_comp_schnet_qm9_noise

covid_moonshot_ml/train_dd.py
```diff
@@ -90,15 +90,25 @@ def build_model_schnet(qm9=None, dg=False, qm9_target=10):
...
-                model_qm9.atomref.weight.detach().clone())
-            model.load_state_dict(model_qm9.state_dict())
+            model = SchNetBind(*model_params)
         else:
-            model = model_qm9
+            model = SchNet(*model_params)
+        model.load_state_dict(wts)
 
     ## Set interatomic cutoff to 3.5A (default of 10) to make the graph smaller
     model.cutoff = 3.5
```
jchodera, line 114: Should we be experimenting with larger cutoffs too? Perhaps this could be set by the CLI?

jchodera (commented):
Thanks for tagging me in!

A few comments here:
* It looks like this is getting closer to the predicted scale of values we expect, but they're still small and centered around 0, which is wrong. I wonder if we're missing a standard state correction, which would account for the fact that the virtual concentration of the ligand should be 1 M. This would be some kind of additive offset, though---we should presumably be able to simply learn that.
* The SchNet `model.cutoff` is hard-coded to 3.5 A. We should probably make this an option we experiment with---perhaps something we could serialize along with the model so we don't have to hard-code or hand-set it in the training and inference scripts?
* We might want to move the models into a `models/` subpackage, where we can call the files (modules) `schnet` and `e3nn` so you can `from .models.e3nn import E3NNBind` and `from .models.schnet import SchNetBind`. These are just stylistic choices, but will make things cleaner.
* The training scripts should become [entry points](https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts). These allow you to organize the scripts that do things instead into (1) a Python API you can use from another program, (2) a method you can call with command-line arguments, and (3) an entry point defined in `setup.py` that builds an executable script on the path that calls (2) with the command-line arguments when you run it. @yuanqing-wang, @ijpulidos, or @mikemhenry can help you with these if you haven't used them before---they're super useful!
* Can you produce training curves for these models? We should examine the plot of training and validation error every iteration, ensure training error continues to go down and validation error goes down and then climbs way back up again, and then pick the model with the lowest validation error to evaluate on the out-of-sample test set. Otherwise, it's hard to know if we are training for an apporpriate amount of time or selecting an appropriate model.
* We should also bootstrap over the number of molecules in the test set to produce 95% CIs for various statistics---or better yet, perhaps we could just use [arsenic](https://github.com/openforcefield/arsenic) to implement best practices plots and statistics for us. @ijpulidos can help you add calls to this.
* The conversion in [these lines](https://github.com/choderalab/covid-moonshot-ml/blob/6557b9994e2a942e524ca88ae1d83e3eac8d4b82/covid_moonshot_ml/SchNetBind.py#L22-L29) looks right to me, but since it's easy to mess up units, I wonder if we want to make the scaling factor out front learnable instead of something we try to compute without error.

jchodera (commented):
Thanks for tagging me in!

A few comments here:
* It looks like this is getting closer to the predicted scale of values we expect, but they're still small and centered around 0, which is wrong. I wonder if we're missing a standard state correction, which would account for the fact that the virtual concentration of the ligand should be 1 M. This would be some kind of additive offset, though---we should presumably be able to simply learn that.
* The SchNet `model.cutoff` is hard-coded to 3.5 A. We should probably make this an option we experiment with---perhaps something we could serialize along with the model so we don't have to hard-code or hand-set it in the training and inference scripts?
* We might want to move the models into a `models/` subpackage, where we can call the files (modules) `schnet` and `e3nn` so you can `from .models.e3nn import E3NNBind` and `from .models.schnet import SchNetBind`. These are just stylistic choices, but will make things cleaner.
* The training scripts should become [entry points](https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts). These allow you to organize the scripts that do things instead into (1) a Python API you can use from another program, (2) a method you can call with command-line arguments, and (3) an entry point defined in `setup.py` that builds an executable script on the path that calls (2) with the command-line arguments when you run it. @yuanqing-wang, @ijpulidos, or @mikemhenry can help you with these if you haven't used them before---they're super useful!
* Can you produce training curves for these models? We should examine the plot of training and validation error every iteration, ensure training error continues to go down and validation error goes down and then climbs way back up again, and then pick the model with the lowest validation error to evaluate on the out-of-sample test set. Otherwise, it's hard to know if we are training for an apporpriate amount of time or selecting an appropriate model.
* We should also bootstrap over the number of molecules in the test set to produce 95% CIs for various statistics---or better yet, perhaps we could just use [arsenic](https://github.com/openforcefield/arsenic) to implement best practices plots and statistics for us. @ijpulidos can help you add calls to this.
* The conversion in [these lines](https://github.com/choderalab/covid-moonshot-ml/blob/6557b9994e2a942e524ca88ae1d83e3eac8d4b82/covid_moonshot_ml/SchNetBind.py#L22-L29) looks right to me, but since it's easy to mess up units, I wonder if we want to make the scaling factor out front learnable instead of something we try to compute without error.

mikemhenry:
RE:

> The training scripts should become [entry points](https://packaging.python.org/en/latest/specifications/entry-points/#use-for-scripts). These allow you to organize the scripts that do things instead into (1) a Python API you can use from another program, (2) a method you can call with command-line arguments, and (3) an entry point defined in setup.py that builds an executable script on the path that calls (2) with the command-line arguments when you run it. @yuanqing-wang, @ijpulidos, or @mikemhenry can help you with these if you haven't used them before---they're super useful!

Yes these are super useful and I would be happy to help! This blog as a nice example on how they work, and goes into more advance things like plugin detection, so you can stop before the "Snek for Everyone" example https://amir.rachum.com/blog/2017/07/28/python-entry-points/

jchodera:
> @jchodera the plots in this PR are all with random parameters. the trained model (with atomref removed) shows stronger agreement with experimental values:

This is great! It looks like these are overfit now, so it will be important to monitor the validation loss and pick a model using early stopping.

jchodera:
@kaminow : I just realized this branch hasn't been merged yet! Is it outdated and should be closed? Or should it be updated and merged?

kaminow:
thanks for the ping, this should definitely be updated and merged to main! I'll take care of it today

PR #11, opened 2022-07-27
Add pIC50 uncertainties in schema objects

kaminow:
Updated the CDD parsing functions to calculate and store the variance in the pIC50 measurement.


covid_moonshot_ml/datasets/utils.py
```diff
@@ -130,13 +131,32 @@ def cdd_to_schema(cdd_csv, out_json, achiral=False):
...
+    pic50_var = []
+    for _, (ci_lower, ci_upper) in df[[ci_lower_key, ci_upper_key]].iterrows():
+        if pandas.isna(ci_lower) or pandas.isna(ci_upper):
+            pic50_var.append(np.nan)
+        else:
+            ## First convert bounds from IC50 (uM) to pIC50
+            pic50_ci_upper = -np.log10(ci_upper*10e-6)
+            pic50_ci_lower = -np.log10(ci_lower*10e-6)
+            ## Assume size of 95% CI == 4*sigma => calculate variance from stdev
+            pic50_var.append(((pic50_ci_lower - pic50_ci_upper) / 4) ** 2)
```
jchodera, line 149: Suggest storing the `stderr` (standard error) is better than the `var` (variance).

jchodera (changes requested):
Apologies for not seeing this earlier!
Suggest changing variance -> standard error, but otherwise this looks great!

jchodera:
Also, I'm not sure if you're storing the semiquantitative data yet, but if you are, then you can set the standard error to a default value (e.g. 0.5 pIC50 units) if there is no confidence interval defined.

kaminow:
> Also, I'm not sure if you're storing the semiquantitative data yet, but if you are, then you can set the standard error to a default value (e.g. 0.5 pIC50 units) if there is no confidence interval defined.

Should this be handled when parsing the data or should the value be left up to the user at train/test time? Currently I'm storing NaNs where there is no CI info, but I can change that to 0.5 or whatever value we want. I am storing semiquantitative data: https://github.com/choderalab/covid-moonshot-ml/blob/add-uncertainties/covid_moonshot_ml/datasets/utils.py#L216-L217
so this can be used as well

jchodera:
> Should this be handled when parsing the data or should the value be left up to the user at train/test time?
> Currently I'm storing NaNs where there is no CI info, but I can change that to 0.5 or whatever value we want.

Good question!

For preparing CSV files for redistribution, it would be best to pre-populate the uncertainty with a sensible value that reflects an appropriate uncertainty for the limits here, which may be tighter than 0.5 pIC50 units.

For use in your ML models, it may makes the most sense to enable you to interpret the NaNs and experiment with different uncertainties for the semiquantiative data.

kaminow:
@jchodera I've updated the code to now fill the standard error for any compounds outside the dynamic range of the fluorescence assay with the mean of the compounds that _are_ within the dynamic range. The compounds still keep track of whether the pIC50 value was above, within, or below the dynamic range, so it's trivial to update the default value.

jchodera:
That sounds like a decent approach! It looks like you just need to rebase or merge in `main` and then we can merge!

jchodera (approved):

jchodera (approved):
Looks good, thanks!

ijpulidos:
Closing this PR in order to avoid unintended side effects when removing the data from git and github history.

PR #16, opened 2022-08-10
Moved imports to be more local.

### LINKED ISSUE #12: Make imports local
In order to make it easier to reuse modules, it's helpful to make imports as local as possible.
If we `import` a package just before we need to use it within a method, it becomes possible to import a module from this repo even if some of the methods require packages that are not yet installed.

kaminow:
Move imports to be more local, to resolve #12.


kaminow:
@jchodera does this look right?

jchodera (approved):
This is great! Thanks!

PR #22, opened 2022-08-22
Add ligand rmsd code

apayne97:
## Description
Combine the RMSD and analysis stuff I've written for looking at the results into 'hallucinate-fragalysis' before merging it with main

## Status
- [ ] Ready to go


covid_moonshot_ml/data/openeye.py
```diff
@@ -58,4 +59,24 @@ def split_openeye_mol(complex_mol: oechem.OEMolBase):
...
+
+    ## this gets the coordinates into a numpy array
+    ref_xyz = np.array(list(ref.GetCoords().values()))
+    mobile_xyz = np.array(list(mobile.GetCoords().values()))
+    n_atoms = len(ref_xyz)
+
+    ## assuming the reference is the one missing hydrogen atoms, and
+    ## that openeye will order them first, then this will get all the non hydrogen atoms
+    ## probably a better way to do this but this works for now
+    rmsd = np.sqrt((((ref_xyz - mobile_xyz[-n_atoms:]) ** 2) * 3).mean())
```
kaminow, line 80: I think it might be better to manually delete the Hs instead of assuming. it shouldn't take too long to iterate through all the atoms since the ligands are quite small

kaminow (commented):
Looks good to me!

PR #33, opened 2022-09-02
Add loss function class to handle step and uncertainty loss functions

kaminow:
Add `MSELoss` class as a wrapper over `torch.MSELoss` that can handle step loss and MSE loss with uncertainty (given as $L(\theta) = \frac{1}{N_{compounds}} \sum_{i=1}^{N_{compounds}}[\frac{pIC50_{i}^{exp}-pIC50^{pred}(\mu_{i}|\theta)}{\delta_{pIC50}}]^{2}$)

  - [x] Ready to go


kaminow:
@jchodera does this math look right?

covid_moonshot_ml/nn/loss/MSELoss.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Dispatch method for calculating loss. All arguments should be passed
+        regardless of actual loss function to keep an identical signature for
+        this class. Data is passed to `self.loss_function`.
+        """
+
+        if self.loss_type is None:
+            return self.loss_function(input, target).mean()
+        elif self.loss_type == "step":
+            return self.loss_function(input, target, in_range)
```
jchodera, line 44: Can you document that this calls `self.loss_function = self.step_loss`?

covid_moonshot_ml/nn/loss/MSELoss.py
```diff
@@ -0,0 +1,113 @@
...
+        elif self.loss_type == "uncertainty":
+            return self.loss_function(input, target, uncertainty)
```
jchodera, line 46: Can you document that this calls `self.loss_function = self.uncertainty_loss`?

covid_moonshot_ml/nn/loss/MSELoss.py
```diff
@@ -0,0 +1,113 @@
...
+            Uncertainty in `target` measurements
+
+        Returns
+        -------
+        """
+        ## Calculate loss
+        loss = super(MSELoss, self).forward(input, target)
+
+        ## Divide by uncertainty squared
+        loss /= uncertainty**2
```
jchodera, line 111: If the [pytorch `MSELoss`](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/loss.py#L532-L536) has already applied the `mean` reduction across the data, isn't it too late to scale the data by `uncertainty` if each measurement has a different value of `uncertainty`?

covid_moonshot_ml/nn/loss/MSELoss.py
```diff
@@ -0,0 +1,113 @@
...
+            of < 0 for `target` below lower bound, > 0 for `target` above upper
+            bound, and 0 or None for inside range
+
+        Returns
+        -------
+        torch.Tensor
+            Calculated loss
+        """
+        ## Calculate loss
+        loss = super(MSELoss, self).forward(input, target)
```
jchodera, line 73: Shouldn't `target` be whichever edge of the measurable interval the measurement is closest to?

jchodera (commented):
Apologies for the delay in review!

I'm not sure if I'm capable of giving this the detailed review it needs. Perhaps @yuanqing-wang could help?

I'm not quite clear what loss functions these two new modes are implementing.

I think we want to be very clear about documenting what we're trying to do here. We want to add two loss functions:
* **Uncertainty-weighting:** The loss is the negative log likelihood with a normal error function `loss = - \ln \sum_{n=1}^N Normal( (model_n - obs_n), uncertainty_n^2 )`, and we only included quantitative measurements.
* **Semiquantitative uncertainty-weighting:** If the measurement is within the quantitative range, use uncertainty-weighting. If the measurement is above or below the quantitative range, use the negative log likelihood for the deviation outside the interval, `loss = - \ln Normal( (model_n - boundary), sigma^2 )` where `sigma` is something like the maximum over all the `uncertainty_n` or some other suitable uncertainty that is of an appropriate order of magnitude.

In future, we can use a better loss function that can deal with individual measurements.

covid_moonshot_ml/nn/loss/MSELoss.py
kaminow: In [line 19](https://github.com/choderalab/covid-moonshot-ml/blob/add-loss-functions/covid_moonshot_ml/nn/loss/MSELoss.py#L19) the `MSELoss` class is initialized with `reduction="none"`, so the squared error for each example is returned.

covid_moonshot_ml/nn/loss/MSELoss.py
kaminow: Yes, that is what the value of `target` should be. I can hard-code these values somewhere though if you think that makes more sense.

covid_moonshot_ml/nn/loss/MSELoss.py
kaminow: done in [4fbf3c0](https://github.com/choderalab/covid-moonshot-ml/pull/33/commits/4fbf3c0e4f00b7bc3f65413484f7cbdb9987003c)

covid_moonshot_ml/nn/loss/MSELoss.py
kaminow: donein [4fbf3c0](https://github.com/choderalab/covid-moonshot-ml/pull/33/commits/4fbf3c0e4f00b7bc3f65413484f7cbdb9987003c)

covid_moonshot_ml/nn/loss/GaussianNLLLoss.py
```diff
@@ -0,0 +1,61 @@
...
+            Uncertainty in `target` measurements
+
+        Returns
+        -------
+        """
+        ## Fill in semiquant values
+        if self.include_semiquant and self.fill_value is not None:
+            idx = [r != 0 for r in in_range]
+            ## Clone to avoid modifying the original uncertainty measurements
+            uncertainty = uncertainty.clone()
```
JenkeScheen, line 48: I might be misunderstanding, but would it not be safer to use e.g. `uncertainty_clone` and continue with that? Not sure how the original variable is handled/whether it can be overwritten by the current design?

covid_moonshot_ml/utils.py
```diff
@@ -185,6 +186,8 @@ def train(
         Number of epochs to train for
     device : torch.device
         Where to run the training
+    loss_fn : cml.nn.MSELoss
+        Loss function
```
JenkeScheen, line 190: some more details on the type of loss fn would be helpful here

JenkeScheen (approved):
Looking good - looking forward to seeing more results on this work!

covid_moonshot_ml/nn/loss/GaussianNLLLoss.py
kaminow: I don't think it's necessary, but I've added this in [c720b64](https://github.com/choderalab/covid-moonshot-ml/pull/33/commits/c720b64c30399f5d831b6f4e70d2befd2b405601) just to be safe

covid_moonshot_ml/utils.py
kaminow: done in [6a84868](https://github.com/choderalab/covid-moonshot-ml/pull/33/commits/6a84868ad01de3f85c2305059ec20e1874292ec1)

PR #42, opened 2022-09-13
Add lig based filtering

JenkeScheen:
Implementation of docking input filter for https://github.com/choderalab/covid-moonshot-ml/issues/35. 

1. The current code has a placeholder for docking inputs but this should at some point be replaced with `ExperimentalCompoundData` in the same format (i.e. SMILES and some pointer, e.g. compound names); I wasn't able to find a way to request sample docking inputs from `ExperimentalCompoundData` in the current codebase. 

2. I wasn't sure where to place the SMARTS queries file, we can discuss the placement here if you like. For now I've put the file in `scripts/filter_inputs/smarts_queries.csv' but that wouldn't be the most user-friendly method in the long run.


kaminow:
with regard to 2, maybe we can add a `data/` folder in the top-level directory? not sure how accepted this is in practice though

apayne97:
re 2) i would vote for top level `data/` folder! we can also potentially put the resulting csv files and stuff in there

apayne97:
One bit of confusion is that there are multiple ways we can get a list of SMILES strings.

Currently for `hallucinate_fragalysis.py` the bit that does that is here:
https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/scripts/hallucinate_fragalysis.py#L418

with this code generating the `CrystalCompoundData` object: https://github.com/choderalab/covid-moonshot-ml/blob/5d2be92c053a39840f675a91fde699ffe6dd5bc1/covid_moonshot_ml/datasets/utils.py#L328

JenkeScheen:
thanks! will play around with thisf or a bit to see if I can get it to work. Could you point me to example input files for `hallucinate_fragalysis.py`?

--> I'm used to discussing on PRs like this (not much difference as the issue is referenced anyway), happy to switch back to the issue page if you prefer though!

apayne97:
So `parse_fragalysis_data` won't work as is for you because you don't have access to the full fragalysis dataset yet; although you should be able to download it from the website manually. On lilac it is here, in @kaminow's directory: `/data/chodera/kaminowb/stereochemistry_pred/fragalysis`

However, the relevant csv from that dataset is here (this is the `args.x` parameter in `hallucinate_fragalysis`):
`/data/chodera/kaminowb/stereochemistry_pred/fragalysis/extra_files/Mpro_compound_tracker_csv.csv`
[Mpro_compound_tracker_csv.csv](https://github.com/choderalab/covid-moonshot-ml/files/9565961/Mpro_compound_tracker_csv.csv)

And you could take a part of the `parse_fragalysis_data` function and ignore any references to the directory (I think right now the script uses the directory to identify the filename of the associated sdf file).

> I'm used to discussing on PRs like this (not much difference as the issue is referenced anyway), happy to switch back to the issue page if you prefer though!
as long as they're linked I guess it doesn't matter?

apayne97:
this test function should work for you? this will get you a list of CrystalCompoundData objects

```
def parse_fragalysis_csv(frag_fn):
        mols_wo_sars2_xtal = sars2_filtered[sars2_filtered["Dataset"].isna()][["Compound ID", "SMILES", "Dataset"]]
        mols_w_sars2_xtal = sars2_filtered[~sars2_filtered["Dataset"].isna()][["Compound ID", "SMILES", "Dataset"]]
        
        ## Load in csv
            sars2_structures = pandas.read_csv(frag_fn).fillna("")
   
         ## Construct sars_xtal list
            sars_xtals = {}
            for data in sars2_structures.to_dict('index').values():
                cmpd_id = data["Compound ID"]
                dataset = data["Dataset"]
                if len(dataset) > 0:
                    if not sars_xtals.get(cmpd_id) or '-P' in dataset:
                        sars_xtals[cmpd_id] = CrystalCompoundData(
                            smiles=data["SMILES"],
                            compound_id=cmpd_id,
                            dataset=dataset,
                            sdf_fn=None
                        )
            return sars_xtals
```

covid_moonshot_ml/datasets/utils.py
```diff
@@ -4,7 +4,7 @@
 import pandas
 from rdkit.Chem import CanonSmiles, FindMolChiralCenters, MolFromSmiles
 import re
-
+import csv
```
kaminow, line 7: is there a reason to use `csv` instead of `pandas`? just for speed purposes?

covid_moonshot_ml/datasets/utils.py
```diff
@@ -504,3 +504,116 @@ def get_ligand_RMSD_mdtraj(ref_fn, mobile_fn):
...
+    with open(smarts_queries, "r") as queries:
+        queries_reader = csv.reader(queries)
+        query_smarts = [ q[0] for q in queries_reader ]
+
+        if ignore_comment:
+            # only keep SMARTS queries that are not commented.
+            query_smarts = [ q for q in query_smarts if not q[0] == "#"]
+        else:
+            # some of the SMARTS queries are commented - remove these for now.
+            query_smarts = [ q if not q[0] == "#" else q[1:] for q in query_smarts]
```
kaminow, line 590: should these be flipped? if not, could you maybe rewrite the comment for the `ignore_comment` option to make it a bit more clear? as it's currently written, it sounds like `ignore_comment=True` means you ignore any leading `#` and add the SMARTS query, while `ignore_comment=False` means remove any commented SMARTS

scripts/filter_inputs/smarts_queries.csv
```diff
@@ -0,0 +1,13 @@
+#A1Ac2ccc(Cl)cc2C2(CCN(c3cncc4ccccc34)C2=O)A1,tetralinlike-5spiro-isoquinoline
```
kaminow, line 1: can we delete this file now that we have the copy in `data/`?

kaminow (changes requested):

covid_moonshot_ml/datasets/utils.py
JenkeScheen: mostly just did what I'm used to - there some advantages with handling large datasets using a CSV iterator. Changed to `pandas` to align with the rest of the codebase, see https://github.com/choderalab/covid-moonshot-ml/pull/42/commits/94965a717af26883e9dd701470c6970c32744f7f

covid_moonshot_ml/datasets/utils.py
JenkeScheen: I think this kind of construct is inherently confusing; I adjusted the docstring to make it clearer (hopefully): https://github.com/choderalab/covid-moonshot-ml/pull/42/commits/82d7c0657a5b665338b4742f5fa896cc91fc10a0

scripts/filter_inputs/smarts_queries.csv
JenkeScheen: yes, forgot to `git rm`. see https://github.com/choderalab/covid-moonshot-ml/pull/42/commits/00e08fbd3e8560b59f94f70910c7496a23544282

apayne97 (approved):
I think this will work for now, but something to note for the future is that I'm not sure whether or not OpenEye will be able to handle the enhanced stereochemistry that ASAP will be using for molecules as described here:
https://depth-first.com/articles/2022/02/09/v3000-molfile-enhanced-stereochemistry-representation/.
I think the solution if OpenEye doesn't work would be to use RDKit instead.

covid_moonshot_ml/datasets/utils.py
apayne97: Maybe just rename to a longer boolean? i.e. if `drop_commented_smarts_strings == True`, then you ignore the queries that are commented, but if it is false, you don't `drop_commented_smarts_strings`. I thing the issue is that here "comment" could either refer to the syntax of using comments, or to the commented lines themselves.

PR #44, opened 2022-09-14
Add ability to build both monomer and dimer design units

### LINKED ISSUE #34: Update Design Unit builder to keep the protein dimer
# To-Do
- [x] keep the dimer in the design unit and confirm that it is getting written out

kaminow:
Updated the function to build design units to be able to handle building both monomers and dimers. Updated the `hallucinate_fragalysis.py` script to automatically generate both. Resolves #34

### DIFF
--- a/scripts/hallucinate_fragalysis.py
+++ b/scripts/hallucinate_fragalysis.py
@@ -117,13 +117,37 @@ def write_fragalysis_output(in_dir, out_dir, sort_key="POSIT_prob"):
         prot = oechem.OEGraphMol()
         du.GetProtein(prot)
         lig = load_openeye_sdf(f"{tmp_in_dir}/docked.sdf")
+
+        ## Set ligand title
         lig.SetTitle(f"{compound_id}_{best_str}")
 
         ## First save apo
         save_openeye_pdb(prot, f"{tmp_out_dir}/{best_str}_apo.pdb")
 
-        ## Combine protein and ligand and save
+        ## Combine protein and ligand
         oechem.OEAddMols(prot, lig)
+
+        # ## Get number of residues per chain
+        # num_res_chain_dict = {}
+        # hv = oechem.OEHierView(prot)
+        # for chain in hv.GetChains():
+        #     max_resid = 0
+        #     for frag in chain.GetFragments():
+        #         frag_max = max(
+        #             [r.GetResidueNumber() for r in frag.GetResidues()]
+        #         )
+        #         if frag_max > max_resid:
+        #             max_resid = frag_max
+        #     num_res_chain_dict[chain.GetChainID()] = max_resid
+        # ## Rename and renumber ligand residue
+        # for a in prot.GetAtoms():
+        #     r = oechem.OEAtomGetResidue(a)
+        #     if r.GetName() == "UNL":
+        #         r.SetHetAtom(True)
+        #         r.SetName("LIG")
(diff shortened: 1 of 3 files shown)


apayne97:
Does this code include ligands from chain A and chain B? https://github.com/choderalab/covid-moonshot-ml/blob/add-dimer/covid_moonshot_ml/datasets/utils.py#L490

Because I think what we want is to be able to specify the chain of interest and then keep the ligand from only that chain, even if we keep the protein from both chain A and B.

kaminow:
this is handled by the `lig_chain` option here:
https://github.com/choderalab/covid-moonshot-ml/blob/add-dimer/covid_moonshot_ml/datasets/utils.py#L494-L500

apayne97:
Oh! cool. Do you have a relative path to a dataset you've generated using this? seems like you've solved the challenge in #34!

apayne97 (approved):

scripts/hallucinate_fragalysis.py
```diff
@@ -117,23 +133,49 @@ def write_fragalysis_output(in_dir, out_dir, sort_key="POSIT_prob"):
...
+        #             max_resid = frag_max
+        #     num_res_chain_dict[chain.GetChainID()] = max_resid
+        # ## Rename and renumber ligand residue
+        # for a in prot.GetAtoms():
+        #     r = oechem.OEAtomGetResidue(a)
+        #     if r.GetName() == "UNL":
+        #         r.SetHetAtom(True)
+        #         r.SetName("LIG")
+        #         r.SetResidueNumber(num_res_chain_dict[r.GetChainID()] + 100)
+        #     oechem.OEAtomSetResidue(a, r)
```
ijpulidos, line 165: Not a blocker. Is this code expected to be uncommented and used eventually? If not it should be better if we get rid of it :)

ijpulidos (approved):
Looks good! Just one comment that is not blocking.

Also, I noticed that some of the cli arguments only have a short form (for example `x` for the crystal structure). I don't really know what's the best practice, but I think we also want a longer form, such that when we use them we don't have hard to infer variables such as `args.x` but instead something like `args.structure` or something like that. This is not a blocker, just to keep in mind for future iterations.

ijpulidos:
I see there is a merge conflict that need to be attended. @kaminow is this something that can be easily solved or does it need our merging and rebasing "magic"?

kaminow:
I think it's something I can take care of easily. this branch was created after the license scrubbing, so no need to rebase before merging

mikemhenry:
> I don't really know what's the best practice, but I think we also want a longer form, such that when we use them we don't have hard to infer variables such as args.x but instead something like args.structure or something like that. This is not a blocker, just to keep in mind for future iterations.

+1 to using meaningful variables.

apayne97:
@kaminow I *think* the conflict here is that @JenkeScheen had commented out most of `datasets/utils.py` in this commit:
https://github.com/choderalab/covid-moonshot-ml/commit/94965a717af26883e9dd701470c6970c32744f7f

apayne97:
resolved with this commit: https://github.com/choderalab/covid-moonshot-ml/commit/56dfe6bef2dbc46721ac1b99ebc37bbedbab5239

JenkeScheen:
oh no :anguished: so sloppy of me. Thanks for resolving this @apayne97!

PR #52, opened 2022-09-28
Hotfix: Using correct function names and imports -- General clean up

ijpulidos:
## Description
This fixes some references problems that the `hallucinate_fragalysis` branch had. I created a new branch such that it's easier to check what was changed.  This one is aimed to be merged into the `hallucinate_fragalysis` branch.

Other than that I cleaned up some unused imports and references, and also made some comments in TODOs for future reference. Such that we can improve the code and avoid unexpected behavior.

The aim was not to change any of the behavior/logic just make the code work where it was broken and tidying up a bit.

## Status
- [x] Ready to go

### DIFF
--- a/covid_moonshot_ml/nn/models/SchNetBind.py
+++ b/covid_moonshot_ml/nn/models/SchNetBind.py
@@ -11,8 +11,6 @@ class SchNetBind(SchNet):
     def __init__(self, *args, **kwargs):
         super(SchNetBind, self).__init__(*args, **kwargs)
 
-    # TODO: Does not match signature in base class
-    #  https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.models.SchNet.forward
     def forward(self, z, pos, lig):
         """
         Forward pass through the model. Each forward pass through this class
--- a/covid_moonshot_ml/nn/models/SchNetBind.py
+++ b/covid_moonshot_ml/nn/models/SchNetBind.py
@@ -11,6 +11,8 @@ class SchNetBind(SchNet):
     def __init__(self, *args, **kwargs):
         super(SchNetBind, self).__init__(*args, **kwargs)
 
+    # TODO: Does not match signature in base class
+    #  https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.models.SchNet.forward
     def forward(self, z, pos, lig):
         """
         Forward pass through the model. Each forward pass through this class
--- a/covid_moonshot_ml/tests/test_covid_moonshot_ml.py
+++ b/covid_moonshot_ml/tests/test_covid_moonshot_ml.py
@@ -5,10 +5,6 @@ Unit and regression test for the covid_moonshot_ml package.
 # Import package, test suite, and other packages as needed
 import sys
 
-import pytest
-
-import covid_moonshot_ml
-
 
 def test_covid_moonshot_ml_imported():
     """Sample test, will always pass so long as import statement worked."""
--- a/scripts/hallucinate_fragalysis.py
+++ b/scripts/hallucinate_fragalysis.py
@@ -317,6 +317,8 @@ def mp_func(
         ret_code = poser.Dock(pose_res, dock_lig)
 
     ## Check results
+    # FIXME: This uses the `pose_res` defined inside previous conditional.
+    #  What if that previous conditional doesn't happen? `pose_res` might be undefined.
     if ret_code == oedocking.OEDockingReturnCode_Success:
         if dock_sys == "posit":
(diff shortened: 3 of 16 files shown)


covid_moonshot_ml/tests/test_covid_moonshot_ml.py
```diff
@@ -5,10 +5,6 @@
 # Import package, test suite, and other packages as needed
 import sys
 
-import pytest
```
apayne97, line 8: I think we actually want to keep these since this will be a unit test in the future but since it is so undeveloped now I don't think its worth haggling about, instead I made this issue: https://github.com/choderalab/covid-moonshot-ml/issues/53

scripts/plot_docking_results.py
```diff
@@ -1,5 +1,5 @@
 import sys, os, argparse
-import numpy as np
+# TODO: Do we need to add plotly to our environment yaml?
```
apayne97, line 2: yes! first I think we should figure out the best way to save the env: https://github.com/choderalab/covid-moonshot-ml/issues/41

scripts/upset_plot_ranked.py
```diff
@@ -4,6 +4,7 @@
 import os
 import pandas
 import pickle as pkl
+# # TODO: Do we need to add upsetplot to our environment yaml?
```
apayne97, line 7: yes! first I think we should figure out the best way to save the env: https://github.com/choderalab/covid-moonshot-ml/issues/41

apayne97 (approved):
Looks good to me. Added this issue to follow up with the unittest changes: https://github.com/choderalab/covid-moonshot-ml/issues/53

covid_moonshot_ml/nn/models/SchNetBind.py
```diff
@@ -11,6 +11,8 @@ class SchNetBind(SchNet):
     def __init__(self, *args, **kwargs):
         super(SchNetBind, self).__init__(*args, **kwargs)
 
+    # TODO: Does not match signature in base class
+    #  https://pytorch-geometric.readthedocs.io/en/latest/modules/nn.html#torch_geometric.nn.models.SchNet.forward
```
kaminow, line 15:
```suggestion
```
the signature of this model isn't supposed to match the base class, since we need the ligand labels in order to perform our two forward passes through the model. if there's a better way of doing this, can you please update the comment to add some details. otherwise, I think let's delete the comment to avoid confusion

scripts/hallucinate_fragalysis.py
```diff
@@ -317,6 +317,8 @@ def mp_func(
         ret_code = poser.Dock(pose_res, dock_lig)
 
     ## Check results
+    # FIXME: This uses the `pose_res` defined inside previous conditional.
+    #  What if that previous conditional doesn't happen? `pose_res` might be undefined.
```
kaminow, line 321:
```suggestion
```
the logic here is a bit convoluted, but `pose_res` is initially created above in line 282. the above conditional only runs if the initial `POSIT` run with relaxation failed, and its job is to re-run the docking without relaxation

scripts/hallucinate_fragalysis.py
```diff
@@ -325,6 +327,7 @@ def mp_func(
         ## Get the Chemgauss4 score (adapted from kinoml)
         pose_scorer = oedocking.OEScore(oedocking.OEScoreType_Chemgauss4)
         pose_scorer.Initialize(du)
+        # TODO: If `dock_sys` is not "posit", then `posed_mol` might be undefined here.
```
kaminow, line 330:
```suggestion
```
if `dock_sys` is `"hybrid"` then `posed_mol` is created directly on line 292. the reason it's set up like this is because of the way the different docking methods are run in OpenEye

scripts/hallucinate_fragalysis.py
```diff
@@ -369,6 +372,7 @@ def mp_func(
         dimer,
         f"{out_base}/docked.sdf",
         rmsd,
+        # TODO: If `dock_sys` is not "posit", then `posit_prob` might be undefined here.
```
kaminow, line 375:
```suggestion
```
if `dock_sys` is `"hybrid"` then `posit_prob` is manually set to `-1` on line 295

kaminow (changes requested):
these are good catches, thanks! a couple of the `TODO` comments you left in the `hallucinate_fragalysis` script I think are unnecessary (I think just due to some slightly convoluted logic obscuring parts of the code, which I think is mostly OpenEye's fault but probably at least a bit mine as well)

ijpulidos:
@kaminow Thanks for looking into these, I was just making sure these were intended. I agree with your comments.

kaminow (approved):
looks good to me, thanks again!

PR #56, opened 2022-09-29
add ligand filtering

apayne97:
## Description

## Todos
  - [x] need to confirm that this filtering will work in the docking pipeline
  - [x] @kaminow can you try running this on the cluster?

## Status
- [x] Ready to go

### DIFF
--- a/covid_moonshot_ml/datasets/utils.py
+++ b/covid_moonshot_ml/datasets/utils.py
@@ -483,6 +483,7 @@ def parse_fragalysis_data(frag_fn, x_dir, cmpd_ids=None, o_dir=False):
         cmpd_id = data["Compound ID"]
         dataset = data["Dataset"]
         if len(dataset) > 0:
+            ## TODO: is this the behaviour we want? this will build an empty object if there isn't a dataset
             if not sars_xtals.get(cmpd_id) or "-P" in dataset:
                 sars_xtals[cmpd_id] = CrystalCompoundData(
                     smiles=data["SMILES"],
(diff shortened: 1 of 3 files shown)


mikemhenry (approved):
LGTM!

kaminow (approved):
I reworked the logic for filtering in the `hallucinate_fragalysis.sh` script and fixed a small typo. also changed `--smarts_queries` to not have a default in order to avoid accidental filtering. other than these changes, looks good!

apayne97:
> I reworked the logic for filtering in the `hallucinate_fragalysis.sh` script and fixed a small typo. also changed `--smarts_queries` to not have a default in order to avoid accidental filtering. other than these changes, looks good!

everyone but what I asked about makes sense! good catch on the typo

covid_moonshot_ml/datasets/utils.py
```diff
@@ -483,6 +483,7 @@ def parse_fragalysis_data(frag_fn, x_dir, cmpd_ids=None, o_dir=False):
         cmpd_id = data["Compound ID"]
         dataset = data["Dataset"]
         if len(dataset) > 0:
+            ## TODO: is this the behaviour we want? this will build an empty object if there isn't a dataset
```
kaminow, line 486: not sure if this needs to be resolved in this PR, but this is a good point. might make more sense to just not add anything if the `"Dataset"` value is empty

JenkeScheen (approved):
LGTM2! I just got ready with some coffee to write this code but looks like you beat me to it 👍  I imagine there will be some kinks appearing when testing it in the actual pipeline @kaminow, happy to help straighten them out!

kaminow:
actually, everything ran just fine on the cluster using `data/smarts_queries.csv`! I think this can be merged now

apayne97:
@ijpulidos @mikemhenry do we want to be squashing the commits or not?

mikemhenry:
When merging PRs to main, I like to squash since then it is easy to use got bisect to find which PR broke something + it keeps the history on main clean. That's we do we for other projects.

Issue #47, opened 2022-09-20
Use protein prep tools to standardize targets


apayne97:
# Protein Prep Stuff to Implement
- [x] make mutation using SEQRES
- [x] re-build side chains
- [x] #80
- [x] #62
- [x] #60 
- [x] #61 

# Relevant PRs
- #58

apayne97:
In order to standardize everything, we should include a DBREF and SEQRES code to set all the structures to. I downloaded the fasta sequences of all our MERS PDB structures (in https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/data/mers-structures.yaml) here:
[rcsb_pdb_20220921044206.txt](https://github.com/choderalab/covid-moonshot-ml/files/9616317/rcsb_pdb_20220921044206.txt)

And used clustal omega to align them:
[clustalo-I20220921-124415-0802-27476461-p1m.clustal_num.txt](https://github.com/choderalab/covid-moonshot-ml/files/9616319/clustalo-I20220921-124415-0802-27476461-p1m.clustal_num.txt)

Most have this sequence (included from 5C3N here):
```
DBREF1 5C3N A    1   306  UNP                  A0A0A7E693_9BETC
DBREF2 5C3N A     A0A0A7E693                       3248        3553
DBREF1 5C3N B    1   306  UNP                  A0A0A7E693_9BETC
DBREF2 5C3N B     A0A0A7E693                       3248        3553
SEQRES   1 A  306  SER GLY LEU VAL LYS MET SER HIS PRO SER GLY ASP VAL
SEQRES   2 A  306  GLU ALA CYS MET VAL GLN VAL THR CYS GLY SER MET THR
SEQRES   3 A  306  LEU ASN GLY LEU TRP LEU ASP ASN THR VAL TRP CYS PRO
SEQRES   4 A  306  ARG HIS VAL MET CYS PRO ALA ASP GLN LEU SER ASP PRO
SEQRES   5 A  306  ASN TYR ASP ALA LEU LEU ILE SER MET THR ASN HIS SER
SEQRES   6 A  306  PHE SER VAL GLN LYS HIS ILE GLY ALA PRO ALA ASN LEU
SEQRES   7 A  306  ARG VAL VAL GLY HIS ALA MET GLN GLY THR LEU LEU LYS
SEQRES   8 A  306  LEU THR VAL ASP VAL ALA ASN PRO SER THR PRO ALA TYR
[...]

Issue #46, opened 2022-09-20
Re-write hallucinate_fragalysis as a more flexible API


apayne97:
# Intro
Currently the main `hallucinate_fragalysis.py` script includes a combination of hard-coded settings, pulling from the `covid_moonshot_ml` api, and self-contained functions. 

The goal is to section off parts of the code into functions within the API and to have a much simpler script that pulls from the API.

# Steps to Separate Out
|Goal | Inputs | Outputs | Moved (Y/N) | New Code Location | 
|-|-|-|-|-|
| 1. Downloading PDB files | list of PDBs | directory of downloaded PDBs | N | |
| 2. Re-build side chains, missing loops |  directory of downloaded PDBs, re-building options | directory of re-built PDBs | N | |
| 3. Aligning PDB files to a reference | path to reference, directory of re-built PDBs | directory of aligned PDBs | N | |
| 4. Build design units | path to aligned PDB, path to fragalysis structure | `Compound_ID_Structure_Source`.oedu, `Compound_ID_Structure_Source`_seqres.pdb, `Compound_ID_Structure_Source`_predocked.sdf | N | |
| 5. Run docking | path to design unit | `Compound_ID_Structure_Source`_bound.pdb, `Compound_ID_Structure_Source`_docked.sdf | N | |
| 6. Calculate RMSD |  predocked `OEMol`, docked `OEMol` | RMSD (float) | N | |
| 7. Pull all scores from results | path to `*_docked.sdf` OR OE objects | scores | N | |
| 8. Write CSV of docking results | `Complex_ID`, `Compound_ID`, `Structure_Source`, `Output_Dir`, and any scores | all_results.csv | N | |
| 9. Analyze results | path to `all_results.csv` | `all_results_cleaned.csv`, `by_compound.csv`, `by_structure.csv`, `mers_fauxalysis.csv` | Y | `scripts/clean_results_csv.py` |
| 10. Generate "fauxalysis" directory | path to docking results, path to `mers_fauxalysis.csv` | fauxalysis directory | Y* | `scripts/fauxalysis_from_docking.py` | 
| 11. Plot docking results | path to outputs from `9.`| a bunch of plots! | Y | `scripts/plot_docking_results.py` |

*10 is also in hallucinate fragalysis but they have very different implementations: https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/scripts/hallucinate_fragalysis.py#L54

## The current plan is to have these as individual scripts:
[...]

PR #58, opened 2022-09-30
split up protein prep and use SEQRES to make mutations

apayne97:
## Description
Enable preparing the receptors as an independent step from the whole 'hallucinate fragalysis' pipeline.

## Todos
  - [x] implement everything described in this issue #47 
  - [x] track the new changes highlighted here #46 
  - [x] implement Ben's protein prep improvements for MERS structure prep
  - [x] add functionality to dock an SDF file of ligands against a directory of prepped `.oedu` files

## Questions
- there are still protein prep things to fix but we'll have to live with this for now

## Status
- [x] Ready to go


apayne97:
Here was the traceback for the kinoml test that didn't work:
```
OEProteinStructureFeaturizer:   0%|          | 0/1 [00:00<?, ?it/s]DPI: 0.23, RFree: 0.27, Resolution: 2.34
Processing BU # 1 with title: 3C-LIKE PROTEINASE, chains AB
Found unresolved N-terminal with 1 residues before GLY 2   A 1  , with sequence S
Found 2 residue gap between PRO 75   A 1   and LEU 78   A 1  , with sequence AN
Found 2 residue gap between LEU 170   A 1   and GLY 173   A 1  , with sequence AN
Found unresolved C-terminal with 6 residues after ILE 300   A 1  , with sequence MGVVMQ
Found unresolved N-terminal with 1 residues before GLY 2   B 2  , with sequence S
Found 2 residue gap between ALA 74   B 2   and ASN 77   B 2  , with sequence PA
Found unresolved C-terminal with 5 residues after MET 301   B 2  , with sequence GVVMQ
Warning: There was a problem building some missing pieces, built as much as was possible
   Falling back to charging protein with OEMMFF94Charges
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/scripts/test-protein-prep.py", line 123, in <module>
    main()
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/scripts/test-protein-prep.py", line 119, in main
    featurizer.featurize(systems)
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/features/core.py", line 63, in featurize
    features = self._featurize(systems)
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/features/core.py", line 309, in _featurize
    features = [self._featurize_one(s) for s in tqdm(systems, desc=self.name)]
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/features/core.py", line 309, in <listcomp>
    features = [self._featurize_one(s) for s in tqdm(systems, desc=self.name)]
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/features/protein.py", line 226, in _featurize_one
    protein = self._process_protein(
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/features/core.py", line 1124, in _process_protein
    protein_structure = apply_mutations(protein_structure, amino_acid_sequence)
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/kinoml/modeling/OEModeling.py", line 1423, in apply_mutations
    if oespruce.OEMutateResidue(
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/site-packages/openeye/oespruce.py", line 992, in OEMutateResidue
    return _oespruce.OEMutateResidue(*args)
TypeError: in method 'OEMutateResidue', argument 3 of type 'std::string const'
OEProteinStructureFeaturizer:   0%|          | 0/1 [00:11<?, ?it/s]
```

covid_moonshot_ml/datasets/utils.py
```diff
@@ -11,59 +11,8 @@
...
-SEQRES  21 B  306  VAL LEU ASP MET CYS ALA SER LEU LYS GLU LEU LEU GLN
-SEQRES  22 B  306  ASN GLY MET ASN GLY ARG THR ILE LEU GLY SER ALA LEU
-SEQRES  23 B  306  LEU GLU ASP GLU PHE THR PRO PHE ASP VAL VAL ARG GLN
-SEQRES  24 B  306  CYS SER GLY VAL THR PHE GLN
-"""
-
-
-def add_seqres(pdb_in, pdb_out=None):
+
+def add_seqres(pdb_in, seqres_str=False, dbref_str=False, pdb_out=None):
```
kaminow, line 15:
```suggestion
def add_seqres(pdb_in, seqres_str=None, dbref_str=None, pdb_out=None):
```
I think it's better to have the defaults as `None` since we really only want one type per argument (either `str` or `None` in this case). also, need docstrings for the new args!

covid_moonshot_ml/datasets/utils.py
```diff
@@ -81,9 +30,12 @@ def add_seqres(pdb_in, pdb_out=None):
     pdbfile_lines = [re.sub("CL", "Cl", l) for l in pdbfile_lines]
     # # remove ligand hetatoms
     # pdbfile_lines = [ line for line in pdbfile_lines if 'LIG' not in line ]
+    if seqres_str:
+        pdbfile_lines = [line for line in pdbfile_lines if not "SEQRES" in line]
     pdbfile_contents = "".join(pdbfile_lines)
+
     if not "SEQRES" in pdbfile_contents:
-        pdbfile_contents = MPRO_SEQRES + pdbfile_contents
+        pdbfile_contents = seqres_str + pdbfile_contents
```
kaminow, line 38: I think we only want to add the `seqres_str` in the event that something is actually passed in. the way it's currently set up, if there's just no SEQRES header in the PDB file, you'll get an error with the default arguments when trying to add a `bool` and a `str`. something like this maybe:
```suggestion
        pdbfile_contents = seqres_str + "".join(pdbfile_lines)
    else:
        pdbfile_contents = "".join(pdbfile_lines)
```

covid_moonshot_ml/datasets/utils.py
```diff
@@ -93,6 +45,23 @@ def add_seqres(pdb_in, pdb_out=None):
     print(f"Wrote {pdb_out}", flush=True)
 
 
+def seqres_to_res_list(SEQRES_str):
```
kaminow, line 48:
```suggestion
def seqres_to_res_list(seqres_str):
```
I think Python variable names shouldn't have capital letters

covid_moonshot_ml/datasets/utils.py
```diff
@@ -93,6 +45,23 @@ def add_seqres(pdb_in, pdb_out=None):
...
+    https://www.wwpdb.org/documentation/file-format-content/format33/sect3.html#SEQRES
+    Parameters
+    ----------
+    SEQRES_str
+
+    Returns
+    -------
+
+    """
+    seq_lines = [line[19:] for line in SEQRES_str.split("\n") if " A " in line]
```
kaminow, line 59: is this for checking that the SEQRES is only for chain A? I think it might be better to manually check the column in the entry that gives the chain, just in case the `" A "` pattern somehow appears elsewhere in the entry

kaminow (commented):
just some stuff that initially jumped out at me. still need to look more in depth at the mutating code

kaminow:
that's strange... it looks like there's no `res_name` for that residue either, so maybe it's some kind of ghost residue? I added a new implementation of the function that I think is more direct (using `OEMutateResidues` to mutate them all at once rather than doing each mutation individually). can you lmk if you still get the same error with this implementation?

apayne97:
Running! I was hoping to avoid mutating all the residues since its much slower to do it that way. The point was to find only the residues where the res_names are different. It looks like it worked in theory although we know that the sequences aren't the same length for each one so that will need to be adjusted!

kaminow:
sorry, I should've been clearer about that, it doesn't mutate every single residue, just the ones that are different like you said (that's what the `if` statement in the `mut_map` dict comprehension does). also, the only length requirement is that the `SEQRES` entry be the same length as the number of residues in the protein, which I think should be the case but I could be wrong about that

apayne97:
But the proteins have different residue lengths (i.e. 7DR8 is missing the initial Serine, and some of the proteins have a series of histidines at the n-terminus)
also, I think it also fails if we keep waters!
Being able to connect the residue number to the SEQRES we want to use would be nice

apayne97:
ughhhhh also if there are missing sequences (i.e. in the loops) it won't skip them intelligently:
image

kaminow:
ah yeah those are definitely issues. I have an idea though, hopefully will commit in a few minutes

apayne97:
> ah yeah those are definitely issues. I have an idea though, hopefully will commit in a few minutes

sorry just saw this, my most recent commit works

covid_moonshot_ml/datasets/utils.py
apayne97: agreed, done here https://github.com/choderalab/covid-moonshot-ml/pull/58/commits/4486d0fd4a101824941f03ff1cde7e636a3b5c15

covid_moonshot_ml/datasets/utils.py
apayne97: agreed, done here: https://github.com/choderalab/covid-moonshot-ml/pull/58/commits/4486d0fd4a101824941f03ff1cde7e636a3b5c15

covid_moonshot_ml/datasets/utils.py
apayne97: done here: https://github.com/choderalab/covid-moonshot-ml/pull/58/commits/4486d0fd4a101824941f03ff1cde7e636a3b5c15

kaminow:
I think your new approach looks good. it will fail though in the case that chain B has a different sequence than chain A, is this something that can happen?

apayne97:
> it will fail though I think in the case that chain B has a different sequence than chain A, is this something that can happen?

Which chain A and chain B are you referring to? the one in the starting molecule or the one in the sequence that we want?

kaminow:
I meant in the SEQRES, but since it's a homodimer I guess that shouldn't really happen

scripts/test-protein-prep.py
```diff
@@ -0,0 +1,136 @@
...
+            ref_prot=args.ref_prot,
+            dimer=True,
+            mobile_chain=mobile_chain,
+            ref_chain="A",
+        )
+        aligned_fn = f"{out_name}_03aligned_chain{mobile_chain}.pdb"
+        save_openeye_pdb(aligned_prot, aligned_fn)
+
+        ## Prep the receptor using various SPRUCE options
+        site_residue = "HIS:41: :A"
```
apayne97, line 112: obviously it would be better to do this in a way that isn't hard-coded!

If we don't solve this in this PR then we should make an issue for it.

scripts/run_docking_oe.py
```diff
@@ -0,0 +1,310 @@
...
+
+        dataset = dataset.groups()[0]
+        du = oechem.OEDesignUnit()
+        if not oechem.OEReadDesignUnit(fn, du):
+            print(f"Failed to read DesignUnit {fn}", flush=True)
+            continue
+        try:
+            du_dict[dataset].append(du)
+        except KeyError:
+            du_dict[dataset] = [du]
```
apayne97, line 95: @kaminow is this what is capturing the prepped du directory name? I think thats why we've dropped the chain information downstream, it's only capturing before the `_`? We can get the original compound_id but we should definitely keep track somewhere if we're getting the, i.e. Mpro-P0060_0A or Mpro-P0060_0B structures.

kaminow: fixed now!

kaminow (approved):
sorry for the delay, but this looks good! let's try and get this merged back into `main` soon

PR #65, opened 2022-10-04
Update fauxalysis script

apayne97:
## Description

## Todos
  - [x] combine @kaminow and @apayne97 's "make fauxalysis dataset from docking results" implementations into a single functioning script that can be run independently of the rest of the pipeline
  - [x] add ability to copy fragalysis structure
  - [x] test generation of complete fragalysis dataset
  - [x] add comments
  - [x] add ability to subset the fauxalysis dataset by the contents of a yaml file
  - [x] correct ligand chain and names
  - [x] improve (?) way to create combined sdf file
  - [x] create a design unit-saving function for OpenEye Design Units that solves #75 

## Status
- [x] Ready to go


apayne97:
this should be good to go now!

scripts/clean_results_csv.py
```diff
@@ -36,29 +47,46 @@ def main():
...
-    ]
+    # ToDo: deal with this issue of the "_bound" compound IDs in the fragalysis database.
+    # dr.df.Compound_ID = [
+    #     string.replace("_bound", "") for string in dr.df.Compound_ID
+    # ]
 
     ## Add Complex_ID
-    dr.df["Complex_ID"] = f"{dr.df.Compound_ID}_{dr.df.Structure_Source}"
+    ## This is not the same thing as f"{dr.df.Compound_ID}_{dr.df.Structure_Source}", as this enables rowwise addition
+    ## as opposed to adding the *entire* series as a single string
```
ijpulidos, line 88: We can just remove this line/suggestion since it is not the same thing. My bad!
```suggestion
```

scripts/create_fauxalysis_subset_from_yaml.py
```diff
@@ -0,0 +1,84 @@
+import sys, os, argparse, yaml, shutil
```
ijpulidos, line 1: I think it would be helpful to have some kind of docstring here or in the functions to explain what this script does and how it is supposed to be used.

scripts/write_cmpd_to_frag_dict.py
```diff
@@ -0,0 +1,56 @@
+"""
+Build library of ligands from a dataset of holo crystal structures docked to a
+different dataset of apo structures.
+"""
```
ijpulidos, line 4: I think we should say here that the output is a YAML file and maybe talk about the format of the file in broad terms.

scripts/clean_results_csv.py
```diff
@@ -72,7 +100,8 @@ def main():
 
     ## Get several dataframes
     dr.get_compound_df()
-    dr.get_structure_df(resolution_csv="../data/mers_structures.csv")
+    if not args.keep_structure_source:
+        dr.get_structure_df(resolution_csv="../data/mers_structures.csv")
```
ijpulidos, line 104: I think we should try to avoid hardcoded paths here, since it make it depend on where you run it. Maybe we can use `__file__` variable for this
```suggestion
        dr.get_structure_df(resolution_csv=os.path.join(os.path.dirname(__file__), "..", "data", "mers_structures.csv"))
```
Or something similar.

ijpulidos (commented):
Overall it looks very good, just a few comments that I think should be addressed before trying to merge. Good work!

scripts/clean_results_csv.py
```diff
@@ -1,3 +1,10 @@
+"""
+This script cleans up the CSV currently (as of 2022.10.14) generated by `hallucinate_fragalysis.py` into
+a format the following scripts can use.
```
JenkeScheen, line 3: consider mentioning (some of) the following scripts

scripts/clean_results_csv.py
```diff
@@ -1,3 +1,10 @@
...
+
+Example Usage:
+    python clean_results_csv.py -i /lila/data/chodera/kaminowb/stereochemistry_pred/mers/mers_fragalysis/posit_hybrid_no_relax_keep_water_filter/all_results.csv -o /lila/data/chodera/paynea/posit_hybrid_no_relax_keep_water_filter_frag -d -s
```
JenkeScheen, line 6: best to format this in the same way as the docstring in scripts/fauxalysis_from_docking.py (i.e. with newlines)

JenkeScheen (commented):
Looks great, just some small formatting comments.

scripts/clean_results_csv.py
```diff
@@ -20,6 +27,10 @@ def get_args():
     parser.add_argument(
         "-o", "--output_dir", required=True, help="Path to output directory"
     )
+    parser.add_argument("-d", "--dimer", action="store_true", default=False)
+    parser.add_argument(
+        "-s", "--keep_structure_source", action="store_true", default=False
+    )
```
kaminow, line 33: I think it would be helpful to have a `help` field for both of these, just to make it a bit more obvious what they do

scripts/create_fauxalysis_subset_from_yaml.py
```diff
@@ -0,0 +1,84 @@
...
+    parser.add_argument(
+        "-d", "--fauxalysis_dir", required=True, help="Path to fauxalysis_dir."
+    )
+    parser.add_argument(
+        "-o", "--output_dir", required=True, help="Path to output dir"
+    )
+    parser.add_argument(
+        "-y",
+        "--yaml_file",
+        default="../data/luttens2022ultralarge.yaml",
```
kaminow, line 21: another hardcoded path (see Iván's other comment)

scripts/fauxalysis_from_docking.py
```diff
@@ -1,37 +1,229 @@
...
         "-f",
         "--fragalysis_dir",
-        default=False,
+        default=None,
         help="Path to fragalysis results directory.",
     )
+    parser.add_argument(
+        "-y",
+        "--fragalysis_yaml",
+        default="../data/cmpd_to_frag.yaml",
```
kaminow, line 47: hardcoded

scripts/fauxalysis_from_docking.py
```diff
@@ -1,37 +1,229 @@
...
+
+        ## Set ligand title, useful for the combined sdf file
+        lig.SetTitle(f"{compound_id}_{best_str}")
+
+        ## Give ligand atoms their own chain "L" and set the resname to "LIG"
+        for atom in lig.GetAtoms():
+            residue = oechem.OEAtomGetResidue(atom)
+            residue.SetChainID("L")
+            residue.SetName("LIG")
+            oechem.OEAtomSetResidue(atom, residue)
```
kaminow, line 162: I think if we add `residue.SetHetAtom(True)` before this line it should take care of #66

kaminow (changes requested):
a couple of minor changes, but overall looks good! thanks!

scripts/clean_results_csv.py
apayne97: eh lets keep it so that we don't change it again?

scripts/fauxalysis_from_docking.py
```diff
@@ -1,37 +1,234 @@
...
+        save_openeye_pdb(prot, f"{compound_out_dir}/{best_str}_apo.pdb")
+
+        ## Combine protein and ligand and save
+        oechem.OEAddMols(prot, lig)
+        save_openeye_pdb(prot, f"{compound_out_dir}/{compound_id}_bound.pdb")
+
+        ## Remove Hs from lig and save SDF file
+        for a in lig.GetAtoms():
+            if a.GetAtomicNum() == 1:
+                lig.DeleteAtom(a)
```
apayne97, line 179: @kaminow I copied this over from your script, is there a reason you did this? Because I think this is what is causing John's error here:
https://asapdiscovery.slack.com/archives/C03QD9UM9TP/p1665812798018309

kaminow: oh interesting, yeah I did it because the SDF files in fragalysis don't have Hs in them, but there's probably a better way of filtering atoms to write out than just deleting them

apayne97: in this case I think the fragalysis SDFs are wrong and we should write out Hs

scripts/clean_results_csv.py
```diff
@@ -1,3 +1,14 @@
+"""
+This script cleans up the CSV currently (as of 2022.10.14) generated by `hallucinate_fragalysis.py` into
+a format the following scripts (such as fauxalysis_from_docking.py and plot_docking_results.py)
```
ijpulidos, line 3: I believe the sentence is not finished here :)

scripts/write_cmpd_to_frag_dict.py
```diff
@@ -0,0 +1,69 @@
...
+The goals is to create a saved dictionary mapping the compound_ID to the fragalysis structure we are using.
+i.e. `AAR-POS-0daf6b7e-1: Mpro-x1311`
+The input is the compound_tracker.csv file, the output is a yaml file (default is in data/cmpd_to_frag.yaml).
+This generates a required input for the `fauxalysis_from_docking.py` script.
+Example Usage:
+    python
```
ijpulidos, line 7: I believe this is missing the script name?
```suggestion
    python write_cmpd_to_frag_dict.py
```

ijpulidos (commented):
Minor comments. Looking good!

scripts/fauxalysis_from_docking.py
```diff
@@ -1,39 +1,242 @@
...
+
+        ## Clean up PDB info by re-perceiving, perserving chain ID, residue number, and residue name
+        preserve = (
+            oechem.OEPreserveResInfo_ChainID
+            | oechem.OEPreserveResInfo_ResidueNumber
+            | oechem.OEPreserveResInfo_ResidueName
+        )
+        oechem.OEPerceiveResidues(prot, preserve)
+
+        save_openeye_pdb(prot, f"{compound_out_dir}/{compound_id}_bound.pdb")
```
apayne97, line 184: This addresses some of these issues:
#75

apayne97: I think this code should be moved to its own API somewhere (maybe in openeye or in modeling).

JenkeScheen (approved):

scripts/fauxalysis_from_docking.py
apayne97: done in https://github.com/choderalab/covid-moonshot-ml/pull/65/commits/f9a8a8b8e5b8dc8c418d9971f86bd6fdc30ee2b0

scripts/fauxalysis_from_docking.py
```diff
@@ -42,26 +233,79 @@ def main():
...
+            cmpd_to_frag_dict = yaml.safe_load(f)
+    else:
+        cmpd_to_frag_dict = None
+    write_fragalysis_output(
+        args.input_dir,
+        args.output_dir,
+        best_structure_dict,
+        args.fragalysis_dir,
+        cmpd_to_frag_dict,
+    )
```
ijpulidos, line 308: We should use the keys here for the keyword arguments just such that if things change in the signature of the function (order or names, etc.) we can still expect the same behavior.
```suggestion
    write_fragalysis_output(
        args.input_dir,
        args.output_dir,
        best_structure_dict,
        frag_dir=args.fragalysis_dir,
        cmpd_to_frag_dict=cmpd_to_frag_dict,
    )
```
Maybe we should use keywords for *all* of the arguments here (not only the optional ones, python allows this), but whatever feels more comfortable with you in that case.

scripts/write_cmpd_to_frag_dict.py
```diff
@@ -0,0 +1,69 @@
...
+    parse_fragalysis_data,
+)
+
+
+def get_args():
+    parser = argparse.ArgumentParser(description="")
+
+    parser.add_argument(
+        "-f",
+        "--frag_csv",
```
ijpulidos, line 27: Not a blocker. I would just call it `input` here and adapt the arguments as needed. The help should say whatever format it expects or if it's related to fragalysis, etc.

ijpulidos (commented):
Looks great! Just a few minor comments, some of them are not blockers.

scripts/fauxalysis_from_docking.py
```diff
@@ -1,39 +1,230 @@
...
-    f"{os.path.dirname(os.path.dirname(os.path.abspath(__file__)))}"
-)
 from covid_moonshot_ml.docking.analysis import DockingResults
+from covid_moonshot_ml.datasets.utils import (
+    load_openeye_pdb,
+    load_openeye_sdf,
+    save_openeye_pdb,
+    save_openeye_sdf,
+)
+from covid_moonshot_ml.data.openeye import save_openeye_design_unit
```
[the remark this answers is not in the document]
apayne97: line 16 and 17 make this possible! you just add the directory containing the repo to the sys_path

I agree we should have a common way of importing but I think we should wait until the repo re-factor to fix this

covid_moonshot_ml/data/openeye.py
```diff
@@ -29,6 +29,31 @@ def load_openeye_sdf(sdf_fn):
...
+    """
+    ifs = oechem.oemolistream()
+    ifs.SetFlavor(
+        oechem.OEFormat_SDF,
+        oechem.OEIFlavor_SDF_Default,
+    )
+    cmpd_list = []
+    if ifs.open(sdf_fn):
+        for mol in ifs.GetOEGraphMols():
+            cmpd_list.append(mol.CreateCopy())
```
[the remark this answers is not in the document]
apayne97: > Not a blocker. Is there a reason why we need to create a copy and not just appending the mol object itself?
it doesn't work! I *think* it points to an object that gets overwritten by `ifs.GetOEGraphMols()` at each iteration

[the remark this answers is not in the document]
apayne97: >Also, just to make sure we have thought about this. Openeye's doc says the destructor should call the .close() method on the file stream. But maybe we want to call it manually after reading the molecules just to be on the safe side? Not a blocker.

done!

scripts/write_cmpd_to_frag_dict.py
apayne97: since this script is specific to generating the `COMPOUND_ID: FRAGALYSIS_DATASET` dictionary from the fragalysis csv file I this it makes sense, esp since we're using `parse_fragalysis_data` here to process the csv

scripts/fauxalysis_from_docking.py
```diff
@@ -1,39 +1,230 @@
...
 
 
 def get_args():
     parser = argparse.ArgumentParser(description="")
 
     ## Input arguments
     parser.add_argument(
-        "-i",
+        "-c",
```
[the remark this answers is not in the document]
apayne97: fixed

scripts/create_fauxalysis_subset_from_yaml.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            in_file = os.path.join(args.fauxalysis_dir, full_path)
+            print(f"Copying {in_file}")
+            shutil.copy2(in_file, args.output_dir)
+
+    sdf_list = []
+    for flag in pdb_dict.keys():
+
+        ## if our flag of interest is in any complex, include it!
+        ## TODO: replace with REGEX?
```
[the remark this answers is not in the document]
apayne97: fair enough! just acknowledging that it could be improved in the future

ijpulidos (approved):
LGTM!

PR #81, opened 2022-11-01
Add 2d graph models

kaminow:
## Description
Add a new class for a `GAT` model, update the `train.py` script to include that model, and add `GraphDataset` class for holding a 2D graph data object.
Also add code to interface with [Weights & Biases](https://wandb.ai/) to track model training. This code only runs if `wandb` is installed, so everything else should still work even if you don't have it set up.

## Todos
  - [x] Add `GAT` model
  - [x] Add `GraphDataset` class in `covid_moonshot_ml.data.dataset` as a container for 2D graphs
  - [x] Add `wandb` code
  - [x] Other stuff from #84

## Status
- [x] Ready to go


scripts/train.py
```diff
@@ -1,8 +1,10 @@
 import argparse
```
ijpulidos, line 1: We probably want a docstring here saying in a few words what this script does (and outputs) and with what kind of input data.

ijpulidos: Also an example of how it should be called or used.

scripts/train.py
```diff
@@ -475,6 +664,30 @@ def main():
         val_loss = None
         test_loss = None
 
+    ## Update experiment configuration
+    exp_configure.update({"start_epoch": start_epoch})
+
+    ## Try and start wandb
+    try:
+        import wandb
```
ijpulidos, line 672: I think we would benefit from having this logic in a function and maybe it is better to have a CLI argument handle if we want to use `wandb` or not, defaulting to whatever is the expected sane behavior. Does that sound reasonable?

ijpulidos (commented):
This is looking great! I'd like to eventually know how to use this :) Just a few comments and minor corrections.

scripts/train.py
kaminow: resolved in 778aee4

JenkeScheen:
Are there any blockers to this PR or can it be marked as ready to go?

kaminow:
I'm still testing some stuff out, but I think it should be ready for review today/tomorrow

kaminow:
Everything should be ready to merge now! @yuanqing-wang if you could take a look at the `covid_moonshot_ml.utils.train` function to make sure I haven't done anything stupid that would be super helpful!

covid_moonshot_ml/data/dataset.py
```diff
@@ -21,40 +29,236 @@ def __init__(self, str_fns, compounds, ignore_h=True, lig_resn="LIG"):
...
+        self.structures = []
+        for i, (compound, struct) in enumerate(zip(compounds, all_structures)):
+            try:
+                self.compounds[compound].append(i)
+            except KeyError:
+                self.compounds[compound] = [i]
+            self.structures.append(struct)
+
+    @staticmethod
+    def _load_structure(fn, compound, ignore_h, lig_resn, extra_dict):
```
JenkeScheen, line 76: Insert docstring for this fn

covid_moonshot_ml/data/dataset.py
```diff
@@ -21,40 +29,236 @@ def __init__(self, str_fns, compounds, ignore_h=True, lig_resn="LIG"):
...
+            return (compounds, str_list)
+        else:
+            return (compounds[0], str_list[0])
+
+    def __iter__(self):
+        for s in self.structures:
+            yield (s["compound"], s)
+
+
+### TODO before PR: fix compound_id_dict here to match with new format
```
JenkeScheen, line 181: is this completed in the current version? if so can remove TODO

JenkeScheen: if not, would this be a blocker for merge?

JenkeScheen (approved):
LGTM, just some small comments. Nice work!

covid_moonshot_ml/data/dataset.py
kaminow: added in [51b4b38](https://github.com/choderalab/covid-moonshot-ml/pull/81/commits/51b4b386c8df7364c871a0c7e73e602f94bda601)

covid_moonshot_ml/data/dataset.py
kaminow: fixed now

JenkeScheen (approved):

Issue #41, opened 2022-09-09
Save sufficient conda environments with intelligent names


apayne97:
# To-Do
### Conclusion: Use this version to save docking envs
`conda env export --no-builds`

### 1. This one is in a nice format for figuring out exactly where the packages are from (although I guess its a csv, not a yaml file) 
- https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/devtools/conda-envs/mers-docking-clean.yaml
```
# packages in environment at /Users/alexpayne/opt/anaconda3/envs/mers-docking:
#
# Name                    Version                   Build  Channel
_py-xgboost-mutex         2.0                       cpu_0    conda-forge
abseil-cpp                20211102.0           h96cf925_1    conda-forge
absl-py                   1.1.0              pyhd8ed1ab_0    conda-forge
aiohttp                   3.8.1            py39h63b48b0_1    conda-forge
...
```
- [ ] change file extension

### 2. This one is much easier to read and is the most minimal
- https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/devtools/conda-envs/mers-docking2.yaml
- `conda env export --from-history > mers-docking2.yaml`
```
name: mers-docking2
channels:
  - schrodinger
  - salilab
  - openeye
  - conda-forge
  - http://conda.anaconda.org/omnia
  - defaults
dependencies:
  - biopython
  - pydantic
  - python=3.8
  - pandas
  - pymol
  - pyyaml
...
```
### 3. This one also includes the pip dependencies and is the most complete
- https://github.com/choderalab/covid-moonshot-ml/blob/hallucinate-fragalysis/devtools/conda-envs/mers-docking2_complete.yaml
- `conda env export > mers-docking2_complete.yaml`
```
name: mers-docking2
channels:
  - anaconda
  - schrodinger
  - salilab
  - openeye
  - conda-forge
  - http://conda.anaconda.org/omnia
  - defaults
dependencies:
  - ambertools=22.0=py38hd6424c8_0
  - appdirs=1.4.4=pyh9f0ad1d_0
  - appnope=0.1.3=pyhd8ed1ab_0
  - argon2-cffi=21.3.0=pyhd8ed1ab_0
  - argon2-cffi-bindings=21.2.0=py38hed1de0f_2
  - arpack=3.7.0=hefb7bc6_2
  - asttokens=2.0.5=pyhd8ed1ab_0
[...]

PR #94, opened 2022-11-30
Remove kinoml and rdkit dependencies

apayne97:
## Description
We aren't really using kinoml or rdkit anymore and they make the environment unnecessarily bloated.

## Todos
  - [x] remove kinoml dependencies
  - [x] #41 add new docking environment
 
### specific code to re-write
  - [x] remove deprecated `run_docking.py` script that relied on kinoml (has since been replaced with `run_docking_oe.py`
  - [x] add pdb downloading functions to replace kinoml usage

## Tabled for later
- [ ] removing rdkit
- [ ] `find_all_mcs.py` uses `data.utils.load_exp_from_sdf` which currently depends on rdkit

## Questions
- [x] for jupyter notebooks that aren't tutorials but just were times we wanted to explore some things, should we go back and retroactively fix them or just let them get deprecated over time? maybe the simple thing is just to include in the repo what environment was used at the time
- [x] we use rdkit in more places, and I think it makes more sense as a package to keep. not sure it's worth going back to rewrite stuff in order to remove the dependency?

## Status
- [x] Ready to go


JenkeScheen (approved):
LGTM. Agreed with keeping the RDKit dependency for now, personally I'm much more familiar with that library than with OE so I'll probably be using it for quick scripting for a while longer.

PR #98, opened 2022-12-02
Converted `asapdiscovery` into implicit namespace package

dotsdl:
## Description
This repo features the core `asapdiscovery` Python package, but this namespace structure allows for other packages, e.g. `asapdiscovery-foo` to be created later with import semantics as `import asapdiscovery.foo` that just work.

For comparison, this is exactly the approach taken by OpenFF for `openff.toolkit`, `openff.evaluator`, etc., which appear under the same `openff` namespace, but are developed in separate repos and also packaged separately.

## Status
- [x] Ready to go


dotsdl:
Looks like versioneer may give us headaches with this approach. Since we don't have a single `_version.py` file, but instead one for each submodule, we would need some way to specify these [here](https://github.com/choderalab/covid-moonshot-ml/blob/d14e6ebf69ed357f675373dc2e11949ceef4cdc6/setup.cfg#L25-L26). Not sure if this is possible.

ijpulidos:
Yes, it seems like versioneer is not meant to be used this way. Maybe we should reconsider having different repos for each of the modules?

ijpulidos:
via @mikemhenry this might be helpful https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

dotsdl:
Okay, got this working! We're also now using `versioningit`, which handles versions based on git tags/refs in the same way as versioneer but without versioneer's ugliness. Works just fine with namespace packaging approach we've done here.

I need:
- an email address to use for the authors
- current dependencies for each of our packages

dotsdl:
Need to fix all the tests yet. @mikemhenry can I get feedback on the approach we've taken here; as in, are you happy with it?

dotsdl:
CI should now work; adjusted to conform to the new layout.

dotsdl:
I'm happy to convert the `ad-*` directory names over to full `asapdiscovery-*` if that's preferable.

dotsdl:
Went ahead with renaming `ad-*` to `asapdiscovery-*` to be explicit.

dotsdl:
@jchodera is there an email address for asapdiscovery we should use for our authors here?

dotsdl:
@kaminow and @apayne97 : can you give me short descriptions of each package?

dotsdl:
@ijpulidos I believe you had concerns with the structure; can you raise those here?

ijpulidos:
@dotsdl Yes! Apologies for the delay with this. I see the current structure is something like the following:
```
.
├── asapdiscovery-data
│   ├── asapdiscovery
│   │   └── data
│   │       ├──...
│   ├── ...
├── asapdiscovery-dataviz
│   ├── asapdiscovery
│   │   └── dataviz
│   │       ├── ...
│   ├── ...
├── asapdiscovery-docking
│   ├── asapdiscovery
│   │   └── docking
│   │       ├── ...
│   ├── ...
├── asapdiscovery-ml
│   ├── asapdiscovery
│   │   └── ml
│   │       ├── ...
│   ├── ...
├── ...
```

Is there a reason why it couldn't be as follows? What would break?
```
.
├── asapdiscovery
│   │   └── data
│   │       ├──... (each subdir here with their own pyproject.toml, MANIFEST.in, etc.)
│   │   └── dataviz
│   │       ├── ...
│   │   └── docking
│   │       ├── ...
│   │   └── ml
│   │       ├── ...
│   ├── ...
├── ...
```

I realize that it doesn't follow the structure suggested in https://packaging.python.org/en/latest/guides/packaging-namespace-packages/

I'm just trying to understand why it couldn't be that way (which seems more intuitive/compact), is there something special here about having the `asapdiscovery-subpackage` as top dir here? I hope I am making myself clear.

dotsdl:
@ijpulidos thanks for clarifying! The second scenario looks more compact, but inside of each `...` is also the full module contents sitting alongside the `pyproject.toml`, `MANIFEST.in`, etc. Although I'm not certain this *wouldn't* work, it definitely falls outside what packaging tools expect; it would mix packaging bits with the Python module contents itself.

So, I don't think it would work in practice due to these issues. The important aspect about the first scenario (which we've implemented here) is that the top-level directory for each package includes the packaging bits, with `asapdiscovery` on down the actual Python module contents/structure to be packaged.

ijpulidos:
> Does this sufficiently address your concerns?

 @dotsdl yes! Thanks for the detailed answer and the clarification.

dotsdl:
@ijpulidos please review and approve so we can proceed to merge.

ijpulidos (approved):
Thanks a lot for all the work on this and solving my doubts, it looks great! I just want to add that we soon will be having a new subpackage `asapdiscovery-simulation`. I think it's better if we add the file structure skeleton for it in this PR.

I'm approving it since I don't think this is a blocker, it would just be convenient to have right now.

apayne97:
@dotsdl There will also be `modeling` and `simulation` packages!
Descriptions:
- modeling: "API and scripts for preparing protein and ligand models for downstream applications"
- simulation: "API and scripts for running molecular dynamics simulations and free energy calculations"

dotsdl:
Done! Also, @kaminow I've put your email in place for contact as before; we can always change this later if something else is more appropriate.

PR #101, opened 2022-12-07
Fix duped load funcs

kaminow:
## Description
Small PR to remove some duplicated functions between `asapdiscovery.data.openeye` and `asapdiscovery.data.utils` and fix related import statements.

## Status
- [x] Ready to go


apayne97:
@kaminow I had meant to fix this already, but I think we should have the openeye stuff in the openeye module? or else just get rid of that module

kaminow:
ah yeah I think moving the functions into the `openeye` module probably makes more sense

asapdiscovery/data/fragalysis.py
```diff
@@ -66,6 +66,8 @@ def parse_xtal(x_fn, x_dir):
     x_dir : str
         Path to directory containing directories with crystal structure PDB
         files
+    p_only : bool, default=True
+        Whether to filter to only -P* files in fragalysis
```
JenkeScheen, line 70: a quick explanation of what -P* files are would be helpful here

JenkeScheen (approved):
LGTM!

asapdiscovery/data/fragalysis.py
kaminow: new comment in [d36aa85](https://github.com/choderalab/covid-moonshot-ml/pull/101/commits/d36aa85956f6936d00eff149accf7da0e8bac400) that I think should help

PR #103, opened 2022-12-08
Remove extra ligands

apayne97:
## Description
We've had a hard time tracking down when OpenEye grabs the ligand from chain A vs chain B to use to create the receptor grid. In order to do this explicitly, @kaminow has written a function to first remove the ligand from the chain we don't need.

## Todos
  - [x] remove ligands from chains that aren't in `xtal.active_site_chain`
  - [x] confirm that resulting docked poses are aligned with the ligands aligned to each other

## Status
- [x] Ready to go

### DIFF
--- a/asapdiscovery/docking/modeling.py
+++ b/asapdiscovery/docking/modeling.py
@@ -573,3 +573,46 @@ def build_dimer_from_monomer(prot):
 
     print(all_chain_ids)
     return prot
+
+
+def remove_extra_ligands(mol, lig_chain=None):
+    """
+    Remove extra ligands from a molecule. Useful in the case where a complex
+    crystal structure has two copies of the ligand, but we only want one. If
+    `lig_chain` is not specified, we'll automatically select the first chain
+    (as sorted alphabetically) to be the copy to keep.
+
+    Creates a copy of the input molecule, so input molecule is not modified.
+
+    Parameters
+    ----------
+    mol : oechem.OEMolBase
+        Complex molecule.
+    lig_chain : str, optional
+        Ligand chain ID to keep.
+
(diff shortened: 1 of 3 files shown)


asapdiscovery/docking/scripts/prep_proteins.py
```diff
@@ -122,6 +123,11 @@ def prep_mp(
             initial_prot, res_list, xtal.protein_chains
         )
 
+    ## Delete extra copies of ligand in the complex
+    initial_prot = remove_extra_ligands(
+        initial_prot, lig_chain=xtal.active_site_chain
+    )
+
```
apayne97, line 130: i think in the future we'll need to add this as an option or something for the other structures but that's for a future PR to fix

apayne97:
image
image
image
image
image

asapdiscovery/docking/modeling.py
```diff
@@ -573,3 +573,46 @@ def build_dimer_from_monomer(prot):
...
+                for a in mol.GetAtoms(all_lig_filter)
+            }
+        )[0]
+
+    ## Copy molecule and delete all lig atoms that don't have the desired chain
+    mol_copy = mol.CreateCopy()
+    for a in mol_copy.GetAtoms(all_lig_filter):
+        if oechem.OEAtomGetResidue(a).GetExtChainID() != lig_chain:
+            mol_copy.DeleteAtom(a)
+    return mol_copy
```
JenkeScheen, line 618: for debugging purposes it would be super helpful if you also return the deleted molecules here

JenkeScheen (approved):
LGTM!

PR #108, opened 2022-12-15
take SMILES of last conformer instead of taking list of SMILES

JenkeScheen:
## Description
In `run_docking_oe.py`, SMILES strings were not written to `results_df` correctly, resulting in empty fields in the pickled output files. 

## Todos
  - [x] fix SMILES field output in pickled docking outputs. (instead of iterating over conformers and appending SMILES to a list, just take the SMILES of the last conformer)

## Status
- [x] Ready to go

### DIFF
--- a/asapdiscovery/docking/scripts/run_docking_oe.py
+++ b/asapdiscovery/docking/scripts/run_docking_oe.py
@@ -148,7 +148,7 @@ def mp_func(out_dir, lig_name, du_name, *args, **kwargs):
         posit_probs = []
         posit_methods = []
         chemgauss_scores = []
-        smiles = []
+
         for conf in posed_mol.GetConfs():
             rmsds.append(
                 float(oechem.OEGetSDData(conf, f"Docking_{docking_id}_RMSD"))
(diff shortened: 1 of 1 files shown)


kaminow (approved):
made one small addition, but otherwise looks good. thanks for catching this!

apayne97:
Haven't tested it but it looks good to me!

apayne97 (approved):

asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -173,7 +173,7 @@ def mp_func(out_dir, lig_name, du_name, *args, **kwargs):
         posit_methods = [""]
         chemgauss_scores = [-1.0]
         clash = -1
-        smiles = ["None"]
+        smiles = "None"
```
ijpulidos, line 176: Is there a special reason why we want a string here instead of the python `None` built-in?

ijpulidos (commented):
Looks good. Just some minor comment.

asapdiscovery/docking/scripts/run_docking_oe.py
JenkeScheen: This is because these values on this `else` statement get written to a `.csv` on failure - I think we would prefer to have a '..,None,..' in the `.csv` rather than a '..,,..'. Would be worth revisiting if these variables get picked up in future development!

PR #127, opened 2023-01-22
Simplify and clean up docking environments

apayne97:
## Description
I had missed a commit in the previous PR in which I cleaned up the various docking environments and added one with kinoml removed.


apayne97:
I think maybe because of the 'squash and merge' its keeping the changes from the earlier commits to this branch. So instead I made a new branch and will PR here instead #128.

PR #129, opened 2023-02-01
Add code to simulate and analyze docked or prepared structures

apayne97:
## Description
NOT READY FOR MERGING.
Adding / updating code to simulate and analyze docked / prepared structures. 
Rotation project of @kendallrlemons. 

## Todos
  - [x] add initial openmm-setup script
  - [x] add cluster script
  - [x] run first round of simulation to see what pain points there are
  - [x] add code to use openff forcefields for ligand parameterization
  - [x] troubleshoot slow minimization
  - [x] add virtual bond code
  - [ ] write out xtc instead of DCD
  - [ ] make new version of script (with new name) that can load in any structures / ligands and has dynamically generated file names / output directory

## Features to Add (in no particular order)
  - [ ] use yaml file of simulation arguments instead of having them be hard coded
  - [x] use logger instead of print statements to be able to catch progress
- [ ] parse out some of the bits of this code into an api
- [ ] create simplified simulation environment and add to environment options for repo
- [ ] integrate ideas from #139 

## Questions
- [ ] 

## Status
- [ ] Ready to go


asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+print('Simulating...')
+simulation.reporters.append(PDBReporter(os.path.join(args.output_dir,'output(P0009).pdb'), 10))
+simulation.reporters.append(DCDReporter(os.path.join(args.output_dir,'output(P0009).dcd'), 10))
+simulation.currentStep = 0
+simulation.step(steps)
+
+
+print('Saving...')
+positions = simulation.context.getState(getPositions=True).getPositions()
+PDBFile.writeFile(simulation.topology, positions, open(os.path.join(args.output_dir,'output(P0009*).pdb'), 'w'))
```
apayne97, line 117: I think its more standard / cleaner to use a `with open:` block here for the filepath to write the PDB file to

apayne97 (commented):
Looks ok to start but we will want to add some features!

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+# Input Files
+
+pdb = PDBFile(args.input_pdb_path)
+forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
+
+# System Configuration
+
+nonbondedMethod = CutoffPeriodic
```
apayne97, line 53: The standard is to use PME right now so let's use that instead

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#BSUB -J run_single_simulation
+#BSUB -oo log_files/prep_full_mpro_AP_test.out
+#BSUB -e log_files/prep_full_mpro_AP_test.stderr
```
apayne97, line 4: change this names to `log_files/MProP0009_Sub.out` etc

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
```diff
@@ -0,0 +1,22 @@
...
+mkdir -p /data/chodera/asap-datasets/prepped_mpro_P0009/
+
+source ~/.bashrc
+conda activate docking
+
+python /data/chodera/lemonsk/covid-moonshot-ml/asapdiscovery/simulation/scripts/MproP0009.py \
+
+-i /data/chodera/asap-datasets/full_frag_prepped_mpro_12_2022/Mpro-P0009_0A_MAT-POS-f2460aef-1/prepped_receptor_0.pdb \
+-o /data/chodera/lemonsk/default_out \
+
```
apayne97, line 18: the backslash is commenting out the newline, but it won't comment lines 15 and 18, so this will throw an error. remove lines 15 and 18. (the empty ones)

apayne97 (commented):
Two comments that will require changing but after that should be good to go!

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+    step=True, speed=True, progress=True, potentialEnergy=True, temperature=True, separator='\t')
+checkpointReporter = CheckpointReporter(os.path.join(args.output_dir,'checkpoint.chk'), 10000)
+
+# Prepare the Simulation
+
+print('Building system...')
+topology = pdb.topology
+positions = pdb.positions
+modeller = Modeller(pdb.topology, pdb.positions)
+modeller.addSolvent(forcefield, boxSize=Vec3(5.0, 3.5, 3.5)*nanometers, model='tip3p')
```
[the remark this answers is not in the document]
kendallrlemons: Will fix the padding distance, just went with the default parameters given by OpenMM-Setup. Should have paid closer attention to that piece!

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
+
+# System Configuration
+
+nonbondedMethod = CutoffPeriodic
+nonbondedCutoff = 1.0*nanometers
+constraints = HBonds
+rigidWater = True
+constraintTolerance = 0.000001
+hydrogenMass = 1.5*amu
```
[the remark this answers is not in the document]
kendallrlemons: Not sure what happened here either, another OpenMM-Setup input!

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+import openmm
+import argparse
+
+
+
+# This script was generated by OpenMM-Setup on 2023-01-17.
+
+from openmm import *
+from openmm.app import *
+from openmm.unit import *
```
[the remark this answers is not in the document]
kendallrlemons: Will circle back on this and see what's being called directly

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#BSUB -J run_single_simulation
+#BSUB -oo log_files/MProP0009_Sub.out
+#BSUB -e log_files/MProP0009_Sub.stderr
+#BSUB -n 1
+#BSUB -R rusage[mem=4]
+#BSUB -q gpuqueue
+#BSUB -W 12:00
+mkdir -p /data/chodera/asap-datasets/prepped_mpro_P0009/
```
apayne97, line 9: You should change this to make a directory in your own `/data/chodera` folder!

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
```diff
@@ -0,0 +1,20 @@
...
+
+source ~/.bashrc
+conda activate docking
+
+python /data/chodera/lemonsk/covid-moonshot-ml/asapdiscovery/simulation/scripts/MproP0009.py \
+-i /data/chodera/asap-datasets/full_frag_prepped_mpro_12_2022/Mpro-P0009_0A_MAT-POS-f2460aef-1/prepped_receptor_0.pdb \
+-o /data/chodera/lemonsk/default_out \
+--include_non_Pseries
```
apayne97, line 17: you don't need this line since it isn't an argument in your script

apayne97 (commented):
Just two last comments on the submission script! Then you should run it!

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
kendallrlemons: Ah, I thought I made that distinction, will correct that!

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
kendallrlemons: Line 17?

apayne97: yes! sorry only meant to highlight that one

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+#BSUB -J run_single_simulation
+#BSUB -oo log_files/MProP0009_Sub.out
+#BSUB -e log_files/MProP0009_Sub.stderr
+#BSUB -n 1
+#BSUB -R rusage[mem=4]
+#BSUB -q gpuqueue
+#BSUB -W 12:00
+mkdir -p /data/chodera/lemonsk/asap-datasets/prepped_mpro_P0009/
```
apayne97, line 9: This directory and the directory in line 16 should be the same, right? Since that's what you're actually writing out to. The other thing, which is potentially better, would be to delete this line and just have in your script a little line that creates the directory passed to `-o` if it doesn't exist.

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,119 @@
...
+    parser.add_argument(
+        "-o", 
+        "--output_dir",
+        type = str,
+        default = "/data/chodera/lemonsk/default_out",
+        help="Output simulation directory."
+    )
+    args = parser.parse_args()
+    return args
+
```
apayne97, line 44: You need to add a line:
```
args = get_args()
```
in order for `args.input_pdb_path` to actually work

asapdiscovery/simulation/scripts/MProP0009_Sub.sh
kendallrlemons: I see, as another alternative can I just go into my directory and just make this new directory altogether?

apayne97: sure! that can get annoying if you want to do that many times for many different things but it will work fine here!

asapdiscovery/simulation/scripts/MproP0009.py
apayne97: you could also just put everything in the `get_args` function in the main script. which is a bit simpler.

kendallrlemons (approved):

kendallrlemons (approved):

kendallrlemons (approved):

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -39,18 +32,67 @@ def get_args():
...
+forcefield_kwargs = {'constraints': app.HBonds, 'rigidWater': True, 'removeCMMotion': False, 'hydrogenMass': 4*unit.amu }
+periodic_forcefield_kwargs = {'nonbondedMethod': app.PME}
+system_generator = SystemGenerator(
+    forcefields=['amber/ff14SB.xml', 'amber/tip3p_standard.xml'],
+    small_molecule_forcefield='openff-1.3.1',
+    molecules=[ligand_mol], cache='cache.json',
+    forcefield_kwargs=forcefield_kwargs, periodic_forcefield_kwargs=periodic_forcefield_kwargs);
+
+# Use Modeller to combine the protein and ligand into a complex
+from openmm.app import PDBFile
```
apayne97, line 74: your imports are really messy!
You're importing something from openmm.app multiple times. I know John likes lazy imports but I think this could be the problem

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -39,18 +32,67 @@ def get_args():
...
 
-pdb = PDBFile(args.input_pdb_path)
+from openmm.app import ForceField
+#pdb = PDBFile(args.input_pdb_path)
 forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
 
 # System Configuration
 
+from openmm.app import *
+from openmm.unit import *
```
apayne97, line 94: I think the error is here. Your error message was
```
 File "/data/chodera/lemonsk/covid-moonshot-ml/asapdiscovery/simulation/scripts/MproP0009_2.py", line 137, in <module>
  modeller.addSolvent(forcefield, padding=0.9*nanometers, model='tip3p')
AttributeError: module 'openmm.app.modeller' has no attribute 'addSolvent'
```
And `modeller` shouldn't be a module, it should be an object.
But since you've used a star import after creating the modeller object, which overwrites the variable name (i.e. modeller is now a module name instead of an object name in the script namespace) later when you go to addSolvent, instead of using the object that you've already created, you're trying to call `openmm.app.modeller.addSolvent`, which doesn't exist. The function you'd like to use is `openmm.app.modeller.Modeller.addSolvent` which does exist.

http://docs.openmm.org/latest/api-python/generated/openmm.app.modeller.Modeller.html?highlight=openmm%20app%20modeller

https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/

asapdiscovery/simulation/scripts/MproP0009.py
kendallrlemons: Yep, my imports are very messy lol. I was still polishing but went ahead and pushed. I was initially trying to fix the star imports, but I was leaving a lot of things out hence the repeated imports from the different components. Yes, this needs to be cleaned up a ton, but at this moment, from what I can tell its not causing any errors

asapdiscovery/simulation/scripts/MproP0009.py
kendallrlemons: See, I know where you're coming from. I just haven't seen that before. Even Jenke had what I'm trying to do, but I'll try it and see. I just worry we have too many "modeller" objects out there, which could cause an issue? Maybe I'm being irrational here. But yes, no more star imports

apayne97: Right, this is Jenke's code:
```
from openmm.app import Modeller
modeller = Modeller(protein_pdb.topology, protein_pdb.positions)
# This next bit is black magic.
# Modeller needs topology and positions. Lots of trial and error found that this is what works to get these from
# an openforcefield Molecule object that was created from a RDKit molecule.
# The topology part is described in the openforcefield API but the positions part grabs the first (and only)
# conformer and passes it to Modeller. It works. Don't ask why!
modeller.add(ligand_mol.to_topology().to_openmm(), ligand_mol.conformers[0].to_openmm())

# We need to temporarily create a Context in order to identify molecules for adding virtual bonds
log.info(f':microscope:  Identifying molecules')
import openmm
integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
system = system_generator.create_system(modeller.topology, molecules=ligand_mol)
context = openmm.Context(system, integrator, openmm.Platform.getPlatformByName('Reference'))
molecules_atom_indices = context.getMolecules()
del context, integrator, system

# Solvate
log.info(':droplet:  Adding solvent...')
# we use the 'padding' option to define the periodic box. The PDB file does not contain any
# unit cell information so we just create a box that has a 9A padding around the complex.
modeller.addSolvent(system_generator.forcefield, model='tip3p', padding=9.0*unit.angstroms)
```

In between creating the modeller object and adding solvent there are no star imports, i.e. there's no time for the name of the object to get overwritten!

apayne97: The problem isn't that star imports are always bad and wrong, its just that in this specific case your object is getting overwritten which is why you see the error message that you do!

apayne97: Simple example.
The first time, it at least knows what the function is, where as the second time it doesn't!
```
modeller = Modeller(pdb.topology, pdb.positions)
modeller.addSolvent()
  File "/Users/alexpayne/opt/anaconda3/envs/asap-simulation/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3442, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-9-3b4dda05aca7>", line 1, in <module>
    modeller.addSolvent()
TypeError: Modeller.addSolvent() missing 1 required positional argument: 'forcefield'
from openmm.app import *
modeller.addSolvent()
  File "/Users/alexpayne/opt/anaconda3/envs/asap-simulation/lib/python3.11/site-packages/IPython/core/interactiveshell.py", line 3442, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-11-3b4dda05aca7>", line 1, in <module>
    modeller.addSolvent()
    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'openmm.app.modeller' has no attribute 'addSolvent'
```

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -39,18 +32,67 @@ def get_args():
...
+# The topology part is described in the openforcefield API but the positions part grabs the first (and only)
+# conformer and passes it to Modeller. It works. Don't ask why!
+modeller.add(ligand_mol.to_topology().to_openmm(), ligand_mol.conformers[0].to_openmm())
+
 # Input Files
 
-pdb = PDBFile(args.input_pdb_path)
+from openmm.app import ForceField
+#pdb = PDBFile(args.input_pdb_path)
 forcefield = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml')
```
apayne97, line 89: we don't need this line anymore since we should be using `system_generator.forcefield` as the forcefield!

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -80,11 +127,13 @@ def get_args():
...
+from openmm import LangevinMiddleIntegrator
+from openmm.app import *
 print('Building system...')
-topology = pdb.topology
-positions = pdb.positions
-modeller = Modeller(pdb.topology, pdb.positions)
+#topology = pdb.topology
+#positions = pdb.positions
+#modeller = Modeller(pdb.topology, pdb.positions)
 modeller.addSolvent(forcefield, padding=0.9*nanometers, model='tip3p')
```
apayne97, line 137: here we want to use system_generator.forcefield instead as well!

asapdiscovery/simulation/scripts/MproP0009.py
kendallrlemons: Yes, very true!

kendallrlemons (approved):

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,170 @@
...
+print('Equilibrating...')
+simulation.context.setVelocitiesToTemperature(temperature)
+simulation.step(equilibrationSteps)
+
+
+# Simulate
+
+print('Simulating...')
+simulation.reporters.append(PDBReporter(os.path.join(args.output_dir,'output(P0009).pdb'), 10))
+simulation.reporters.append(DCDReporter(os.path.join(args.output_dir,'output(P0009).dcd'), 10))
```
apayne97, line 161: Some things:
1. you have already made a dcdreporter in line 124, use that instead of remaking it here!
2. let's change the writing frequency back to a reasonable number (10-100 ps / frame, or whatever Jenke does)
3. We don't want to be using a PDBReporter and a DCDReporter. The PDBReporter will write out the trajectory as a multi-object pdb file, which is wildly data inefficient. DCD / XTC just keeps track of the positions (and velocities? I don't remember) but doesn't keep the topology information, which saves space. Then we can use the original or output PDBFile (on line 168) to load in the topology file for analyzing. 
4. Let's never use parentheses in file names. It's a headache to deal with. For this instead we can just do output_P0009.dcd. In the future we'll want to make this dynamic with something like an "output_prefix" or something so that the `P0009` part of the name isn't hard coded.

kendallrlemons: For some reason (just did another test run), the dcdreporter I created in line 124 is still not receiving any info. Yet the other one ( on line 161) is. I've been playing around with this for a while so still trying to figure it out. My log.txt file is consistently empty as well

asapdiscovery/simulation/scripts/MproP0009.py
```diff
@@ -0,0 +1,182 @@
...
+from openmm import MonteCarloBarostat
+from openmm import LangevinMiddleIntegrator
+from openmm.app import *
+print('Building system...')
+#topology = pdb.topology
+#positions = pdb.positions
+#modeller = Modeller(pdb.topology, pdb.positions)
+modeller.addSolvent(system_generator.forcefield, padding=0.9*unit.nanometers, model='tip3p')
+system = forcefield.createSystem(modeller.topology, nonbondedMethod=nonbondedMethod, nonbondedCutoff=nonbondedCutoff,
+    constraints=constraints, rigidWater=rigidWater, ewaldErrorTolerance=ewaldErrorTolerance, hydrogenMass=hydrogenMass)
```
apayne97, line 151: @kendallrlemons this overwrites the system created on line 92, which loses the virtual bond!

asapdiscovery/simulation/scripts/MProSimulation.py
```diff
@@ -0,0 +1,170 @@
...
+pressure = 1.0 * unit.atmospheres
+barostatInterval = 25
+nonbondedMethod = PME
+nonbondedCutoff = 1.0*unit.nanometers
+ewaldErrorTolerance = 0.000001
+constraints = HBonds
+rigidWater = True
+constraintTolerance = 0.000001
+hydrogenMass = 4.0*unit.amu
+platform = Platform.getPlatformByName('CUDA')
```
apayne97, line 55: This means that this script will fail to run if the GPU isn't able to use 'CUDA'! I would suggest making sure it fails quickly if so.

asapdiscovery/simulation/scripts/MProSim_Sub.sh
```diff
@@ -3,15 +3,18 @@
 #BSUB -oo log_files/MProSim_Sub.out
 #BSUB -e log_files/MProSim_Sub.stderr
 #BSUB -n 1
+#BSUB -n 2 -W 2:00 -q gpuqueue -gpu 'num=1' -R 'span[hosts=1]'
```
apayne97, line 6: Remember that some of these flags are redundant! This line can just be
`-gpu 'num=1'`

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        default = "/data/chodera/lemonsk/asap-datasets/openmm_setup_processed/prepped_receptor_0-processed.pdb",
+        help = "Path to PDB file to simulate"
+    )
+    parser.add_argument(
+        "-l", 
+        "--ligand",
+        type = str,
+        default = "/data/chodera/lemonsk/covid-moonshot-ml/asapdiscovery/simulation/tests/inputs/MAT-POS-f2460aef-1.sdf",
+        help="Ligand"
+    )
```
apayne97, line 84: Instead of these arguments, you should have strings for globbing! 
i.e
```
parser.add_argument(
        "-p", 
        "--protein_glob", 
        type = str,
        default = "*/*protein.pdb",
        help = "String to correctly glob protein file using input directory"
    )
    parser.add_argument(
        "-l", 
        "--ligand_glob",
        type = str,
        default = "*/*.sdf",
        help="String to correctly glob ligand file using input directory"
    )
```

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        "--nsteps",
+        type = int,
+        default = 12,
+        help = "Number of Steps"
+    )
+    parser.add_argument(
+        "-o",
+        "--output_dir",
+        type = str,
+        default = "/data/chodera/lemonsk/asap-datasets/MPro_Simulations/",
```
apayne97, line 96: I don't think this should be a default argument. If you forget to change it when you try to re-run it, it could overwrite previous results.

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+    logger.info(':CD: Saving...')
+    positions = simulation.context.getState(getPositions=True).getPositions()
+    PDBFile.writeFile(simulation.topology, positions, open(os.path.join(output_dir + save + '_simulation' + '_final.pdb'), 'w'))
+    simulation.saveState(os.path.join(output_dir,'output.xml'))
+    print('Done')
+
+    def main():
+        args = get_args()
+
+        protein = Path(args.protein)
```
apayne97, line 222: this line can be deleted now!

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        proteinpaths = protein.glob("*protein.pdb")
+        ligand = Path(args.ligand)
```
apayne97, line 224: this line can also be deleted

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+    positions = simulation.context.getState(getPositions=True).getPositions()
+    PDBFile.writeFile(simulation.topology, positions, open(os.path.join(output_dir + save + '_simulation' + '_final.pdb'), 'w'))
+    simulation.saveState(os.path.join(output_dir,'output.xml'))
+    print('Done')
+
+    def main():
+        args = get_args()
+
+        protein = Path(args.protein)
+        proteinpaths = protein.glob("*protein.pdb")
```
apayne97, line 223: so instead of this we want to do something like
```
input_dir = Path(args.input_dir)
proteinpaths = list(input_dir.glob(args.protein_glob))
ligandpaths = list(input_dir.glob(args.ligand_glob))
```

What you currently have will search for all files with a path like `path_to_protein/protein.pdb*protein.pdb` which won't work

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        "--output_dir",
+        type = str,
+        default = "/data/chodera/lemonsk/asap-datasets/MPro_Simulations/",
+        help = "Output simulation directory."
+    )
+    args = parser.parse_args()
+    return args
+
+
+def simulate_mpro(input_dir,protein,ligand,nsteps,output_dir):
```
apayne97, line 103: these arguments need to match the arguments you pass to mpro_args in `main`! So I think you can get rid of the `input_dir` argument here.

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        proteinpaths = protein.glob("*protein.pdb")
+        ligand = Path(args.ligand)
+        ligandpaths = ligand.glob("*.sdf")
+        output = Path(args.output_dir)
+        output.mkdir(exist_ok=True)
+
+        mpro_args = [
+            [proteinpaths, ligandpaths, output]
+            for proteinpaths, ligandpaths in input_dir
+        ]
```
apayne97, line 232: This for loop won't really work. The `Path.glob()` function will make a Python Generator object that we can convert to a list, then we can zip the lists and iterate through those.

We should first check that proteinpaths and ligandpaths are the same length.
Then we can do something like
```
simulate_args = [(protein, ligand, nsteps, output_dir) for protein, ligand in zip(proteinpaths, ligandpaths)]
```

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+        default = "/data/chodera/lemonsk/asap-datasets/MPro_Simulations/",
+        help = "Output simulation directory."
+    )
+    args = parser.parse_args()
+    return args
+
+
+def simulate_mpro(input_dir,protein,ligand,nsteps,output_dir):
+    ## Necessary for Reporters
+    save = re.search(r"Mpro-P+\d{4}", protein).group(0)
```
apayne97, line 105: This will fail on the non Pseries structures and will not capture the ligand information.
Instead we can get the save name by calling `protein.parent.name` which will get the name of the parent directory of the input protein file. 

https://docs.python.org/3/library/pathlib.html

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+    ligand_mol = Molecule(rdkitmolh)
+
+    # Initialize a SystemGenerator
+    logger.info(':factory: System Combination')
+    forcefield_kwargs = {'constraints': app.HBonds, 'rigidWater': True, 'removeCMMotion': False, 'hydrogenMass': 4*unit.amu }
+    periodic_forcefield_kwargs = {'nonbondedMethod': app.PME}
+    system_generator = SystemGenerator(
+        forcefields=['amber/ff14SB.xml', 'amber/tip3p_standard.xml'],
+        small_molecule_forcefield='openff-1.3.1',
+        molecules=[ligand_mol], cache='cache.json',
```
apayne97, line 135: you define simulations parameters at the beginning but then you don't use them here. You should use them!

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+    # an openforcefield Molecule object that was created from a RDKit molecule.
+    # The topology part is described in the openforcefield API but the positions part grabs the first (and only)
+    # conformer and passes it to Modeller. It works. Don't ask why!
+    modeller.add(ligand_mol.to_topology().to_openmm(), ligand_mol.conformers[0].to_openmm())
+
+    # We need to temporarily create a Context in order to identify molecules for adding virtual bonds
+    logger.info(f':microscope:  Identifying molecules')
+    integrator = openmm.VerletIntegrator(1*unit.femtoseconds)
+    system = system_generator.create_system(modeller.topology, molecules=ligand_mol)
+    context = openmm.Context(system, integrator, openmm.Platform.getPlatformByName('CUDA'))
```
apayne97, line 154: you should use your platform variable here

asapdiscovery/simulation/scripts/MProFunction.py
```diff
@@ -0,0 +1,239 @@
...
+pressure = 1.0 * unit.atmospheres
+barostatInterval = 25
+nonbondedMethod = PME
+nonbondedCutoff = 1.0*unit.nanometers
+ewaldErrorTolerance = 0.000001
+constraints = HBonds
+rigidWater = True
+constraintTolerance = 0.000001
+hydrogenMass = 4.0*unit.amu
+platform = Platform.getPlatformByName('CUDA')
```
apayne97, line 57: does this fail at this step if CUDA isn't found? (it would be good if so)

apayne97 (commented):
Good start! You can test if this works by starting with an input directory and glob strings that point to a single protein / ligand!

I appreciate that the * imports have been removed!

asapdiscovery/simulation/scripts/MProFunction.py
kendallrlemons: Yep! But not as early I'd like it to, I might run again and see where it fails again

asapdiscovery/simulation/scripts/MProFunction.py
kendallrlemons: Ah I see. This bit is from Jenke's script, I will go in and remove some fluff

asapdiscovery/simulation/scripts/MProSim_Sub.sh
kendallrlemons: This is old I, need to push the new shell script, but yes - you're right!

asapdiscovery/simulation/scripts/MProFunction.py
kendallrlemons: Got it!

asapdiscovery/simulation/scripts/MProFunction.py
kendallrlemons: Got it!

asapdiscovery/simulation/scripts/MProFunction.py
kendallrlemons: Hmm, will need to follow up here tomorrow to get some clarity

PR #130, opened 2023-02-02
Improve CDD filtering

kaminow:
## Description
Improve the filtering of data downloaded from CDD, using code from [a perses example](https://github.com/choderalab/perses/blob/moonshot-examples/examples/moonshot-sulfonamides/molecules/filter-cdd-export.py).

## Todos
  - [x] Add salt stripping
  - [x] Add binding affinity calculations
  - [x] Update `cdd_to_schema` to accommodate the new format
  - [x] Update `cdd_to_schema_pair` to accommodate the new format

## Status
- [x] Ready to go


JenkeScheen (commented):
LGTM - I think it would be very helpful to include in this script a list of minimally required columns in the CSV - should be okay to just have it as a docstring somewhere. Potentially it would be a great test just to make sure the user has included the correct collection of columns.

asapdiscovery/data/utils.py
```diff
@@ -207,7 +207,9 @@ def cdd_to_schema(cdd_csv, out_json=None, out_csv=None, achiral=False):
 
     ## Load and remove any straggling compounds w/o SMILES data
     df = pandas.read_csv(cdd_csv)
-    df = df.loc[~df["suspected_SMILES"].isna(), :]
+    smiles_key = "smiles" if "smiles" in df.columns else "suspected_SMILES"
+
+    df = df.loc[~df[smiles_key].isna(), :]
```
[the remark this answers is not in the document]
kaminow: added in [f86bd72](https://github.com/choderalab/covid-moonshot-ml/pull/130/commits/f86bd72a5588fe1e3d9b0fc7127e08fdbfa5f559)

kaminow:
@JenkeScheen done!

JenkeScheen (approved):
great job!

asapdiscovery/data/utils.py
```diff
@@ -220,7 +224,11 @@ def cdd_to_schema(cdd_csv, out_json=None, out_csv=None, achiral=False):
 
     ## Get rid of the </> signs, since we really only need the values to sort
     ##  enantiomer pairs
-    pic50_key = "ProteaseAssay_Fluorescence_Dose-Response_Weizmann: Avg pIC50"
+    pic50_key = (
+        "pIC50"
+        if "pIC50" in df.columns
+        else "ProteaseAssay_Fluorescence_Dose-Response_Weizmann: Avg pIC50"
```
jchodera, line 230: Since the names of these CDD Protocols will change from assay to assay (`ProteaseAssay_Fluorescence_Dose-Response_Weizmann`), can we move these to some file-level `dict` for easier updating, or perhaps a YAML file we can modify as needed?

jchodera: Also, we don't want to use the `Avg pIC50` since this is the average of multiple measurements. We want to use the _best quality_ single measurement IC50 and lower/upper CIs.
image

asapdiscovery/data/utils.py
```diff
@@ -230,36 +238,44 @@ def cdd_to_schema(cdd_csv, out_json=None, out_csv=None, achiral=False):
...
+                ci_lower = float(ci_lower)
+                ci_upper = float(ci_upper)
+            except ValueError:
+                pic50_stderr.append(np.nan)
+                continue
+            if pandas.isna(ci_lower) or pandas.isna(ci_upper):
+                pic50_stderr.append(np.nan)
+            else:
+                ## First convert bounds from IC50 (uM) to pIC50
+                pic50_ci_upper = -np.log10(ci_upper * 10e-6)
```
jchodera, line 270: Shouldn't these be reversed, since the smallest IC50 will be the biggest pIC50?
Also, there's a conversion error here: You're using `10e-6`, which is 10 µM. You want `1e-6`.
Perhaps we can use `pint` or some other units library to handle units for us to avoid conversion errors like this?

asapdiscovery/data/utils.py
```diff
@@ -670,17 +788,61 @@ def filter_molecules_dataframe(
...
+        pIC50_upper_series.append(pIC50_upper)
 
     mol_df["pIC50"] = pIC50_series
     mol_df["pIC50_stderr"] = pIC50_stderr_series
+    mol_df["pIC50_95ci_lower"] = pIC50_lower_series
+    mol_df["pIC50_95ci_upper"] = pIC50_upper_series
+
+    ## Compute binding affinity in kcal/mol
+    # use kT = 0.593, kcal/mol for 298 K/25C
+    deltaG = lambda pIC50: -0.593 * np.log(10.0) * float(pIC50)
```
jchodera, line 809: We should do a better job of the conversion here:
* Have the `temperature` explicitly specified here so we know what assay temperature is assumed. (We could even ask Haim what temperature their facility is controlled to.)
* Use the Km for the peptide substrate to convert to a real DeltaG using [Cheng-Prusoff](https://pharmacologycanada.org/Cheng-Prusoff-equation). From [slack](https://covidmoonshot.slack.com/archives/C01J3BBP6CA/p1671451633342139?thread_ts=1671016079.148199&cid=C01J3BBP6CA), we have the Km ~ 9.5 µM for the peptide used here. From Matteo Ferla (who emailed Haim):
> The Vmax was not converted into kcat, but left in fluorescence (2,000 AU/min). It's machine dependent but given that ~10,000 AU is saturation in most machines (AFAIK) and it's 375 nM ([5-FAM]-AVLQ↓SGFR-[Lys(Dabcyl)]-K-amide substrate, it strikes me as a reassuringly healthy enzyme. (I say this because in [Okamoto et al 2010](https://pubmed.ncbi.nlm.nih.gov/21087086/) for SARS-original Mpro the numbers are baffling: a hydrolase with a kcat of 0.007 s-1 for its physiological activity is really bad)

jchodera (changes requested):
Excited to see this coming along! I've left some feedback on a few issues I spotted.

asapdiscovery/data/utils.py
ijpulidos: If we want to use some units library, may I suggest using [`openff-units`](https://github.com/openforcefield/openff-units)? That way we can convert to/from openmm units directly, which should be convenient for our purposes.

kaminow:
@jchodera I think I've addressed all of your concerns. some new updates:
* the assay name and temperature for delta G conversion are now args for the `filter_molecules_dataframe` function
* use `simtk` for constant values for improved accuracy
* add option for calculating delta G using the Cheng-Prussof equation, with the parameters for this equation also being args in the `filter_molecules_dataframe` function

asapdiscovery/data/utils.py
```diff
@@ -580,107 +651,228 @@ def filter_molecules_dataframe(
...
-            IC50 = float(row[f"{assay_name}: IC50 (µM)"]) * 1e-6  # molar
-            IC50_lower = (
-                float(row[f"{assay_name}: IC50 CI (Lower) (µM)"]) * 1e-6
-            )  # molar
-            IC50_upper = (
-                float(row[f"{assay_name}: IC50 CI (Upper) (µM)"]) * 1e-6
-            )  # molar
-
-            pIC50 = -np.log10(IC50)
+            IC50 = float(row[f"{assay_name}: IC50 (µM)"])
```
jchodera, line 724: Let's convert IC50 to M (from µM) here, so that subsequent work stores `IC50` in units of M.

jchodera (commented):
I'm not really able to evaluate this at this point since I'm not really sure what this code is supposed to do now. Could we find a time to talk through this interactively, and discuss
* What is this code supposed to do?
* What CDD Vaults / projects is it supposed to run on?
* What test we want to run to ensure it works?

kaminow:
@jchodera I believe I've addressed all of your specific comments. I sent an invitation for us to discuss this further offline, but in the meantime would it be OK to merge so we can have the new functionality in `main`?

jchodera (approved):
Looks great now! Thanks!

PR #135, opened 2023-02-04
enable downloading of MERS biological assemblies

apayne97:
## Description
Had a shower thought that it hadn't occurred to me to check whether we could download the biological assembly in CIF format and whether that would enable OpenEye to read the two "states" as two different chains. And it does!

## Todos
- the next step would be to add a protein prep script that can prepare these downloaded models! 

## Status
- [ ] Ready to go


asapdiscovery/data/pdb.py
```diff
@@ -39,30 +42,45 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
+            pdb_path = os.path.join(
+                directory, f"rcsb_{pdb_id.upper()}-assembly1.cif"
+            )
+        if not os.path.exists(pdb_path):
+            print(f"Downloading PDB entry as {pdb_type}...")
+            if download_file(url, pdb_path):
+                return pdb_path
+        else:
+            print(f"{pdb_path} already exists!...")
             return pdb_path
```
ijpulidos, line 77: (Completely optional). I had a bit of a hard time following this function, I have the feeling it could be written in a more idiomatic way. _IF_ we accept that this function should only try to download the format the user specifies then we could write it in a more readable. Perhaps something like the following (warning! untested code)
```python
url_base_str = f"https://files.rcsb.org/download/"  # base str to use for URLs
# Dictionary with allowed formats and their upstream basenames
format_to_basename = {
    "pdb": f"{pdb_id.lower()}.pdb",
    "cif": f"{pdb_id}.cif",
    "cif1": f"{pdb_id.lower()}-assembly1.cif"
}

allowed_types = format_to_basename.keys()
# Make sure pdb_type can be handled
if pdb_type not in allowed_types:
    raise NotImplementedError(
        f"pdb_type expected to be one of {allowed_types}, not '{pdb_type}'"
    )

basename = format_to_basename[pdb_type]
local_path = os.path.join(directory, f"rcsb_{basename}")
# Download only if it doesn't exist locally
if not os.path.exists(local_path):
    url = f"{url_base_str}{basename}"
    if download_file(url, local_path):
        result = local_path
    else:
        result = False  # I'm not fond on things returning bools
        
else:
    print(f"{local_path} already exists!...")
    result = local_path

return result
```

asapdiscovery/data/pdb.py
```diff
@@ -39,30 +42,45 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
-            return cif_path
-    else:
-        return cif_path
+
     print(f"Could not download PDB entry {pdb_id}.")
     return False
 
 
-def download_PDBs(pdb_list, pdb_dir):
+def download_PDBs(pdb_list, pdb_dir, pdb_type="pdb"):
```
ijpulidos, line 83: Should we put the code in this function in the entry point? That is in the main part of `scripts/download_pdbs.py`. Seems a bit unnecessary to have this here not doing much.

asapdiscovery/data/scripts/download_pdbs.py
```diff
@@ -0,0 +1,50 @@
...
+        default="pdb",
+        help="pdb_type",
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = get_args()
+    pdb_list = load_pdbs_from_yaml(args.pdb_yaml_path)
+    download_PDBs(pdb_list, args.pdb_dir_path, args.pdb_type)
```
ijpulidos, line 46: Just check the comment on this function above. We might consider doing the looping of the `pdb_list` right here. That should save having to do one extra step/hop in the middle.

ijpulidos (changes requested):
This is a good idea, being able to download the assembly cif sounds helpful. Just some comments and requested changes (some of them are not blocking). This is looking good!

ijpulidos:
I was thinking about the whole "download PDBs" functionality, and I think we might benefit from a general refactor of it (not necessarily a big effort). In that sense, maybe some of the proposed changes in my review (especially the optional ones) can be in a separate issue/PR.

asapdiscovery/data/pdb.py
```diff
@@ -17,7 +17,7 @@ def load_pdbs_from_yaml(pdb_list_yaml):
     return pdb_dict
 
 
-def download_pdb_structure(pdb_id: str, directory: str):
+def download_pdb_structure(pdb_id: str, directory: str, pdb_type: str = "pdb"):
     """
     Download a PDB structure. If the structure is not available in PDB format, it will be download
     in CIF format.
```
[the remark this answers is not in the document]
apayne97: I agree, I think. I think @kaminow wrote this originally so I'm not sure if he's using it for anything important but I would be down to change it to fail.

asapdiscovery/data/pdb.py
apayne97: Yeah the only reason why it is so convoluted was to retain the functionality it had before!

apayne97: Fine if we change it to this (could you make those commits?)

asapdiscovery/data/scripts/download_pdbs.py
apayne97: Yeah it might over-engineering to add this function. I'm fine deleting it in favor of looping in the `download_pdbs.py` script!

asapdiscovery/data/pdb.py
ijpulidos: Yes, I'll contribute those changes to this branch/PR.

ijpulidos:
I changed the behavior and signature of the `download_pdb_structure` as per my comments in the previous review. It runs fine with the code in `test-pdb-download.py` but I bet it can still create issues deeper in a pipeline. @kaminow @apayne97 can you please review this and make sure it won't create deeper issues with your workflows? Thanks!

asapdiscovery/data/pdb.py
```diff
@@ -30,39 +29,51 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
-    cif_path = os.path.join(directory, f"rcsb_{pdb_id.upper()}.cif")
-    if not os.path.exists(cif_path):
-        print("Downloading PDB entry in CIF format ...")
-        if download_file(url, cif_path):
-            return cif_path
+    url_base_str = f"https://files.rcsb.org/download/"  # base str to use for URLs
+    # Dictionary with allowed formats and their upstream basenames
+    format_to_basename = {
+        "pdb": f"{pdb_id.lower()}.pdb",
+        "cif": f"{pdb_id}.cif",
```
kaminow, line 48: does the pdb id need to be any kind of format? eg
```suggestion
        "cif": f"{pdb_id.lower()}.cif",
```

asapdiscovery/data/pdb.py
```diff
@@ -30,39 +29,51 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
+
+    basename = format_to_basename[file_format]
+    local_path = os.path.join(directory, f"rcsb_{basename}")
+    # Download only if it doesn't exist locally
+    if not os.path.exists(local_path):
+        url = f"{url_base_str}{basename}"
+        if download_file(url, local_path):
+            result = local_path
+        else:
+            result = False  # I'm not fond on things returning bools
```
kaminow, line 67: would it be better to return `None` or an empty string? or just to raise an exception or something?

kaminow (commented):
two minor comments but otherwise looks good! I don't think the changes to `download_pdb_structure` should affect anything for me

asapdiscovery/data/pdb.py
```diff
@@ -30,39 +29,51 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
-    if not os.path.exists(cif_path):
-        print("Downloading PDB entry in CIF format ...")
-        if download_file(url, cif_path):
-            return cif_path
+    url_base_str = f"https://files.rcsb.org/download/"  # base str to use for URLs
+    # Dictionary with allowed formats and their upstream basenames
+    format_to_basename = {
+        "pdb": f"{pdb_id.lower()}.pdb",
+        "cif": f"{pdb_id}.cif",
+        "cif1": f"{pdb_id.lower()}-assembly1.cif"
```
apayne97, line 49: Actually it doesn't matter
```suggestion
        "pdb": f"{pdb_id}.pdb",
        "cif": f"{pdb_id}.cif",
        "cif1": f"{pdb_id}-assembly1.cif"
```

apayne97: @kaminow

asapdiscovery/data/pdb.py
ijpulidos: It would be better to raise an exception whenever the file isn't found, but then it seems like a "major" change of behavior, that's why I didn't propose it in these changes. Since I don't know how much of the rest of the code that uses this function depends on it returning `False`

asapdiscovery/data/pdb.py
ijpulidos: I'm actually not sure about this, I just maintained the same structure as before, but do notice that the file names are case-sensitive. Are we sure the case doesn't matter here and this works as expected?

ijpulidos: They end up being downloaded as the same thing, but the pdb id case in the filename in RCSB might not be the same for every format.

apayne97: You can use any case for downloading any of the formats!

apayne97: even camel case =P

kaminow: I think the main thing is to make sure that the downloaded files have the same name regardless of what case the query was

asapdiscovery/data/pdb.py
apayne97: This is kinda nice because if you wanted to download 100 pdb files, and it only downloads 80 for some reason, I think you'd prefer it to go ahead and download 80 and just return a list of the ones it didn't work for rather than dying as soon as one isn't downloaded.

asapdiscovery/data/pdb.py
apayne97: Ohhhhhhh sorry I got it. I missed that the local path now uses the `basename`. So maybe changing to upper case here is good.

asapdiscovery/data/pdb.py
ijpulidos: @apayne97 when you want that what you do is catching the exception and handle it however you want, that's why it's advisable since you can actually track if a specific exception was raised and do something with it.

kaminow: I think I missed the context of this function, for some reason I thought it was handling a list of URLs, in which case I think throwing an exception in the middle of the list might interrupt pipelines. but since this is only called for a single file at a time, I think it makes sense to raise an exception

ijpulidos: Yes, I prefer having "atomic" building blocks that do less things than having functions that do a lot, that way you can mix and match in better ways, that also helps testing the API more easily. As far as I can see, this function should be handling downloading just a single file in the format specified. We can make other helper functions that use this to download many files, if needed.

kaminow: yeah I fully agree. is this something that we want to implement in this PR? I'd be happy to take care of it if so

ijpulidos: Yes, I think we should be able to work this one out in this same PR, it shouldn't take much effort. Please consider then trying to catch the exception in the `download_PDBs` function. Thanks!

asapdiscovery/data/scripts/download_pdbs.py
ijpulidos: I think this one now actually has a use-case, which is catching the exception if something cannot be downloaded and do something with it.

asapdiscovery/data/pdb.py
kaminow: @ijpulidos I think I've implemented what you wanted but please take a look when you get the chance

asapdiscovery/data/pdb.py
```diff
@@ -30,60 +31,92 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
+            result = local_path
+        elif response.ok:
+            raise requests.HTTPError(
+                (
+                    f"Received status code {response.status_code}, "
+                    "file not downloaded."
+                )
+            )
+        else:
+            response.raise_for_status()
```
ijpulidos, line 79: @kaminow Thanks for this improvement. Just one question, why is it that we have to check for `200` explicitly? I guess it can return `response.ok==True` even without downloading it, what's exactly happening in that case then? Can you clarify that for me? Thanks.

asapdiscovery/data/pdb.py
```diff
@@ -30,60 +31,92 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
     """
     Downloads pdbs from pdb_list_yaml using Kinoml.
 
     Parameters
     ----------
     pdb_list
     pdb_dir
+    ignore_errors : bool, default=True
+        If a PDB file failed to download, either catch the error and ignore, or
+        raise the error
```
ijpulidos, line 97: We are missing the descriptions for the other parameters and the whole docstring for the `file_format` parameter.

asapdiscovery/data/scripts/download_pdbs.py
ijpulidos: I'm marking this one as resolved as per latest changes.

ijpulidos (changes requested):
This is looking much better, in my opinion. Just a few comments and requested changes, thanks!

ijpulidos:
@kaminow  @apayne97 By the way, I'm more than happy to make the changes if you don't want to spend time on these, since they are basically minor changes.

asapdiscovery/data/pdb.py
kaminow: `response.ok` will return `True` for any code < 400, which I think includes codes that mean the file wasn't downloaded. I'm certainly no networking expert though so I could be wrong about this

ijpulidos: I see, thanks for clarifying!

apayne97:
> @kaminow @apayne97 By the way, I'm more than happy to make the changes if you don't want to spend time on these, since they are basically minor changes.

That would be great!

ijpulidos (approved):
Great! LGTM! :rocket:

ijpulidos:
@kaminow Can you take a look into this one and review after the namespace and file layout changes from merging main? Just to have a second pair of eyes to double check things look as expected. Thanks!

apayne97:
@ijpulidos not a result of your changes but this test now fails due to module not found, what are we supposed to use instead? #143 
`test-pdb-download.py`:

`ModuleNotFoundError: No module named 'asapdiscovery.data.pdb'`

kaminow:
it might be an issue with your path? with the new refactor I think you need to make sure each of the individual directories `asapdiscovery-*` are in your path

asapdiscovery-data/asapdiscovery/data/pdb.py
```diff
@@ -30,60 +31,96 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
-        if download_file(url, cif_path):
-            return cif_path
+    url_base_str = (
+        f"https://files.rcsb.org/download/"  # base str to use for URLs
+    )
+    # Dictionary with allowed formats and their upstream basenames
+    format_to_basename = {
+        "pdb": f"{pdb_id}.pdb",
+        "cif": f"{pdb_id}.cif",
+        "cif1": f"{pdb_id}-assembly1.cif",
```
kaminow, line 55: do we want to standardize the `pdb_id` here? eg:
```suggestion
        "pdb": f"{pdb_id.lower()}.pdb",
        "cif": f"{pdb_id.lower()}.cif",
        "cif1": f"{pdb_id.lower()}-assembly1.cif",
```

asapdiscovery-data/asapdiscovery/data/scripts/download_pdbs.py
```diff
@@ -0,0 +1,50 @@
...
+    )
+    parser.add_argument(
+        "-r",
+        "--ref_path",
+        default=None,
+        help="Path to pdb reference file to align to",
+    )
+    parser.add_argument(
+        "-n", "--ref_name", default=None, help="Name of reference"
+    )
```
kaminow, line 33:
```suggestion
```
can get rid of these since we're no longer aligning in this script

asapdiscovery-data/asapdiscovery/data/pdb.py
```diff
@@ -30,60 +31,96 @@ def download_pdb_structure(pdb_id: str, directory: str):
...
+    }
+
+    allowed_types = format_to_basename.keys()
+    # Make sure pdb_type can be handled
+    if file_format not in allowed_types:
```
kaminow, line 60: should probably also standardize the `file_format` variable to avoid throwing errors from someone passing eg `"PDB"`:
```suggestion
    file_format = file_format.lower()
    if file_format not in allowed_types:
```

kaminow (changes requested):
couple more small things, sorry for not catching these earlier. otherwise looks good!

apayne97:
Hmm so if I leave the `data/__init__.py` uncommented, I get this error:
```
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/tests/test-pdb-download.py", line 6, in <module>
    from asapdiscovery.data.pdb import download_pdb_structure, load_pdbs_from_yaml
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/__init__.py", line 5, in <module>
    __version__ = version("asapdiscovery-data")
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/importlib/metadata.py", line 530, in version
    return distribution(distribution_name).version
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/importlib/metadata.py", line 503, in distribution
    return Distribution.from_name(distribution_name)
  File "/Users/alexpayne/opt/anaconda3/envs/mers-docking2/lib/python3.8/importlib/metadata.py", line 177, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: asapdiscovery-data
```

If I comment out this file as in https://github.com/choderalab/asapdiscovery/pull/135/commits/dad9bca5ed0fd39b2f4b13f97b9dcd9b8c725c42, then it works just fine. This feels like a question for @dotsdl

ijpulidos:
> @ijpulidos not a result of your changes but this test now fails due to module not found, what are we supposed to use instead? #143 `test-pdb-download.py`:
> 
> `ModuleNotFoundError: No module named 'asapdiscovery.data.pdb'`

I think the approach here is that we should do `pip install -e .` from the `asapdiscovery-data` subdir, that should work.

asapdiscovery-data/asapdiscovery/data/tests/test-pdb-download.py
```diff
@@ -1,10 +1,14 @@
+import sys
+
+sys.path.append(
+    "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data"
+)
```
ijpulidos: Adding the path to the sys shouldn't be needed. Doing `pip install -e .` on `asapdiscover-data` should be able to handle things. We also don't want hardcoded paths from specific users.

asapdiscovery-data/asapdiscovery/data/__init__.py
```diff
@@ -1,5 +1,5 @@
 """Computational chemistry packages developed by the ASAP Discovery Consortium."""
 
-from importlib.metadata import version
-
-__version__ = version("asapdiscovery-data")
+# from importlib.metadata import version
+#
+# __version__ = version("asapdiscovery-data")
```
ijpulidos: I believe these are important lines for keeping up with versions for the subpackages, any special reason why we are commenting these?

asapdiscovery-data/asapdiscovery/data/tests/test-pdb-download.py
apayne97: deleted in current version!

asapdiscovery-data/asapdiscovery/data/__init__.py
apayne97: reverted in current version!

apayne97:
> I think the approach here is that we should do pip install -e . from the asapdiscovery-data subdir, that should work.

This worked!

asapdiscovery-data/asapdiscovery/data/tests/test-pdb-download.py
apayne97: Deleted in the current version!

ijpulidos:
@kaminow Can you please give this one more review iteration and check if it seems good to be merged? Thanks!

kaminow (approved):
looks great!

PR #146, opened 2023-02-16
Add moonshot dg analysis notebook

JenkeScheen:
## Description
Adds a notebook that I used to generate figure 2C of the moonshot manuscript. This notebook pulls all available moonshot sprint data and compares predicted free energies to experimental ones, for instance:
image

Note that I used `Cinnabar` for some of these analysis - if you would want to reproduce this notebook you'd have to insert a local clone of that repo.

I'm pushing the notebook here for the sake of posterity and to serve as an example on how to analyse this kind of data (plotting, but also Cheng-Prusoff conversion to pIC50, shifting the additive mean,  etc)

## Status
- [x] Ready to go


kaminow (approved):
looks good! thanks for sharing, this was a good check for me to make sure I'm doing this conversion math right in my code

PR #151, opened 2023-02-19
Feat/add pre commit ci

mikemhenry:
This is a massive reformatting, I committed the file that should make git blame still work as well: 
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


mikemhenry:
We will need to merge this in and not squash it so I can still be blamed for the non-auto formatting things

mikemhenry:
Also before we merge it I'll have to update it to the main branch manually, not with the button since that will cause the new stuff to get a new commit

mikemhenry:
But we can talk about how things have changed @dotsdl @ijpulidos @apayne97

mikemhenry:
@kaminow @hmacdope okay I think that is everyone

mikemhenry:
@JenkeScheen

mikemhenry:
Thanks everyone! I will get this merged in today! I just need to get it up to date with main first  manually so I can keep the automated commits in the same commit

ijpulidos:
@mikemhenry Thanks! Please tag me when you need this to be reviewed, just to be safe.

hmacdope (approved):
LGTM @mikemhenry, I am happy with the settings,  thanks for adding the hook. 

I will let others give the final say but all good from my side.

mikemhenry:
Note: in https://github.com/choderalab/asapdiscovery/pull/151/commits/6e6869fb9276df526e7708ec93d41d45f7e63758 I had to remove a type hint. I have no idea why it wasn't working

ijpulidos (approved):
This looks good to me, thanks!

I could reproduce the problem with the type hint. Strange that happened but it shouldn't be a problem.

asapdiscovery-data/asapdiscovery/data/schema.py
```diff
@@ -1,7 +1,7 @@
-from typing import Dict, List
```
apayne97, line 1: @mikemhenry is there a reason why this was changed? Now I get this error on import:
```
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-modeling/asapdiscovery/modeling/tests/prep_mers_rcsb/test-cif1-to-prepped-du.py", line 30, in <module>
    from asapdiscovery.docking.modeling import align_receptor
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-docking/asapdiscovery/docking/__init__.py", line 10, in <module>
    from .modeling import *  # noqa: F403,F401
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-docking/asapdiscovery/docking/modeling.py", line 15, in <module>
    from asapdiscovery.data.schema import CrystalCompoundData
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/schema.py", line 11, in <module>
    class ExperimentalCompoundData(Model):
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/schema.py", line 43, in ExperimentalCompoundData
    experimental_data: dict[str, float] = Field(
TypeError: 'type' object is not subscriptable
```

I think this is because it it using the default `dict` type instead of the typing.Dict object.

ijpulidos: This was addressed in https://github.com/choderalab/asapdiscovery/pull/157 we were masking the `list` built-in keyword

ijpulidos: Oh, this probably means that there's something masking the `dict` built-in keyword. Or you are using a python version that's not supported.

ijpulidos: I think you need python 3.9+ for this not to be a problem. I guess we decided not to support older python versions.

apayne97: if I revert this change as in this commit:
https://github.com/choderalab/asapdiscovery/pull/104/commits/bd351de95d696e9f70083682b50fc45ac805fc4d

The error goes away. The schema being used by @kaminow here uses special types (?). But we aren't overwriting the defaults, we're just using a different object (Dict vs dict)

apayne97: OH i see

apayne97: I'm using python 3.8, I can try to upgrade to 3.9 and see if anything breaks

PR #154, opened 2023-02-22
Make lack of OE license fail nicely 

### LINKED ISSUE #149: Lack of OE license should fail cleanly
Some of the scripts have OpenEye dependencies, meaning they will hang if the `OE` license is not found or valid. 

We should have a clean fail on import to make sure this doesn't confuse peeople. 

The following snippet from @apayne97 should do the trick.

```python
oechem.OEChemIsLicensed("python")
```

could be turned to 

```python
if not oechem.OEChemIsLicensed("python"):
    raise RunTimeError("OpenEye license required")
```

We could have this run at module import to avoid having to put it in every script, but this would preclude non-openeye dependent scripts from running cleanly. 

Feedback welcome.

hmacdope:
Fixes #149

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/openeye.py
+++ b/asapdiscovery-data/asapdiscovery/data/openeye.py
@@ -2,7 +2,7 @@ from openeye import oechem, oedocking, oegrid
 
 # exec on module import 
 if not oechem.OEChemIsLicensed("python"):
-    raise RuntimeError("OpenEye license required to use openeye module")
+    raise RuntimeError("OpenEye license required to use asapdiscovery openeye module")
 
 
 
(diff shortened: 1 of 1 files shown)


JenkeScheen (commented):
is this a catch-all or should this be implemented at other points throughout the codebase where OE is imported?

hmacdope:
> is this a catch-all or should this be implemented at other points throughout the codebase where OE is imported?

This should trigger when you import the module so is a catch all. To avoid needing open eye for all scripts that use asap-discovery this module should probably be excluded from __init__.py as is the case now. 😀

JenkeScheen (approved):

PR #158, opened 2023-02-25
Feat/have flake8 ignore line lenght

mikemhenry:
Since black will auto format due to line length, we don't need flake8 to check that


ijpulidos (approved):
This looks good.

Just makes me wonder if we would have to manually add/remove these `noqa` entries every time we want to change something like this. I know we have just started with this and that we probably won't be doing that too often, but just wondering what's the common approach here (not a blocker).

Issue #132, opened 2023-02-03
Standardize script / repo organization


apayne97:
# Description
So we have these potential elements
1. python api in modules
2. jupyter notebooks for testing and exploring various things
3. python scripts that import from the api
4. cluster-submission scripts that run the python scripts with specific inputs, potentially useful as examples
5. tests
6. repo / packaging elements (i.e. environment yamls)

I'm not sure exactly where everything should be put but I think we should stick to a specific format once we decide on it. I had been thinking it would be best to split everything up among the modules but it might be better to have an examples script and put the cluster scripts / explanatory jupyter notebooks and README's there.

## Suggested organization
- asapdiscovery
     - docking
          - __init__.py
          - docking.py
          - analysis.py
          ...
          - scripts (maybe rename to cli)
          - tests
     - simulation
     ...
- examples
     - fauxalysis-generation-from-pdbs
          - README.md
          - cluster_script.sh
          - jupyter_notebook.ipynb [either in place of the cluster script or in addition to in order to demonstrate how to use the api]
          - log_files
               - cluster_script.out
               - cluster_script.stderr
- devtools
     - cond-envs
     - troubleshooting-notebooks (jupyter notebooks)

## Potentially related to these issues / pull requests
- [x] #98
- [x] #53 
- [x] #51 
- [x] #49 
- [x] #48 
- [ ] #20 
- [x] #4

dotsdl:
For our approach to CLIs, I'm working to establish a general pattern in `asapdiscovery-data`: https://github.com/choderalab/asapdiscovery/blob/aws-s3-upload/asapdiscovery-data/asapdiscovery/data/cli.py

Basically, each namespace package gets its own CLI, such as `asapdiscovery-data`. This CLI can then have its own tree of subcommands. In the case of `asapdiscovery-data`, we have:
- `asapdiscovery-data postera` : interaction points for PostEra
- `asapdiscovery-data aws s3` : interaction points for AWS S3
- and so on

There may still be a place for standalone scripts within a given namespace package, but I think we should try to push toward CLI points where possible instead.

apayne97:
I think we've made great progress on this!

PR #159, opened 2023-02-25
Added AWS S3 push interface

dotsdl:
## Description
Added S3 object and `aws.s3` submodule.

Also added CLI command tree, and reorganized other command trees (e.g.
`postera`) into own submodule.

## To Do
- [x] CLI tests

## Status
- [x] Ready to go


dotsdl:
Working to fix broken test suite; think it may be due to `moto` causing hangs in server mode.

hmacdope:
> Working to fix broken test suite; think it may be due to `moto` causing hangs in server mode.

@dotsdl I'll let you know if I come up with any info in parallel while trying to get moto working for `alchemiscale` on my mac.

dotsdl:
@mikemhenry or @ijpulidos, are you able to reproduce hangs of the test suite here on your Linux hosts? We're seeing hanging on both the MacOS and Ubuntu CI runners, but I just can't reproduce the hangs at all locally (Arch Linux), so really hard to assess what's causing it.

dotsdl:
Think we'll just drop use of `moto` server mode here, and do our best to test components using its decorator approach only. Will re-organize things around this principle next.

mikemhenry:
I'm going to see if I can reproduce the CI issues

mikemhenry:
@hmacdope if you run these tests locally, do they pass?

dotsdl:
@hmacdope this is ready for review.

mikemhenry:
@dotsdl do you want me to fix the flake8 errors?

dotsdl:
@mikemhenry go for it!

asapdiscovery-data/asapdiscovery/data/cli.py
```diff
@@ -2,181 +2,10 @@
...
-            f"PostEra molecule id field '{id_field}' not found in input file"
-        )
-
-    update_molecule_list = MoleculeUpdateList.from_pandas_df(df, id_field=id_field)
-
-    molecules_updated = msa.update_molecules(molecule_set_id, update_molecule_list)
-    click.echo(f"Updated molecules {molecules_updated} in set {molecule_set_id}")
+# import subcommands to register them with the cli
+from .aws.cli import aws  # noqa: F401, E402
+from .postera.cli import postera  # noqa: F401, E402
```
mikemhenry, line 11: If these can be moved to the top, we can drop the E402 noqa skip

dotsdl: They cannot; that would yield a circular import. 

If we want to make flake8 happy here, we could create a `cli` subdirectory with both an `__init__.py` and a single submodule for the the `def cli` group. The imports of CLI subcommand groups along with the top-level could then live in the `__init__.py`. That seems a bit obtuse to me with little added value, but fine with doing it.

hmacdope:
> @hmacdope if you run these tests locally, do they pass?

@mikemhenry yes they do :)

@dotsdl reviewing now.

asapdiscovery-data/asapdiscovery/data/aws/cli.py
```diff
@@ -0,0 +1,108 @@
+from pathlib import Path
+
+import click
```
hmacdope, line 3: I know some of the other scripts use `argparse`, I think click will work best for a large scale cli (and we discussed as much) but we should try and enforce a consistent approach and a bit of a backport. Lets raise an issue to centralise?

asapdiscovery-data/asapdiscovery/data/aws/cli.py
```diff
@@ -0,0 +1,108 @@
...
+
+from ..cli import cli
+
+# use these extra kwargs with any option from a settings parameter.
+SETTINGS_OPTION_KWARGS = {
+    "show_envvar": True,
+}
+
+
+def aws_params(func):
```
hmacdope, line 13: Are any of these strictly required? I assume some are, but you are probably best place to know which. If so could you mark them as `required=True` to stop downstream failures.

asapdiscovery-data/asapdiscovery/data/postera/cli.py
```diff
@@ -0,0 +1,179 @@
...
+    id_field = click.option(
+        "--id-field",
+        help="Input csv field that stores the PostEra Molecule IDs",
+        default="id",
+    )
+
+    return id_field(func)
+
+
+@cli.group(help="Commands for interacting with the PostEra API")
```
hmacdope, line 41: Again would be good to mark any required args as such (if there are any. )

asapdiscovery-data/asapdiscovery/data/aws/s3.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        bucket = self.resource.Bucket(self.bucket)
+
+        # delete all objects, then the bucket
+        bucket.objects.delete()
+        bucket.delete()
+        bucket.wait_until_not_exists()
+
+    def push_file(self, path, location=None, content_type=None):
+
```
hmacdope, line 54: Docstring

asapdiscovery-data/asapdiscovery/data/tests/aws/test_aws_cli.py
```diff
@@ -0,0 +1,64 @@
...
+    old_env = dict(os.environ)
+    try:
+        os.environ.update(env)
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(old_env)
+
+
+class TestS3:
```
hmacdope, line 30:
```suggestion
class TestS3CLI:
```
to avoid collision with the `TestS3` class in `test_aws_s3`

devtools/conda-envs/test.yaml
```diff
@@ -25,6 +25,9 @@ dependencies:
   - pytorch
   - rdkit
   - seaborn
+  - boto3
+  - moto
+  - click
```
hmacdope, line 31: Does this also need the `flask` stuff in `pyproject.toml` or is this pulled in by something else.

hmacdope (changes requested):
Few little nitpicks but looking good!

asapdiscovery-data/asapdiscovery/data/aws/cli.py
dotsdl: Yes, probably need a standalone issue laying out this vision; I sketched the vision [here](https://github.com/choderalab/asapdiscovery/issues/132#issuecomment-1446756002), but that might not be a clear enough signpost.

asapdiscovery-data/asapdiscovery/data/cli.py
mikemhenry: No we don't need to fight the tools, using `noqa` is way better than jumping though too many hoops to make flak8 happy

asapdiscovery-data/asapdiscovery/data/postera/cli.py
dotsdl: Done!

devtools/conda-envs/test.yaml
dotsdl: Added! In general we want to put as much in the `conda` from `conda-forge` as we can, and the remainder will get pulled in via `pip` on install of the subpackage.

asapdiscovery-data/asapdiscovery/data/aws/s3.py
dotsdl: Done!

hmacdope (approved):
Looks great @dotsdl :)

PR #161, opened 2023-02-27
Change OE imports to be namespaced under `asapdiscovery` for license check. 

### LINKED ISSUE #155: Check that all `openeye` imports fall under `asapdiscovery.data.openeye`
To make sure that a lack of OE license fails cleanly, openeye should only be imported via `asapdiscovery.data.openeye`
We should check that this is the case for all scripts. 

This may require some importlib or lazy-import wrangling to make sure OE functionality as available in an easy to use way.

hmacdope:
Fixes #155 

## TODO
- [x] Do some final checks to make sure everything works 

## Description

Moves all openeye imports to be namespaced under our openeye wrapper so that the license can be checked at import time. 

Tagging @apayne97, @kaminow @ijpulidos @dotsdl @mikemhenry so that y'all are aware that this is how OE stuff should be imported from now on.

```python
from asapdiscovery.data.openeye import oefoo, oebar
```

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/openeye.py
+++ b/asapdiscovery-data/asapdiscovery/data/openeye.py
@@ -1,4 +1,4 @@
-from openeye import oechem, oedepict, oedocking, oegrid, oespruce  # noqa: E401
+from openeye import oechem, oegrid, oedocking, oespruce, oedepict # noqa: F401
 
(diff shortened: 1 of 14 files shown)


mikemhenry (approved):

mikemhenry:
This looks great!

Issue #119, opened 2023-01-18
redirect the stdout and stderr for each individual prep job to their own log.txt


apayne97:
# Description
I have now implemented using logging which outputs a log to individual log.txt files in each individual prepped protein directory. this is nice but it doesn't actually accomplish the potentially most useful thing, which is re-directing the stdout and stderr for each individual prep job to their own log.txt files. Right now when this is run in parallel on the cluster, all the error messages get overlapped in the log for the cluster script.

## Conclusion
The improvements suggested by @hmacdope in #171 work well, with the main trick being these two lines:
```
errfs = oechem.oeofstream(os.path.join(out_dir, f"openeye_{logname}-log.txt"))
oechem.OEThrow.SetOutputStream(errfs)
```
https://docs.eyesopen.com/toolkits/python/oechemtk/OESystemConstants/OEThrow.html#OESystem::OEThrow

# To-Do
- [ ] #171 
- [x] implement for mers prep script in #104 
- in the future this should be done this way for all future scripts that run things in parallel using openeye

## Related Issues / PRs
- [ ] #104

apayne97:
@ijpulidos @mikemhenry @JenkeScheen do you have suggestions for the best way to do this?

apayne97:
@hmacdope would you be able to help with this? this is for the `prep_proteins.py` script in the branch #104

hmacdope:
Yep!

JenkeScheen:
I'll leave the programmatic design choices to @hmacdope but I think it could be useful to write any logfiles that contain *hard* errors (i.e. ones where jobs have failed) to a more high-level location such as `./output/error_logs/` (possibly still writing them to protein folder files). I realise that may be a bit crowded but it's nice to not have to sift through all protein folders.

apayne97:
@JenkeScheen it's pretty easy to `grep failed */*.txt` to see which ones failed! right now OpenEye warnings and errors are all getting written to the top level and its just incomprehensible. The only way in which printing to the top level would be useful would be if all error messages come with the job / complex / etc information in the line.

However, it would be pretty easy to have in the `prep_mp` part of the script just a prefix for all the output error messages that includes the complex id, and then write out to the top level. But I don't think that is any more helpful than grepping the directory?

ijpulidos:
[...]

PR #171, opened 2023-03-08
Add logging utilities and note about how to capture openeye logs

hmacdope:
## Description

Possible fix for #119. 

Openeye logging is a bit complicated, waiting to verify that this works in a `multiprocessing.Pool`
with more than one thread with some test data from @apayne97.


apayne97:
As mentioned in #119, this works for me in PR #104.

I think we should go ahead an pull this addition into main so that we can start using it in other branches!

hmacdope:
> As mentioned in #119, this works for me in PR #104.
> 
> 
> 
> I think we should go ahead an pull this addition into main so that we can start using it in other branches!

Great to hear!

apayne97 (approved):
Looks fine to me!

PR #178, opened 2023-03-14
Weights interface to pull from public S3 bucket

hmacdope:


hmacdope:
@dotsdl @kaminow would I be able to grab a review from one of you? :)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Parameters
+    ----------
+    url : str
+        Remote url to download file from.
+    path : str
+        Local path to save file to.
+    """
+    local_filename = url.split("/")[-1]
+    path = Path(path + local_filename)
```
kaminow, line 134: maybe add a check here for if `path` is a file vs directory, eg
```suggestion
    if os.path.isdir(path):
        local_filename = url.split("/")[-1]
        path = Path(path + local_filename)
    else:
        path = Path(path)
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+):
+    """Fetch weights from yaml spec file.
+
+    Parameters
+    ----------
+    yamlfile : str
+        Path to yaml spec file.
+    models : List[str]
+        Model names to fetch weights for.
+    local_path : str, optional
```
kaminow, line 24: not sure if this is the correct way to do things, but I like to add the default here as well, eg
```suggestion
    local_path : str, default="./_weights/"
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+        Local path to save weights if a remote url is provided. or to check if weights exist locally, by default "./_weights/"
+    force_fetch : bool, optional
```
kaminow, line 26: same here
```suggestion
    force_fetch : bool, default=False
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+    if isinstance(models, str):
+        models = [models]
+
+    weights_files = {}
+    filename_set = set()
+    for model in models:
+        model_spec = spec[model]
+        weights = model_spec["weights"]
+        weights_files[model] = fetch_weights(weights, local_path, force_fetch)
+        if weights in filename_set:
```
kaminow, line 53: can we move this check to before the call to `fetch_weights`? just saves a bit of time and could prevent potentially overwriting something

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+import validators
+import yaml
+
+
+def fetch_weights_from_spec(
+    yamlfile: str,
+    models: Union[list[str], str],
+    local_path: str = "./_weights/",
+    force_fetch: bool = False,
+):
```
kaminow, line 15: add return signature here?
```suggestion
) -> dict:
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+def fetch_weights(weights: str, path: str = None, force_fetch: bool = False) -> Path:
+    """Fetch weights from remote and save to local path.
+
+    Parameters
+    ----------
+    weights : str
+        Remote url or local path to weights.
+    path : str, optional
+        Local path to save weights if a remote url is provided. or to check if weights exist locally, by default None
+    force_fetch : bool, optional
```
kaminow, line 70: same thing as before
```suggestion
    force_fetch : bool, default=False
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+                # fetch from remote
+                logging.info("weights do not exist locally, fetching from remote")
+                if not Path(path).exists():
+                    logging.info(f"Local path {path} does not exist, creating")
+                    Path(path).mkdir(parents=True, exist_ok=True)
+                weights_file = download_file(weights, path)
+                return weights_file
+    else:
+        logging.info(f"Fetching weights from {path + weights}")
+        if not Path(path + weights).exists():
```
kaminow, line 110: what happens here if `path` is `None` (as default)?

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+        Force fetch weights from remote, by default False
+
+    Returns
+    -------
+    Path to weights file.
+    """
+
+    is_url = validators.url(weights)
+
+    if is_url:
```
kaminow, line 80: add a check to make sure `path` is not `None` (since it's required in this case)?

asapdiscovery-ml/pyproject.toml
```diff
@@ -19,13 +19,6 @@ classifiers = [
     "Operating System :: MacOS",
 ]
 dependencies = [
-    "torch",
-    "pandas",
-    "upsetplot",
-    "seaborn",
-    "dgllife",
-    "torch-cluster",
-    "e3nn"
```
kaminow, line 28: do we no longer need these because they're already covered by `mtenn` ?

kaminow (changes requested):
a couple requested changes, but overall looks great!

asapdiscovery-ml/pyproject.toml
hmacdope: Yep :) they are picked up in the environment file.

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def fetch_weights(weights: str, path: str = None, force_fetch: bool = False) -> Path:
+    """Fetch weights from remote and save to local path.
+
+    Parameters
+    ----------
+    weights : str
+        Remote url or local path to weights.
+    path : str, optional
```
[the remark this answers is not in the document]
hmacdope: I set this up to be a directory, do you think a full file file name is required? I can add if so just LMK. :)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: do you mean `path` or `weights`?

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: Yep :)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: then the string concatenation ends up with just `weights`

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: Good spotting but I will use `Dict[str, Path]`

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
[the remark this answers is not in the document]
kaminow: I think either way is fine honestly. in my mind, "path" could refer to a file path or directory path, so just wanted to check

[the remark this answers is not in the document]
hmacdope: Thats a good point, perhaps if i change it to `dir`

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
kaminow: I meant `path`, although maybe I put this comment in a bad place lol. if `path` is `None`, lots of stuff below will panic (casting `None` to `Path` raises a `TypeError`, as does trying to add `None + local_filename`)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
kaminow: I think this actually raises a `TypeError`, as the inner addition will be attempted first, which would evaluate to `None` + some string

hmacdope: I am silly

kaminow: haha happens to the best of us :)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: should be fixed now by requiring a path

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,151 @@
...
+
+    weights_files = {}
+    filename_set = set()
+    for model in models:
+        model_spec = spec[model]
+        weights = model_spec["weights"]
+        if weights in filename_set:
+            raise ValueError(
+                f"Duplicate file {weights} in spec file. Please specify a unique filename for each model."
+            )
```
kaminow, line 55: sorry, I should have caught this before, but isn't this checking if the URL in `weights` has already been downloaded? if so, I think we should probably change this error to just be a warning or something with `logging`

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,151 @@
...
+    weights: str, file_dir: str = "./_weights/", force_fetch: bool = False
+) -> Path:
+    """Fetch weights from remote and save to local path.
+
+    Parameters
+    ----------
+    weights : str
+        Remote url or local path to weights.
+    file_dir : str, default="./_weights/"
+        Local path to save weights if a remote url is provided. or to check if weights exist locally, by default None
```
kaminow, line 71:
```suggestion
        Local path to save weights if a remote url is provided, or to check if weights exist locally, by default "./_weights/"
```

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Returns
+    -------
+    Path to weights file.
+    """
+
+    is_url = validators.url(weights)
+
+    if not file_dir:
+        raise ValueError("file_dir must be provided and must not be falsy.")
```
kaminow, line 83:
```suggestion
        raise ValueError("file_dir must be provided and must not be False.")
```

kaminow (changes requested):
a couple more minor things that I missed before (sorry!)

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: I think falsy is correct in this case because it will tigger on `None` or any other `False-like`  value, not just false.

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -0,0 +1,151 @@
...
+
+    if is_url:
+        logging.info("Weights are a remote url")
+        logging.info(f"Fetching weights from {weights}")
+        if force_fetch:
+            logging.info("Force fetching weights from remote")
+            if not Path(file_dir).exists():
+                logging.info(f"Local path {file_dir} does not exist, creating")
+                Path(file_dir).mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
hmacdope: yes good point.

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
hmacdope: This checks that there are not two of the same key in the `yaml` file, perhaps I should change the name to something less confusing.

hmacdope: I have changed it to `weights_set` and left a comment

hmacdope:
Sorry for the spam @kaminow, hopefully all resolved now.

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
kaminow: ah ok, fair enough

kaminow (approved):
looks great, thanks for making all those changes!

PR #180, opened 2023-03-23
ML inference framework code

hmacdope:
Model inference code to be able to easily call the ML models on a set of data, just a draft at the moment

Currently needs a bit of a cleanup in some of the called functions from `util` but I think this approach will be clean and useful.

Would loves some feedback, from @kaminow and @JenkeScheen


asapdiscovery-ml/asapdiscovery/ml/models.yaml
```diff
@@ -0,0 +1,7 @@
+# This file is used to specify the path to the weights file for each model.
+
+model1:
+  # type of the model, must be one of the following: GAT, ....
+  type: GAT
+  # Path to the weights file for the model1.
+  weights: https://asap-discovery-ml-weights.s3.amazonaws.com/test/model.pth
```
JenkeScheen, line 7: would a `last_updated` tag be helpful here, or is it enough to just grab that from the S3 dashboard when we need it?

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -0,0 +1,108 @@
...
+            logging.info("using weights from specified local file")
+            self.weights = fetch_weights(model_name)
+        logging.info(f"found weights {self.weights}")
+
+        # build model, this needs a bit of cleaning up in the function itself.
+        self.model, _ = build_model(model_type, **build_model_kwargs)
+        logging.info(f"built model {self.model}")
+
+        # load weights
+        self.model = load_weights(self.model, self.weights)
```
JenkeScheen, line 76: [ probably out of scope for this PR ] there should be a test in this function to check whether the weights are compatible with the model. In some cases users will load the 'wrong' weights that are not loadable. If pytorch is anything like TF this will throw a fairly cryptic error, so we should probably catch it more elegantly.

JenkeScheen (approved):
LGTM!

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
kaminow: I have a function `utils.load_weights` that I made to handle loading weights for an mtenn model, maybe we can build that out a bit and use that?

asapdiscovery-ml/asapdiscovery/ml/models.yaml
hmacdope: it would be like a `datetime`? Happy to add that in.

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
hmacdope: we are using `load weights` on L76 happy to build some guardrails into it.

kaminow: ah whoops, completely missed that somehow, my bad!

asapdiscovery-ml/asapdiscovery/ml/tests/test_inference.py
```diff
@@ -0,0 +1,40 @@
...
+    inference_cls = asapdiscovery.ml.inference.GATInference("model1")
+    assert inference_cls is not None
+
+
+def test_gatinference_predict(weights_yaml, test_data):
+    inference_cls = asapdiscovery.ml.inference.GATInference(
+        "gatmodel_test", weights_yaml
+    )
+    assert inference_cls is not None
+    # output = inference_cls.predict(test_data)
```
hmacdope, line 39: Add minimal test on docked unit from @JenkeScheen

asapdiscovery-ml/asapdiscovery/ml/tests/test_inference.py
```diff
@@ -0,0 +1,40 @@
...
+    # assert output is not None
```
hmacdope, line 40: Test / follow up perbation of atom ordering

hmacdope: Done!

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -894,6 +939,8 @@ def load_weights(model, wts_fn):
     """
     import torch
 
+    check_model_file_compatibility(model, wts_fn, check_weights=False)
```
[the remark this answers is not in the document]
hmacdope: I'll make the check optional and disabled by default  that way when we sort out exactly what we want to do we can re-enable it. Thoughts?

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -0,0 +1,156 @@
...
+    Inference base class for PyTorch models in asapdiscovery.
+
+    Parameters
+    ----------
+    model_name : str
+        Name of model to use.
+    model_type : str
+        Type of model to use.
+    from_spec : bool, default=True
+        Whether to fetch weights from asapdiscovery.ml spec file or from local file.
```
kaminow, line 25: I think this should be updated to `model_spec` based on below

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -0,0 +1,156 @@
...
+        logging.info(f"built model {self.model}")
+
+        # load weights
+        self.model = load_weights(self.model, self.model_components.weights)
+        logging.info(f"loaded weights {self.model_components.weights}")
+
+        self.model.eval()
+        logging.info("set model to eval mode")
+
+    def build_model(self, model_type: str, **kwargs):
```
kaminow, line 105: should we also include `*args` here or is there not really any reason for it?

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -0,0 +1,156 @@
...
+        Keyword arguments to pass to build_model function.
+    device : str, default='cpu'
+        Device to use for inference.
+
+
+    Methods
+    -------
+    predict
+        Predict on a batch of data.
+
```
kaminow, line 36: add `build_model` method
```suggestion
    build_model
        Build model from arguments

```

asapdiscovery-ml/asapdiscovery/ml/utils.py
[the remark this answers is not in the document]
kaminow: I think that would work. another option is to just move this check down below where we initialize that layer, and then the check should work as expected

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
hmacdope: Other than `model_type` the rest of the args to `util.build_model` are `kwargs` so I don't see a reason. If we want we can always change it easily down the track.

asapdiscovery-ml/asapdiscovery/ml/utils.py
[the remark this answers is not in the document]
hmacdope: Done, I also added an overload so you check against a file or a `state_dict`. :)

hmacdope:
@kaminow sorry for the spam, assuming CI comes back green this should be ready for a final review.

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+
+    model_type = "GAT"
+
+    def __init__(
+        self,
+        model_name: str,
+        model_spec: Optional[Path] = None,
+        device: str = "cpu",
+        build_model_kwargs: Optional[dict] = None,
```
kaminow, line 154: doesn't really matter at all, but just to match the order of the base class
```suggestion
        build_model_kwargs: Optional[dict] = None,
        device: str = "cpu",
```

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -876,7 +878,54 @@ def load_exp_data(fn, achiral=False, return_compounds=False):
...
+    # Check the model weights
+    if check_weights:
+        for key in model_state_dict.keys():
+            if model_state_dict[key].shape != test_state_dict[key].shape:
+                raise ValueError("Model weights shape doesn't match the file.")
+
+            if not torch.allclose(
+                model_state_dict[key], test_state_dict[key], atol=1e-4
+            ):
+                raise ValueError("Model weights don't match the file.")
```
kaminow, line 922: is there a scenario where we would want to check that a model has the same weights as a file?

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -876,7 +878,54 @@ def load_exp_data(fn, achiral=False, return_compounds=False):
...
+
+        # The file is compatible with the model
+    return None
```
kaminow, line 925: this is the default behavior so we can just get rid of these lines I think
```suggestion
```

kaminow (approved):
mostly just some minor nitpicks, but otherwise looks great! one thing that I think might be missing though is just a small script (`scripts/inference.py` or something) that runs inference in a production-like manner. approving for now though, so feel free to merge if you think that's out of the scope of this PR!

asapdiscovery-ml/asapdiscovery/ml/utils.py
hmacdope: I guess possibly if comparing that two state dicts are equal, but you are right probably not needed right now.

Issue #187, opened 2023-03-25
Ship PDB/CIF input files as part of the asapdiscovery-data package


apayne97:
In order to run tests and provide examples, we need to be able to include some binary files (oedu, npy, etc) as well as large files (pdb, cif, sdf, csv) as a part of our repo. @ijpulidos has been very cautious about this as it results in GitHub keeping track of wayyyy too many changes.

We need some way to be able to ship these files as a part of the repo!

kaminow:
could we potentially have links to some AWS bucket?

dotsdl:
Let's proceed with a similar approach as we're taking for pulling model weights in `asapdiscovery.ml`; we can host the artifacts on a public AWS S3 bucket, and give the examples API in `asapdiscovery.data.examples`.

apayne97:
do you have a protocol for how to set this up? happy to do it but not sure how

hmacdope:
If we can detail what the metadata for each download is I can whip up a setup similar to the one for the ML weights. 

Here is an example of the YAML for the weights. 
```yaml
gat_test_v0:
  type: GAT
  base_url: https://asap-discovery-ml-weights.s3.amazonaws.com/test/
  weights:
    resource: gat_example.pth
    sha256hash: 38e5476ede0ed67ad0951d5d1c7c913ecf0aeaa6a1913d16ee9dca28b5d9795b
  config:
    resource: gat_example_config.json
    sha256hash: d0b4d6c1f74b3e9caacea50af6ff6927803143842a72e74ebc7dbefdb2bcf444
  last_updated: 2023-01-01
```
If we need them to be some kind of coherent set or unit we can have a base level identifier (`model`) in this case. Otherwise we can just have a list of files and `sha256hash`.

```yaml
file1: blah
  resource: blah
  sha256hash: blahblahblahblah
file2: 
  resource: blah
  sha256hash: blahblahblahblah
```

You will also need to point me to the exact files we want and I (or someone else with S3 keys) can upload them.

dotsdl:
Working on this now.

hmacdope:
@dotsdl yeah no worries, its fresh in my mind so happy to. 

My main question is just how structured the data is, do we just want to be able to grab any old file or are they bundled in some kind of consistent logical way. TBH counter-intuitively I think a more unstructured approach may be easier as you can just call the file by name. 

tagging @apayne97 @ijpulidos and @kaminow for some thoughts.

apayne97:
[...]

PR #192, opened 2023-03-28
Add docking tests and logger implementation

apayne97:
## Description
The current docking protocol is prone to difficult debugging. This is to implement some options to help with debugging.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
@@ -286,6 +286,12 @@ def get_args():
         default=1,
         help="Number of poses to return from docking.",
     )
+    parser.add_argument(
+        "--debug_num",
+        type=int,
+        default=-1,
+        help="Number of docking runs to run. Useful for debugging and testing.",
+    )
 
     return parser.parse_args()
 
@@ -415,9 +421,10 @@ def main():
     nprocs = min(mp.cpu_count(), len(mp_args), args.num_cores)
     print(
         f"CPUs: {mp.cpu_count()}\n"
-        f"N Processes: {mp_args}\n"
+        f"N Processes: {len(mp_args)}\n"
         f"N Cores: {args.num_cores}"
     )
+    mp_args = mp_args[: args.debug_num]
     print(f"Running {len(mp_args)} docking runs over {nprocs} cores.")
     with mp.Pool(processes=nprocs) as pool:
         results_df = pool.starmap(mp_func, mp_args)
--- a/asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
+++ b/asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
@@ -1,0 +1,9 @@
+# I know we'd rather use pytest, but this was much faster to write
+run-docking-oe \
+-l './inputs/2022_12_02_fragalysis_correct_bond_orders_220_P_structures.sdf' \
+-r './inputs/Mpro-P0008_0A_ERI-UCB-ce40166b-17/*.oedu' \
+-o . \
+-n 2 \
+--omega \
+--relax clash \
+--debug_num 2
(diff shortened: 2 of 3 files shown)


asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -101,30 +105,30 @@ def load_dus(file_base, by_compound=False):
...
-            print(f"No {search_type} found for {fn}", flush=True)
+            logger.warning(f"No {search_type} found for {fn}")
             continue
 
         dataset = m.groups()[0]
         full_name = m.group()
         du = oechem.OEDesignUnit()
         if not oechem.OEReadDesignUnit(fn, du):
-            print(f"Failed to read DesignUnit {fn}", flush=True)
+            logger.warning(f"Failed to read DesignUnit {fn}")
```
hmacdope, line 120: I would say this is an error

```suggestion
            logger.error(f"Failed to read DesignUnit {fn}")
```

asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
```diff
@@ -0,0 +1,10 @@
+# I know we'd rather use pytest, but this was much faster to write
```
hmacdope, line 1: I agree that pytest will be more intricate,  but this doesn't actually programmatic test anything so we can't really easily see if things have broken.

 I'll have a look to see how hard some pytest tests would be.

hmacdope (changes requested):
I have pushed some minor changes and have two small comments. 
Other than that looks good. I'll have a quick look to see how hard a pytest docking run would be.

asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
```diff
@@ -0,0 +1,10 @@
...
+rm -rf ./outputs/retro_docking_test
```
hmacdope, line 2: Should add `mkdir -p outputs` as is required for script to run.

asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
```diff
@@ -0,0 +1,10 @@
...
+run-docking-oe \
+-l './inputs/Mpro_combined_labeled.sdf' \
```
hmacdope, line 4: This SDF doesn't seem to be vendored along with the inputs  (can it be?) In my case didn't seem to run correctly without it.

hmacdope:
@apayne97 I have added the start of a test harness in `test_docking.py` this should be able to be adapted to your needs. :)

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
apayne97: I agreed at first but I think for this script we'd rather have this as a warning because it means that the program won't quit if just 1 or 2 DUs don't load.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -105,30 +109,30 @@ def load_dus(file_base, by_compound=False):
...
-            print(f"No {search_type} found for {fn}", flush=True)
+            logger.warning(f"No {search_type} found for {fn}")
             continue
 
         dataset = m.groups()[0]
         full_name = m.group()
         du = oechem.OEDesignUnit()
         if not oechem.OEReadDesignUnit(fn, du):
-            print(f"Failed to read DesignUnit {fn}", flush=True)
+            logger.error(f"Failed to read DesignUnit {fn}")
```
apayne97, line 124:
```suggestion
            logger.warning(f"Failed to read DesignUnit {fn}")
```

asapdiscovery-docking/asapdiscovery/docking/tests/test_retro_docking.sh
apayne97: this is because we haven't resolved this issue yet: #187

apayne97:
@hmacdope I kinda think we should leave further improvements of the unit tests to a future. I don't really have the bandwidth to make new changes on that right now and would like to get these improved logging options added

hmacdope (approved):
@apayne97 looks good to me! Feel free to merge when ready.

PR #193, opened 2023-03-29
Improve dataset splitting

kaminow:
## Description
Add CLI args to the training script to allow specifying the seed for shuffling the dataset. Also fix a bug in the splitting method that made it so the dataset wasn't getting shuffled.

## Status
- [x] Ready to go


hmacdope (approved):
LGTM @kaminow

PR #196, opened 2023-03-30
Fix issue #195

### LINKED ISSUE #195: Converting CSV to schema objects fails when all pIC50 measurements are floats
The `asapdiscovery.data.utils.cdd_to_schema` function assumes that the fluorescence measurements are `str` formatted, which is generally true from CDD because there are measurements such as "> 99". This causes an issue though if there are only floats in the column, since `pandas` then casts the column as a `float`.

kaminow:
Add a line to cast fluorescence column to `str` before parsing. Closes #195 .


kaminow:
Yep. The DataFrame in question never leaves this function, which assumes that column is `str` anyway.

PR #199, opened 2023-03-30
Make combined fragalysis sdf

apayne97:
## Description
Update an old script that made a combined SDF files from SMILES strings to circumvent the problem of Fragalysis SDF files not having correct bond order.

This combined SDF file `2022_12_02_fragalysis_correct_bond_orders_220_P_structures.sdf`
Had generated this error when using OMEGA:
```
Omega failed with error Failed due to unspecified stereochemistry
```

The new script combines the SDF files made in the protein_prep pipeline along with information parsed using parse_fragalysis and the metadata.csv file.

## Todos
  - [ ] create new script
  - [ ] update tomls

## Questions
- [ ]

## Status
- [ ] Ready to go


asapdiscovery-data/asapdiscovery/data/fragalysis.py
```diff
@@ -170,10 +170,15 @@ def parse_fragalysis(
...
-        for r in df.loc[:, ["smiles", "crystal_name", "alternate_name"]].iterrows()
-    ]
+    try:
+        xtal_dicts = [
+            dict(zip(("smiles", "dataset", "compound_id"), r[1].values))
+            for r in df.loc[:, ["smiles", "crystal_name", "alternate_name"]].iterrows()
+        ]
+    except KeyError as e:
+        raise Exception(
+            "Did you use 'Mpro_compound_tracker_csv.csv'? Use 'metadata.csv' instead."
```
ijpulidos, line 180: Are these files being provided by our package? If not, then we probably need to specify what the actual format it expects (keys and values) rather than the names of the files.

apayne97: They are Fragalysis-generated files. And since this is a module intended for processing the Fragalysis database I think this make sense?

ijpulidos: I'm just saying that for me this exception message doesn't say much. Are users expected to know what the file names mean and how they are different for years to come? And if we know how they are different maybe that's what we want to print in this message (assuming it can be put in a few words).

apayne97: fair enough. added to the message!

ijpulidos (approved):
Looks good to me! LGTM.

PR #202, opened 2023-04-03
GraphDataset with no experimental data.

### LINKED ISSUE #201: Make loading experimental data optional in `GraphDataset`
Currently the GraphDataset requires that the `ExperimentalCompoundData` schema contains the keys in the following snippet:

```python
 # Build dataframe
        all_compound_ids, all_smiles, all_pic50, all_range, all_stderr = zip(
            *[
                (
                    c.compound_id,
                    c.smiles,
                    c.experimental_data["pIC50"],
                    c.experimental_data["pIC50_range"],
                    c.experimental_data["pIC50_stderr"],
                )
                for c in exp_compounds
```

For some of the inference code, the experimental data may not be known or passed around with the data-structure, so we should remove this requirement.

hmacdope:
Fixes #201 

I have made another class that subclasses `GraphDataset` that is happy without experimental data. :)


kaminow:
another option that I thought of just to avoid code duplication is just to have an `inference` flag in the `GraphDataset` class, and then only add the labels if that flag isn't set. happy to go with either option though, and this looks great if we go with this one!

hmacdope:
@kaminow I also considered that idea, but I just thought the branching might be a bit clunky. Less error prone to disrupt your training class also. :)

kaminow:
fair enough, appreciate the consideration!

PR #203, opened 2023-04-03
Minimal effort  GAT prediction in docking workflow. 

hmacdope:
Adds a very basic attempt at GAT prediction in the current docking workflow.  Thought I would do an MVP first and separate from an in-progress a rewrite of the docking script.

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/scripts/cdd_to_schema.py
+++ b/asapdiscovery-data/asapdiscovery/data/scripts/cdd_to_schema.py
@@ -1,3 +1,17 @@
+"""
+Script to convert a CSV file downloaded (and filtered) from CDD into Schema
+objects that can be used with the rest of the asapdiscovery pipeline. At a
+minimum, the CSV file must have the following columns:
+ * "smiles" or "suspected_SMILES"
+ * "Canonical PostEra ID"
+ * "pIC50" or "ProteaseAssay_Fluorescence_Dose-Response_Weizmann: Avg pIC50"
+
+Minimal example usage:
+python cdd_to_schema.py \
+-i cdd_downloaded_filtered.csv \
+-json cdd_downloaded_filtered.json
+"""
+
 import argparse
 
 from asapdiscovery.data.utils import cdd_to_schema  # noqa: E402
@@ -17,9 +31,6 @@ def get_args():
             "What type of data is " "being loaded (std: standard, ep: enantiomer pairs)"
         ),
     )
-    parser.add_argument(
-        "-achiral", action="store_true", help="Remove chiral molecules."
-    )
 
     return parser.parse_args()
 
@@ -28,7 +39,7 @@ def main():
     args = get_args()
 
     if args.type.lower() == "std":
-        _ = cdd_to_schema(args.i, args.json, args.csv, args.achiral)
+        _ = cdd_to_schema(args.i, args.json, args.csv)
     elif args.type.lower() == "ep":
         _ = cdd_to_schema_pair(args.i, args.json, args.csv)
     else:
--- a/asapdiscovery-data/asapdiscovery/data/scripts/download_moonshot_data.py
+++ b/asapdiscovery-data/asapdiscovery/data/scripts/download_moonshot_data.py
@@ -8,6 +8,7 @@ import os
 from asapdiscovery.data.moonshot import download_molecules  # noqa: E402
 
 
+################################################################################
 def get_args():
     parser = argparse.ArgumentParser(description="")
 
@@ -45,6 +46,32 @@ def get_args():
         action="store_true",
         help="Keep molecules whose IC50 values are out of range.",
     )
+    parser.add_argument(
+        "-an",
+        "--assay_name",
+        default="ProteaseAssay_Fluorescence_Dose-Response_Weizmann",
+        help="Assay name to parse as IC50.",
+    )
+    parser.add_argument(
+        "-T",
+        "--temp",
+        type=float,
+        default=298.0,
+        help="Temperature in K to use for delta G conversion.",
+    )
+    parser.add_argument(
+        "-cp",
+        "--cheng_prusoff",
+        nargs=2,
+        type=float,
+        default=[0.375, 9.5],
+        help=(
+            "[S] and Km values to use in the Cheng-Prusoff equation (assumed to be in "
+            "the same units). Default values are those used in the SARS-CoV-2 "
+            "fluorescence experiments from the COVID Moonshot project (in uM here). "
+            "Pass 0 for both values to disable and use the pIC50 approximation."
+        ),
+    )
 
     return parser.parse_args()
 
@@ -52,13 +79,17 @@ def get_args():
 def main():
     args = get_args()
 
+    # Set to None to force pIC50 usage
+    if args.cheng_prusoff == [0, 0]:
+        args.cheng_prusoff = None
+
     # Set up logging
     logging.basicConfig(level=logging.DEBUG)
 
     if "CDDTOKEN" in os.environ:
         header = {"X-CDD-token": os.environ["CDDTOKEN"]}
-    elif args.tok:
-        header = {"X-CDD-token": "".join(open(args.tok).readlines()).strip()}
+    elif args.token:
+        header = {"X-CDD-token": "".join(open(args.token).readlines()).strip()}
     else:
         raise ValueError(
             "Must pass a file for -tok if the CDDTOKEN environment "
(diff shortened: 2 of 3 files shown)


hmacdope:
This MVP of adding the GAT prediction to the docking workflow is probably ready for a review @kaminow @JenkeScheen

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -173,8 +176,39 @@ def predict(self, g: dgl.DGLGraph):
...
+        """
+        if isinstance(smiles, str):
+            smiles = [smiles]
+
+        gids = GraphInferenceDataset(
+            smiles, node_featurizer=CanonicalAtomFeaturizer(), **kwargs
+        )
+
+        data = [self.predict(g) for g in gids]
+        return np.concatenate(np.asarray(data))
```
JenkeScheen, line 214: in the case of a single SMILES, does the predicted value get returned as `value` or as `[value]`? I think returning as a float would make more sense?

asapdiscovery-ml/asapdiscovery/ml/tests/test_inference.py
```diff
@@ -107,3 +106,22 @@ def test_gatinference_predict_dataset_equal(
...
+    g1_infds, g2_infds, gids = test_inference_data
+    # same data different smiles order
+    assert inference_cls is not None
+    smiles = list(gids.smiles_dict.keys())
+    assert len(smiles) == 2
+    s1 = smiles[0]
+    s2 = smiles[1]
+    output1 = inference_cls.predict(gids[s1])
+    output2 = inference_cls.predict(gids[s2])
+    assert_allclose(output1, output2)
```
JenkeScheen, line 127: Are we testing that predicted values are different from each other? Sometimes with poorly trained models they end up learning to predict the mean label (because they're not able to learn some form of inference so predicting the mean minimizes the cost fn) of the training set and will consistently predict that value.

JenkeScheen (commented):
This is great! For project support I'll probably rename the property to "ML_SCORE" or something (just so that chemists are less confused) but that would be too granular for this code, I think. 

I'll let @kaminow have a look at the `GraphInferenceDataset(Dataset)` class.

asapdiscovery-ml/asapdiscovery/ml/tests/test_inference.py
hmacdope: I can add a test for two distinct smiles and make sure they are different. I think thats a good idea.

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
hmacdope: Yeah its being returned as [value], I can change to returning `value` for a single smiles.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -124,7 +127,7 @@ def load_dus(file_base, by_compound=False):
     return dataset_dict, du_dict
 
 
-def mp_func(out_dir, lig_name, du_name, *args, **kwargs):
+def mp_func(out_dir, lig_name, du_name, GAT_model, *args, **kwargs):
```
kaminow, line 130:
```suggestion
def mp_func(out_dir, lig_name, du_name, *args, GAT_model=None, **kwargs):
```
not sure if this is actually necessary, but just to make it clear that the `GAT_model` isn't necessary

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -137,6 +140,8 @@ def mp_func(out_dir, lig_name, du_name, *args, **kwargs):
         Ligand name
     du_name : str
         DesignUnit name
+    GAT_model : GATInference
```
kaminow, line 143:
```suggestion
    GAT_model : GATInference, optional
```
same was with above

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -333,6 +342,10 @@ def main():
         raise ValueError("Need to specify exactly one of --exp_file or --lig_file.")
     n_mols = len(mols)
 
+    # load ml models
+
+    GAT_model = GATInference("model1")
```
kaminow, line 347: I think we should probably make running the prediction optional, eg with some boolean CLI flag or something

asapdiscovery-ml/asapdiscovery/ml/dataset.py
```diff
@@ -547,22 +564,35 @@ def __init__(
...
+            return self.smiles_dict[idx]
+        elif isinstance(idx, int):
+            return self.graphs[idx]
+        elif isinstance(idx, list):
+            return [self.graphs[i] for i in idx]
+        elif isinstance(idx, slice):
+            # Use the slice indices to slice the list of graphs and create a new GraphInferenceDataset
+            start, stop, step = idx.indices(len(self.graphs))
+            indices = range(start, stop, step)
+            new_dataset = GraphInferenceDataset(
```
kaminow, line 586: I'm not sure which is better here, but `pytorch` has a [`Subset` class](https://pytorch.org/docs/stable/data.html#torch.utils.data.Subset) that I think does what we'd want in the case of getting a `slice` here, and it also prevents any data copying I believe

asapdiscovery-ml/asapdiscovery/ml/dataset.py
```diff
@@ -547,22 +564,35 @@ def __init__(
...
+                exp_compounds=[],  # empty list since we're creating a new dataset
```
kaminow, line 587: why don't we want to copy over the compounds?

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -173,8 +176,43 @@ def predict(self, g: dgl.DGLGraph):
...
             DGLGraph object.
 
+        Returns
+        -------
+        np.ndarray
+            Predictions for each graph.
         """
         with torch.no_grad():
             output_tensor = self.model(g, g.ndata["h"])
+            # unsure exactly why this reshape is done
```
kaminow, line 186: this was done in the training methods to match the shape of the labels, probably isn't needed here

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
```diff
@@ -173,8 +176,43 @@ def predict(self, g: dgl.DGLGraph):
...
+        """
+        if isinstance(smiles, str):
+            smiles = [smiles]
+
+        gids = GraphInferenceDataset(
+            smiles, node_featurizer=CanonicalAtomFeaturizer(), **kwargs
+        )
+
+        data = [self.predict(g) for g in gids]
+        data = np.concatenate(np.asarray(data))
```
kaminow, line 214: is there a reason we need the `np.concatenate` here? I don't think it matters, I'm just curious

asapdiscovery-ml/asapdiscovery/ml/tests/data/make_example_GAT_data.py
```diff
@@ -9,13 +9,15 @@
 
 # lets make a synthetic dataset using a smiles from Fragalysis
 
-# we will use a smiles from P0045_0A:TRY-UNI-2EDDB1FF-7 Fragalysis ID
+# we will use a smiles from P0045_0A:TRY-UNI-2EDDB1FF-7 Fragalysis ID for compound 1 and reordered version
 input_smiles = "Cc1ccncc1NC(=O)Cc1cc(Cl)cc(O[C@H]2CC(=O)N2)c1"
 
 mol = Chem.MolFromSmiles(input_smiles)
 # randomly shuffle the smiles string to test equivariance
 reorder_smiles = Chem.MolToSmiles(mol, doRandom=True)
```
kaminow, line 17: this is a great test!

kaminow (changes requested):
a couple of questions/comments but overall this looks great!

asapdiscovery-ml/asapdiscovery/ml/inference/inference.py
hmacdope: good spotting, the `asarray` probably does the job.

hmacdope: Actually wrong, the concatenate is needed to flatten.

asapdiscovery-ml/asapdiscovery/ml/dataset.py
hmacdope: Fixed now by using subset.

asapdiscovery-ml/asapdiscovery/ml/dataset.py
hmacdope: Using subset now, good tip seems very handy.

hmacdope:
@kaminow hopefully I have addressed your review. :)

PR #207, opened 2023-04-05
Fix issue 205

### LINKED ISSUE #205: Lack of seqres input in `prep_proteins` script should fail cleanly
If you don't add a  value for the  `--seqres_yaml` argument in `modelling/prep_proteins.py` you get a rather cryptic 

```
  File "/store/ASAP_project/asapdiscovery/asapdiscovery-docking/asapdiscovery/docking/modeling.py", line 764, in prep_mp
    seqres=" ".join(res_list),
UnboundLocalError: local variable 'res_list' referenced before assignment
```

rather than a clean failure. Its likely  I don't understand all the use cases but  If the argument is indeed required we should make it so.

kaminow:
Prevent `res_list` from being called before being defined. Resolves #205


hmacdope (approved):
Thanks @kaminow! Works perfectly on my local tests.

Issue #212, opened 2023-04-07
Is it possible that when running docking in parallel on lilac we need to be clearing the memory?


apayne97:
# Description
Not sure if this is a lilac thing, an OpenEye thing, a multiprocessing thing, or just an unrelated set of errors, but @JenkeScheen and I have been getting some core dumps and I'm wondering if in part it is due to the way in which memory is being allocated (and not overwritten by OpenEye).

# Potentially Related
- #181 

# Error Messages
```
/scratch/lsftmp/1680789260.16343999.shell: line 17:  7538 Bus error
```

# Questions
- [ ] how should I go about debugging this efficiently?
- [ ] what can we do in the meantime

JenkeScheen:
looping in @hmacdope @ijpulidos - we should be able to inspect the core dumps using `gdb` : https://stackoverflow.com/questions/5115613/core-dump-file-analysis

JenkeScheen:
I've observed core dumps on `lilac` with non-docking code as well (happened with `find_all_mcs.py`).

hmacdope:
Another thing to try is `valgrind` (I can try it at some point). It will be slow so use a small amount of data if possible.

```
export PYTHONMALLOC=malloc 
valgrind python find_all_mcs.py
```

apayne97:
Instead of actually uncovering the problem, running smaller segments of jobs in job arrays on lilac seems to be working fine for me

PR #213, opened 2023-04-07
Try to fix some docking memory issues

apayne97:
## Description
There seem to be some memory accumulation errors as mentioned in #212 during docking runs. It's not clear when / why this happens. After spending some time trying to debug the issue, instead this PR just adds some code for debugging, and for running docking as a job array to minimize the number of tasks each individual job needs to perform.

## Todos
  - [x] add function to split up multi-ligand sdf into chunks of sdf files `asapdiscovery.data.split_sdf.py`
  - [x] add cluster script to be run as a job array
  - [x] fix some minor bugs in the run_docking_oe script and function
  - [x] add datetime checks to see how long the various steps take
  - [x] add log_name param to `run_docking_oe.py` to use as basename string for high-level log files
  - [x] change import of ML code to only happen if a GAT model is passed
  - [x] add pebble to dependencies
  - [x] get tests working

## Questions
- [x] should I add back the pickle reading and dataframe generating? I'm leaning against it
- I decided to add it back! it didn't seem to require much time and it is kinda convenient to have it in this script

## Status
- [x] Ready to go


asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
```diff
@@ -0,0 +1,67 @@
...
+    n_chunks = ceil(
+        len(mols) / args.chunk_size,
+    )
+    remainder = len(mols) % args.chunk_size
+
+    print(f"Saving {n_chunks} files of {args.chunk_size} molecules each")
+    if not remainder == 0:
+        print(f"Saving {remainder} molecules in the last file")
+
+    import os
```
kaminow, line 54: is there a reason these imports aren't at the top of the script?

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
```diff
@@ -0,0 +1,67 @@
...
+
+    from asapdiscovery.data.openeye import save_openeye_sdfs
+
+    n_chunks = ceil(
+        len(mols) / args.chunk_size,
+    )
+    remainder = len(mols) % args.chunk_size
+
+    print(f"Saving {n_chunks} files of {args.chunk_size} molecules each")
+    if not remainder == 0:
```
kaminow, line 51:
```suggestion
    if remainder:
```

kaminow (approved):
couple minor suggestions, but I think this looks great! would probably be good to get a quick check from another set of eyes though

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
apayne97: LOL

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
apayne97: nope. fixed here: https://github.com/choderalab/asapdiscovery/pull/213/commits/cbe5b792b5efd5a9315ae47b8f0f01748815df31

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
```diff
@@ -0,0 +1,65 @@
...
+    mols = load_openeye_sdfs(args.sdf_fn)
+    print(f"Saving {len(mols)} SDF files to '{args.out_dir}'")
+
+    n_chunks = ceil(
+        len(mols) / args.chunk_size,
+    )
+    remainder = len(mols) % args.chunk_size
+
+    print(f"Saving {n_chunks} files of {args.chunk_size} molecules each")
+    if remainder:
```
hmacdope, line 51: I could be missing something but we arn't doing anything with `remainder`?

devtools/conda-envs/asapdiscovery.yaml
```diff
@@ -54,3 +55,6 @@ dependencies:
     - pyyaml
     # ml
     - git+https://github.com/choderalab/mtenn@0.1.0
+    # testing
```
hmacdope, line 58: These are from #223 ?

hmacdope (changes requested):

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -347,10 +355,10 @@ def get_args():
         "-m",
         "--timeout",
         type=int,
-        default=30,
+        default=-1,
```
JenkeScheen, line 358: is this intended? @hmacdope is this a good idea?

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
JenkeScheen: right, `remainder` still has to be written out to file as well. It's the tail end of mols if `len(mols)` is not divisible by `chunk_size`

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
```diff
@@ -0,0 +1,65 @@
...
+        print(f"Saving {remainder} molecules in the last file")
```
JenkeScheen, line 52: if possible might be good to state the file name of the `remainder` file, this will always be smaller than the rest of the files, so handy for debugging.

asapdiscovery-docking/asapdiscovery/docking/cluster_scripts/run_fragalysis_retrospective_array.sh
```diff
@@ -0,0 +1,23 @@
...
+
+#BSUB -J run_oe_docking
+#BSUB -oo run_fragalysis_retrospective_%I.out
+#BSUB -eo run_fragalysis_retrospective_%I.stderr
+#BSUB -n 32
+#BSUB -q cpuqueue
+#BSUB -R rusage[mem=3]
+#BSUB -W 2:00
+source ~/.bashrc
+conda activate ad-3.9
```
JenkeScheen, line 13: this will break for other users of course, perhaps best to put in a comment to point the users to their own conda envs

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
apayne97: since i'm using the ceiling function, I actually am also saving out the remainder. However I've decided to try to rewrite it in a way that makes it more obvious

apayne97: done in https://github.com/choderalab/asapdiscovery/pull/213/commits/c7ab339ca6eabf1f73a7ca3c2372680aa417f483

asapdiscovery-data/asapdiscovery/data/scripts/split_sdf.py
apayne97: done here: https://github.com/choderalab/asapdiscovery/pull/213/commits/c7ab339ca6eabf1f73a7ca3c2372680aa417f483

asapdiscovery-docking/asapdiscovery/docking/cluster_scripts/run_fragalysis_retrospective_array.sh
apayne97: this is true for all the cluster scripts...i'll make a separate issue for this, maybe there's a better way to keep these in the repo? i'd appreciate keeping them

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
apayne97: this was from #218 and #224; I don't think the default should be to have a timer, since the expected time to completion might be different for every kind of docking run? (and, i.e. will be 10x different if you use the Omega and clash relax flags vs default) what do you think?

asapdiscovery-docking/asapdiscovery/docking/cluster_scripts/run_fragalysis_retrospective_array.sh
apayne97: Added issue here https://github.com/choderalab/asapdiscovery/issues/230

devtools/conda-envs/asapdiscovery.yaml
apayne97: you added these in https://github.com/choderalab/asapdiscovery/commit/adc1150635b9740bf2a18e753c6157caacd2fe97

apayne97: i added the pebble changes because of #218

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
hmacdope: I also had some issues with it being too short on my setup and will differ wildly depending on hardware. I agree its hard to judge exactly how long it should take. TBH I think changing to a large value (say 300 sec) is probably better than disabling it altogether.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -409,18 +417,29 @@ def get_args():
...
+        help="Base name for high-level log file. Defaults to run_docking_oe, "
+        "which enables propagation of log messages to the root logger.",
+    )
 
     return parser.parse_args()
 
 
 def main():
     args = get_args()
+    global log_name
```
hmacdope, line 434: Why do we need this `global` ? I might be missing something but do we access it from global scope ever? Also seems to be defined in two places.  If possible better to define once (non-global) in `main` pass as an argument to whatever you need

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -458,6 +477,8 @@ def main():
     # Set up ML model
     gat_model_string = "asapdiscovery-GAT-2023.04.12"
     if args.gat:
+        from asapdiscovery.ml.inference import GATInference  # noqa: E402
```
hmacdope, line 480: Good idea to move, thanks. :)

devtools/conda-envs/asapdiscovery.yaml
hmacdope: Ah yep ok.

hmacdope (changes requested):
Quick queries but thanks for the improvements. :)

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
JenkeScheen: I agree that we should set an overly long timeout just to prevent jobs from hanging until eternity.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
apayne97: @hmacdope as discussed i'm enabling this now

apayne97: done here: https://github.com/choderalab/asapdiscovery/pull/213/commits/021ee5a4692764931e4ccbfe476390e98dd15508

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
apayne97: changed to 300 seconds here: https://github.com/choderalab/asapdiscovery/pull/213/commits/aab600f4bf0e756aa1865fa0e58952fb9f95efc5

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -535,9 +556,10 @@ def main():
             xtals.extend(dataset_dict[xtal_ids[xtal]])
         new_args = [
             (
-                os.path.join(args.output_dir, f"{compound_ids[i]}_{x}"),
```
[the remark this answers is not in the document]
apayne97: this is using the pathlib `/` operator! output_dir is a Path object.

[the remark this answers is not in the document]
hmacdope: Wow had no idea that was possible.  Great!

hmacdope (approved):
LGTM, merge when ready @apayne97

apayne97:
@hmacdope shocked that you responded so quickly, but I guess it's early enough for me that it isn't *too* late for you. go to bed though :P <3

PR #216, opened 2023-04-10
Add example usage in script docstring.

### LINKED ISSUE #206: Not providing backticks around regex will fail on run_docking_oe.py, should fail more cleanly.
We should check that inputs into  -r argument of `run_docking_oe` are valid glob that will actually return something , and document/remove  the need for quotation marks when running the code.

kaminow:
Address the issues brought up by @hmacdope in #206. Closes #206

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
@@ -1,5 +1,13 @@
 """
-Script to dock an SDF file of ligands to prepared structures.
+Script to dock an SDF file of ligands to prepared structures. Example usage:
+python run_docking_oe.py \
+-e /path/to/experimental_data.json \
+-r '/path/to/receptors/*.oedu' \
+-s /path/to/mcs_sort_results.pkl \
+-o /path/to/docking/output/
+Note that the single ticks around the receptors argument is necessary, as without it the
+OS will automatically try to expand the wildcard, which will results in a list of files
+being passed and will cause an error.
 """
 import argparse
 import multiprocessing as mp


asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -225,6 +262,11 @@ def get_args():
         "--sort_res",
         help="Pickle file giving compound_ids, xtal_ids, and sort_idxs.",
     )
+    parser.add_argument(
```
hmacdope, line 265: Conflicts with other argument with tag `-x` , maybe make it `-re`

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -200,6 +157,85 @@ def mp_func(out_dir, lig_name, du_name, *args, **kwargs):
...
+            os.path.join(receptors, fn)
+            for _, _, files in os.walk(receptors)
+            for fn in files
+            if fn[-4:] == "oedu"
+        ]
+    elif os.path.isfile(receptors):
+        print(f"Using {receptors} as file")
+        df = pandas.read_csv(receptors)
+        all_fns = [
+            os.path.join(os.path.dirname(fn), basefile) for fn in df["Docked_File"]
```
hmacdope, line 203: Perhaps a `try: except` to check if there is a col called `Docked_File` and then a prettier error message than a `KeyError`?

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -1,11 +1,15 @@
 """
-Script to dock an SDF file of ligands to prepared structures.
+Script to dock an SDF file of ligands to prepared structures. Example usage:
+python run_docking_oe.py \
```
hmacdope, line 3: Can you give an example of how the `regex` argument might be used?

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -225,6 +262,11 @@ def get_args():
         "--sort_res",
         help="Pickle file giving compound_ids, xtal_ids, and sort_idxs.",
     )
+    parser.add_argument(
+        "-x",
+        "--regex",
+        help="Regex for extracting DesignUnit identifiers from the filename.",
```
hmacdope, line 268: more specific 

```suggestion 
        help="Regex for extracting DesignUnit identifiers from the  OpenEye  `.oedu` filename.",
```

asapdiscovery-data/asapdiscovery/data/utils.py
```diff
@@ -23,8 +23,12 @@
 MOONSHOT_CDD_ID_REGEX = r"[A-Z]{3}-[A-Z]{3}-[0-9a-z]+-[0-9]+"
 MPRO_ID_REGEX = r"Mpro-.*?_[0-9][A-Z]"
 
+# Regex patterns that match chains as well, but only capture the main part
```
hmacdope, line 26: I would say these need more docs on the expected format and what exactly we are trying to capture. Might be good to add for the above  regex as well.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -333,8 +375,23 @@ def main():
         raise ValueError("Need to specify exactly one of --exp_file or --lig_file.")
     n_mols = len(mols)
 
+    # In case a glob string or directory was actually passed in
+    if len(args.receptor) == 1:
+        args.receptor = args.receptor[0]
+    # Handle default regex
+    if args.regex is None:
```
hmacdope, line 382: Few comments around here will help.

hmacdope (changes requested):
Looking good so far! See comments for detail.  A quick test says it gives the same outputs as current `main (198e0112)` but I think we do really need some programmatic tests to avoid regressions. @apayne97 and I are making some progress towards that in #192.

hmacdope (approved):
Looks good to me. I might ask @JenkeScheen to have a quick play around with the new script to see if it suits his needs?

JenkeScheen:
this seems to be working as intended! 

Just another reminder, when we start adding other non-Mpro targets (the first one is coming within these next few weeks!) all of this code is going to break spectacularly as the naming convention is going to change every time. When we need to overhaul this code to that purpose we'll probably want to move to YAMLs (or whatever @hmacdope thinks is best!). 

IMO the whole regex thing is not very necessary as we can expect from users to point to a directory that contains folders with the intended receptors and not much else. This could take the form of `-r /path/to/receptors`, and then just finding + loading all `.oedu` files in there (and potentially sending out some warnings if anything irregular is detected/loaded).

hmacdope:
@kaminow, would you be able to resolve conflicts here and I think then we are good to move forward. :)

kaminow:
should be good to go! @JenkeScheen I think the main benefit of the regex method is being able to flexibly parse the `.oedu` filenames, which makes it easier to organize the data structures. I suppose we could require that everything be named in a certain format going forward, eg `structure_name:structure_chain:compound_id.oedu`, and we could just split on the ":"

hmacdope:
Codecov being flaky, ive cycled it.

PR #217, opened 2023-04-11
Clean up downloads and specs and add hash checking. 

### LINKED ISSUE #215: Verify data downloads with hashes, possibly switch downloads to using `pooch`
We should verify the integrity of datafiles pulled from  URLs with a hash comparison. The use of pooch https://www.fatiando.org/pooch/latest/about.html as suggested by @mikemhenry may make this easier. 

Indeed, pooch may be able to replace my hand rolled download code altogether.

hmacdope:
Cleans up some of the download code, rename some of the specs add hash checking.  

EDIT: ~~I decided against pooch in the end as I was finding it difficult to give it the caching behaviour I wanted.~~

I also switched the downloads to use caching via pooch. rather than hand rolled code. 

Fixes #215


asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -47,114 +45,50 @@ def fetch_model_from_spec(
...
+    if not local_dir:
+        raise ValueError("local_dir must be provided and must not be falsy.")
+
+    if not Path(local_dir).exists():
+        Path(local_dir).mkdir(parents=True, exist_ok=True)
+
     specs = {}
-    # cannot specify the same weights file for multiple models
-    weights_set = set()
     for model in models:
```
kaminow, line 55: I think we should add checks to make sure that all the necessary keys are present:
 - `model` in `spec`

 - "type" in `model_spec`
 - "base_url" in `model_spec`
 - "weights" in `model_spec`

 - "resource" in `weights`
 - "sha256hash" in `weights`

 - "resource" in `config`
 - "sha256hash" in `config`

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -47,114 +45,50 @@ def fetch_model_from_spec(
...
+        if config_resource:
+            config_file = Path(registry.fetch(config_resource))
         else:
-            logging.info(
-                f"file exists locally at {local_path}, skipping fetch from remote"
-            )
-            return Path(local_path)
+            config_file = None
+        # make model spec
+        specs[model] = ModelSpec(model, model_type, weights_file, config_file)
```
kaminow, line 92: do we want to add a check here to make sure that `model` isn't already in `specs` and we're not silently overwriting stuff?

asapdiscovery-ml/asapdiscovery/ml/weights/weights.py
```diff
@@ -47,114 +45,50 @@ def fetch_model_from_spec(
...
+            path=Path(local_dir).joinpath(Path(subdir)),
+            base_url=base_url,
+            registry=registry,
+        )
+
+        # fetch weights
+        weights_file = Path(registry.fetch(weights_resource))
+        # fetch config
+        if config_resource:
+            config_file = Path(registry.fetch(config_resource))
```
kaminow, line 88: it looks `Pooch.fetch` can raise exceptions, so we should handle those in case of failure

kaminow (changes requested):
I love the change to `pooch`, things are looking much cleaner now! in addition to my suggestions, I also think we should have documentation somewhere for the different fields available/required in the weights yaml file, although I'm not sure exactly where that should go

asapdiscovery-ml/asapdiscovery/ml/models.yaml
```diff
@@ -1,12 +1,24 @@
...
+  # The base url is used to determine where to look for the weights and config files
+  base_url: https://asap-discovery-ml-weights.s3.amazonaws.com/test/
+  # The weights key has the name of the weights file and the sha256 hash of the file.
+  weights:
+    # resource is the name of the file
+    resource: gat_example.pth
+    # sha256hash is the sha256 hash of the file
+    sha256hash: 38e5476ede0ed67ad0951d5d1c7c913ecf0aeaa6a1913d16ee9dca28b5d9795b
+  # The config key has the name of the config file and the sha256 hash of the file.
+  config:
```
kaminow, line 18:
```suggestion
  # This key is optional
  config:
```

kaminow (approved):
looks great now, thanks!

PR #218, opened 2023-04-11
Add docking timeout

kaminow:
Add ability to specify a timeout for each individual docking run, in case some are failing.


asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -437,19 +455,57 @@ def main():
         "SMILES",
         "GAT_score",
     ]
+    if args.num_cores > 0:
```
JenkeScheen, line 458: should this not be >1? Would this form do mp when only 1 core is specified?

JenkeScheen: I think I know why you chose this, just curious for the rationale.

JenkeScheen (approved):
LGTM! This will come in very handy!

PR #219, opened 2023-04-11
Add multi receptor docking

apayne97:
## Description
OpenEye has the ability to be given a set of receptors for a given ligand, dock the ligand to all the receptors, and return the best scoring pose (using POSIT). I would like to add this capability to our code

## Todos
  - [x] changed `run_docking_oe` function to take a set of receptors
  - [x] waiting to hear back from OpenEye about bugs
  - [ ] test docking for a set of receptors with different reference ligands

## Questions & Issues
- ~currently it appears that the "best scoring structure" returned by these functions is always just whichever was added first:~
     - ~RankDesignUnits~
     - ~GetDesignUnit~
     - ~GetReceptorIndex~

- In fact, this is likely due to the fact that POSIT chooses which receptor to use based entirely on the similarity of a target ligand to the bound ligand.
- this should work for future docking goals though! 
> [posit’s](https://docs.eyesopen.com/applications/oedocking/posit/posit.html#chapter-posit) first pass is to identify a target receptor that has the most similar bound ligand. Multiple receptors are not required, but increase the odds of [posit](https://docs.eyesopen.com/applications/oedocking/posit/posit.html#chapter-posit) finding a suitable receptor. Given an input molecule, [posit](https://docs.eyesopen.com/applications/oedocking/posit/posit.html#chapter-posit) identifies the most similar bound ligand using a combination of graph similarity to the bound ligand (MACCS 166 set or OpenEye’s PATH fingerprints) and 3D similarity, TanimotoCombo, to the bound ligand. 
## Status
- [ ] Ready to go


apayne97:
here are the inputs and outputs for `test_multi_receptor_docking.py`
Inputs:
[multireceptor_docking_test.zip](https://github.com/choderalab/asapdiscovery/files/11203883/multireceptor_docking_test.zip)

Outputs:
[multireceptor_docking_test.zip](https://github.com/choderalab/asapdiscovery/files/11203884/multireceptor_docking_test.zip)

Although "docked" and "docked_reversed" sdf files match the individually docked scores, instead of returning the best scoring one (7t41), it will just return whichever was added to the receptor first

apayne97:
This is a simple enough thing to add that will benefit from the changes we are making to tests / docking functions. So I'll close this PR for now

PR #220, opened 2023-04-11
Add temporal split

kaminow:
## Description
Add the ability to split datasets temporally, in order to train on earlier compounds and attempt to predict on later compounds.
~EDIT: Actually, hold off on any reviews for this for the moment, I'm testing the splitting with grouped datasets.~
EDIT 2: OK it seems to work on the grouped datasets, so please review!

## Todos
  - [x] Add date created to CDD export and processing
  - [x] Include date created in dataset creation
  - [x] Method for appropriately splitting molecules based on date created
  - [x] Incorporate temporal splitting into training script

## Status
- [x] Ready to go


asapdiscovery-ml/asapdiscovery/ml/dataset.py
```diff
@@ -357,14 +357,15 @@ def __init__(
         from dgllife.utils import SMILESToBigraph
 
         # Build dataframe
-        all_compound_ids, all_smiles, all_pic50, all_range, all_stderr = zip(
+        all_compound_ids, all_smiles, all_pic50, all_range, all_stderr, all_dates = zip(
```
hmacdope, line 360: what if dates not present?

kaminow: I added `date_created` as an optional field in the `ExperimentalCompoundData` schema class, so any instance of the class will definitely have the field. it may be empty, but that shouldn't cause any issues here, and I think I've covered that case in the splitting methods

hmacdope (changes requested):
Few changes and queries. I will try and compact the `split_temporal` function myself and push my changes.

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1220,6 +1244,171 @@ def split_molecules(ds, split_fracs, generator=None):
...
+    ----------
+    ds : Union[cml.data.DockedDataset, cml.data.GraphDataset]
+        Molecular dataset to split
+    split_fracs : List[float]
+        List of fraction of compounds to put in each split
+    grouped : bool, default=False
+        Splitting a GroupedDockedDataset object
+    reverse : bool, default=False
+        Reverse sorting of data
+    end_splits : int, default=2
```
hmacdope, line 1263: I'm not exactly sure how this works and what a default of `2` corresponds to. I think a practical example will help.

hmacdope (changes requested):
Two quick extra comments, updates to splitting pending us work-shopping together.

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1220,6 +1244,171 @@ def split_molecules(ds, split_fracs, generator=None):
...
+    end_split_fracs = np.asarray(split_fracs[-end_splits:][::-1])
+
+    # Calculate how many molecules we want covered through each split
+    # By the end of each split, we should cumulatively covered the cumulative sum of all
+    #  splits up to and including that one
+    n_mols_front_split = front_split_fracs * len(ds)
+    # Same but for reverse
+    n_mols_end_split = end_split_fracs * len(ds)
+
+    # TODO: make this whole process more compact
```
hmacdope, line 1289: I think I have come up with a much more compact way of doing this, we will just need to work on what we want out of the `end_splits` arg, but this should be much more compact. 

Apologies for providing this in script format but  didn't want to push overwrite your hard work before we had a chance to discuss. 

```python

import pickle
import asapdiscovery
import numpy as np
import torch 

from asapdiscovery.ml.utils import split_temporal, split_molecules

def load_ds_from_pickle(path_to_pickle):
    with open(path_to_pickle, 'rb') as f:
        ds = pickle.load(f)
    return ds

ds = load_ds_from_pickle('blah.pkl')
key_func = lambda x: x["date_created"]
sorted_data_indices = sorted(range(len(ds)), key=lambda i: key_func(ds.structures[i]))

def split_list_into_chunks_range(data, ratios):
    chunks = []
    total_items = len(data)
    start = 0
    for ratio in ratios:
        end = int(start + ratio * total_items)
        chunks.append(data[start:end])
        ret_start = start
        start = end
    # tack any remainder onto the last chunk
    if end != total_items:
        chunks[-1] = data[ret_start: total_items]
    return chunks

chunks =  split_list_into_chunks_range(sorted_data_indices, [0.8, 0.1, 0.1])

ds_train = torch.utils.data.Subset(ds, chunks[0])
ds_test = torch.utils.data.Subset(ds, chunks[1])
ds_val = torch.utils.data.Subset(ds, chunks[2])

print("ds_train: ", ds_train, "length: ", len(ds_train))
print("ds_test: ", ds_test, "length: ", len(ds_test))
print("ds_val: ", ds_val, "length: ", len(ds_val))

# have a look at the dates
print("DS TRAIN\n")
for i in ds_train:
    print(i[1]["date_created"])

print("DS TEST\n")
# have a look at the dates
for i in ds_test:
    print(i[1]["date_created"])


print("DS VAL\n")
# have a look at the dates
for i in ds_val:
    print(i[1]["date_created"])
```

asapdiscovery-ml/asapdiscovery/ml/scripts/build_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+        achiral=args.achiral,
+        cache_fn=args.o,
+        grouped=args.grouped,
+        lig_name=args.n,
+        num_workers=args.w,
+        check_range_nan=args.check_range_nan,
+        check_stderr_nan=args.check_stderr_nan,
+    )
+
+    # GAT model creates a bin file, need to dump the pickle file
```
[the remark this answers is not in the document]
kaminow: the `cache` argument of the `build_dataset` function works differently for the `GAT` models vs the `SchNet` and `e3nn` models. for `GAT`, a `.bin` file is created in one of the `dgl-lifesci` functions, whereas for `SchNet` and `e3nn` we dump the pickle file in the `build_dataset` function directly. tl;dr the script still works for `SchNet` and `e3nn`, we just don't need to do anything extra external to `build_dataset`

asapdiscovery-ml/asapdiscovery/ml/scripts/build_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Parse compound filenames
+        xtal_regex = args.xtal_regex if args.xtal_regex else MPRO_ID_REGEX
+        compound_regex = args.cpd_regex if args.cpd_regex else MOONSHOT_CDD_ID_REGEX
+        compounds = extract_compounds_from_filenames(
+            all_fns, xtal_pat=xtal_regex, compound_pat=compound_regex, fail_val="NA"
+        )
+
+        # Trim compounds and all_fns to ones that were successfully parse
+        idx = [(c[0] != "NA") and (c[1] != "NA") for c in compounds]
```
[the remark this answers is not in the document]
kaminow: ha thanks :)

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1113,12 +1124,20 @@ def split_dataset(
...
+                ds, [train_frac, val_frac, test_frac], grouped=True
+            )
+            n_train = len(ds_train)
+            n_val = len(ds_val)
+            n_test = len(ds_test)
+        else:
+            ds_train, ds_val, ds_test = torch.utils.data.random_split(
+                ds, [n_train, n_val, n_test], g
+            )
+            n_train = int(len(ds) * train_frac)
```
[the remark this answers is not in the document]
kaminow: yeah good spot, idk what I was thinking here lol. should be much cleaner as of [83d65a7](https://github.com/choderalab/asapdiscovery/pull/220/commits/83d65a7252c667bb63e7e571f90645a159fa8a73)

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1113,12 +1124,20 @@ def split_dataset(
...
+        if temporal:
+            ds_train, ds_val, ds_test = split_temporal(
+                ds, [train_frac, val_frac, test_frac], grouped=True
+            )
+            n_train = len(ds_train)
+            n_val = len(ds_val)
+            n_test = len(ds_test)
+        else:
+            ds_train, ds_val, ds_test = torch.utils.data.random_split(
+                ds, [n_train, n_val, n_test], g
```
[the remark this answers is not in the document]
kaminow: should also be addressed in [83d65a7](https://github.com/choderalab/asapdiscovery/pull/220/commits/83d65a7252c667bb63e7e571f90645a159fa8a73)

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1220,6 +1244,171 @@ def split_molecules(ds, split_fracs, generator=None):
...
+    # For each Subset, grab all molecules with the included compound_ids
+    all_subsets = []
+    # Keep track of which structure indices we've seen so we don't double count in the
+    #  end splits
+    seen_idx = set()
+    prev_idx = 0
+    # If there are no end splits, go up to the last split so we can add anything that
+    #  got left out from float rounding
+    if end_splits == 0:
+        final_split = n_mols_front_split.pop()
```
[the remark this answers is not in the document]
kaminow: oh good catch, this used to be a `list` object which does have a `pop` method. fixed in [7d7706f](https://github.com/choderalab/asapdiscovery/pull/220/commits/7d7706f25a259db00b9e5b5846ccaf7b275e8f09)

asapdiscovery-ml/asapdiscovery/ml/utils.py
kaminow: this is definitely much more compact, but I think the main thing this is missing is the ability to have non-contiguous subsets, eg if we want to train on the earliest 30% of the molecules, but use the most recent 20% for the val/test splits. in this case, we would set `split_fracs=[0.3, 0.1, 0.1]` and `end_splits=2`, which means the last 2 entries in the `split_fracs` list should be taken from the end of the list of molecules sorted by date.

as an example, if we have our sorted list of molecules as `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`, and we want the above split of training on the first 0.3 of molecules but test/val of the last 0.1/0.1, your code would give us splits of `[[1, 2, 3], [4], [5, 6, 7, 8, 9, 10]]` (the last split gets all leftover data). using my code, we'd get splits of `[[1, 2, 3], [9], [10]]`
(hopefully this makes things more clear and not more confusing lol)

kaminow: I think there's room to combine both of our implementations though, so let's definitely find a time to discuss!

asapdiscovery-ml/asapdiscovery/ml/utils.py
kaminow: the idea for a default of 2 is that we'd want to take the val and test splits from the end, but I've changed the default to 0 and passed the responsibility of setting this arg into the `split_dataset` dispatch function

hmacdope:
From discussion between @kaminow and I.

- Pad the list of `split_fracs` to make sure they add to one in the `sum(split_fracs) < 1` case (where we want to train on earlier data and evaluate on later data.)
- Raise an error if split fracs > 1 
- Move the chunking logic into a function of some kind.
- rename `end_splits` argument to something like `insert_index` which controls where the `sink` is placed to pad the list of `split_fracs` to sum to one.

kaminow:
@hmacdope thanks again for your helpful feedback! I believe I've implemented everything we talked about during our chat. everything still seems to be working as intended, but the code is now way less cluttered :)

PR #221, opened 2023-04-12
[DNM, depends 217] Add first production GAT model

hmacdope:
@kaminow has cut the first GAT model for use (`asapdiscovery-GAT-2023.04.12`)! It is now up on the public S3 bucket. 

This PR adds it to the pretrained models yaml file and replaces use of the test model in the docking workflow with the production one.  I tested it on a docking run I have setup locally.


hmacdope:
@kaminow @JenkeScheen this is ready for review :)

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
```diff
@@ -389,7 +389,7 @@ def main():
 
     # load ml models
     if args.gat:
-        GAT_model = GATInference("gat_test_v0")
+        GAT_model = GATInference("asapdiscovery-GAT-2023.04.12")
```
JenkeScheen, line 392: we should probably print/log the specific model string, handy for debugging.

JenkeScheen (changes requested):
Excellent! One minor comment.

asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
JenkeScheen: (even though the version will be retrievable from the .yaml, during debugging we'll stumble on the logged version before thinking of checking the .yaml)

hmacdope: @JenkeScheen should be fixed now. :)

JenkeScheen (approved):

PR #231, opened 2023-04-19
Migrate GAT model

kaminow:
Move the GAT code from `asapdiscovery.ml.models` to the [`mtenn`](https://github.com/choderalab/mtenn) package. Requires choderalab/mtenn#22 to be merged first. Resolves #228.

## Todos
  - [x] Update to use `mtenn.GAT` model
  - [x] Get rid of current `GAT` model
  - [x] Update inference to use `mtenn` model
  - [x] Update `asapdiscovery.ml.utils.load_weights` for backwards compatibility

PR #232, opened 2023-04-19
Add GAT model at newest version for migration to MTENN

hmacdope:
Updates the gat model spec and fixes two random logging issues.

Will likely have to go in before #231 and then we can update the docking workflow to the new model in #231. 

If the old models will not be backward compatible as we are pre 0.1.0 we can just chop them out.


kaminow:
as of [7fd0b85](https://github.com/choderalab/asapdiscovery/pull/231/commits/7fd0b85fa6d00446156e4269bb0c413bc1db059d), models should be backwards compatible so I don't think we need to get rid of the old weights files

hmacdope:
@kaminow you happy for us to go ahead here?

PR #235, opened 2023-04-20
Generalize the CDD downloading/processing process

kaminow:
Try to move away from code being built around our previous narrow use case (one specific search from one CDD vault, always being processed with the same fields in the same way).

## Todos
  - [x] Modularize URL in downloading process
  - [x] Split `asapdiscovery.data.utils.filter_molecules_dataframe` into two separate functions for filtering and processing
  - [x] Tests!

## Status
- [x] Ready to go


hmacdope:
@kaminow should we include this in 0.1.0? My guess is yes. You can remove it if you disagree.

asapdiscovery-data/asapdiscovery/data/tests/test_moonshot.py
```diff
@@ -0,0 +1,246 @@
...
+    # Indices with NaN values so they can be compared appropriately
+    nan_idx1 = col1.isna()
+    nan_idx2 = col2.isna()
+
+    # Chceck that nans are the same
+    if not (nan_idx1 == nan_idx2).all():
+        return False
+
+    # Check that all non-nan values are close
+    return np.isclose(col1[~nan_idx1], col2[~nan_idx2]).all()
```
hmacdope, line 246: use `np.allclose(equal_nan=True)` which should sort everything including `NaNs` out properly (I think)

hmacdope (changes requested):
Just one tiny nitpick but looking good! Love the thorough testing.

asapdiscovery-data/asapdiscovery/data/tests/test_moonshot.py
kaminow: ooh I didn't know that exists. thanks!

kaminow:
@hmacdope @ijpulidos should be ready for a review now. not really sure why that one test is failing as it passes on my machine. I can try and recreate it in the morning

hmacdope (changes requested):
@kaminow there are still some tests that are failing for non trivial reasons with mismatched expected and delivered data. You are probably best placed to find out why as I am not super familar with this code. :)

hmacdope (approved):
Thanks so much @kaminow! Great work.

PR #240, opened 2023-04-24
Put val split after training split for temporal splitting

### LINKED ISSUE #239: Temporal val split should come from earlier data
Currently, the default for `asapdiscovery.ml.utils.split_temporal` is to take the val split from the end of the dataset. We probably want to switch the default to be taking the val set right after the train set, as this more closely resembles real life (ie making future decisions based on the current data).

kaminow:
Resolves #239.


hmacdope (approved):
LGTM

PR #255, opened 2023-05-03
Add Mac1 Fragalysis download

### LINKED ISSUE #252: Fragalysis download currently encountering a server error
```
(ad-3.9) ?1 /Users/alexpayne/Scientific_Projects/covid-moonshot-ml % download-fragalysis-data -o /Users/alexpayne/Downloads/test_fragalysis

<!doctype html>
<html lang="en">
<head>
  <title>Server Error (500)</title>
</head>
<body>
  <h1>Server Error (500)</h1><p></p>
</body>
</html>

  File "/Users/alexpayne/opt/anaconda3/envs/ad-3.9/bin/download-fragalysis-data", line 8, in <module>
    sys.exit(main())
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py", line 20, in main
    download(args.o, args.x)
  File "/Users/alexpayne/Scientific_Projects/covid-moonshot-ml/asapdiscovery-data/asapdiscovery/data/fragalysis.py", line 47, in download
    url_dl = r.text.split(':"')[1].strip('"}')
IndexError: list index out of range
```

kaminow:
## Description
Now we can download the Mac1 data as well as Mpro. Also fix an issue that was found in the Mpro API call (closes #252)

## Todos
  - [x] Add Mac1 API call
  - [x] Add target CLI arg for `download_fragalysis_data.py`
  - [x] Fix issue with the Mpro API call

## Status
- [x] Ready to go


hmacdope (changes requested):
Add a test so we can see then it breaks. Other than that looks great.

ijpulidos (changes requested):
Changes look good, thanks! I agree with the comment on making a test for this. Just made a quick comment on something that I think we should fix.

asapdiscovery-data/asapdiscovery/data/fragalysis.py
```diff
@@ -9,41 +9,65 @@
...
+    "cif_info": False,
+    "NAN2": False,
+    "map_info": False,
+    "single_sdf_file": True,
+    "sdf_info": True,
+    "pdb_info": False,
+    "bound_info": True,
+    "metadata_info": True,
+    "smiles_info": True,
+    "static_link": False,
```
ijpulidos, line 49: What is the difference between these two dictionaries? It seems like they share a lot of data and maybe what we want is not having a dictionary for every target but actually have some sort of sane defaults and change the dictionary as needed "on the fly", for example the `target_name` should change but probably many of the other options should stay the same.

kaminow: oh yeah that's a good point, they should be identical other than the `target_name` field. do you want to handle this in your changes or should I?

ijpulidos: I can take care of these changes

ijpulidos:
We have some tests in `test_docking.py` failing here. As far as I can tell the changes in this PR shouldn't affect these tests. @hmacdope since you wrote these tests any ideas why these are failing here?

asapdiscovery-data/asapdiscovery/data/fragalysis.py
```diff
@@ -6,7 +6,7 @@
 
 BASE_URL = "https://fragalysis.diamond.ac.uk/api/download_structures/"
 # Info for the POST call
-MPRO_API_CALL = {
+API_CALL_BASE = {
     "target_name": "Mpro",
```
kaminow, line 10: @ijpulidos should we change this to be `None` or an empty string or something?

ijpulidos: Done in https://github.com/choderalab/asapdiscovery/pull/255/commits/ed38a0311e293c5bacf2c0dfd293448cb8f3177e

hmacdope:
@ijpulidos #259 should have fixed the issue, ive merged main in so hopefully we will see all green in a second.

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
```diff
@@ -17,7 +23,10 @@ def get_args():
 def main():
     args = get_args()
 
-    download(args.o, args.x)
+    # Overwrite the base target with the cli-specified target
+    API_CALL_BASE["target_name"] = args.target.capitalize()
```
hmacdope, line 27: can't we make a copy rather than overwriting some base struct.

asapdiscovery-data/asapdiscovery/data/tests/test_fragalysis.py
```diff
@@ -0,0 +1,69 @@
...
+import pytest
+from asapdiscovery.data import fragalysis
+
+
+@pytest.fixture
+def mpro_fragalysis_api_call():
+    """Fragalysis API call for downloading target data"""
+    from asapdiscovery.data.fragalysis import API_CALL_BASE
+
+    api_call = API_CALL_BASE
```
hmacdope, line 14: same here make a copy?

asapdiscovery-data/asapdiscovery/data/tests/test_fragalysis.py
```diff
@@ -0,0 +1,69 @@
...
+    api_call["target_name"] = "Mpro"
+    return api_call
+
+
+@pytest.fixture
+def mac1_fragalysis_api_call():
+    """Fragalysis API call for downloading target data"""
+    from asapdiscovery.data.fragalysis import API_CALL_BASE
+
+    api_call = API_CALL_BASE
```
hmacdope, line 24: and here?

hmacdope (changes requested):
Nitpick on making a copy of `API_CALL_BASE` rather than setting fields on base struct, but mostly a nitpick.

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
ijpulidos: Sounds good, I agree that would be a better approach. Will make the changes soon.

ijpulidos: Done in https://github.com/choderalab/asapdiscovery/pull/255/commits/e611ba001e35cea7a1577566737b63d335fd23ef

ijpulidos:
@hmacdope Rready to be re-reviewed, I made the changes you suggested. Thanks!

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
```diff
@@ -17,7 +24,11 @@ def get_args():
 def main():
     args = get_args()
 
-    download(args.o, args.x)
+    # Copy the base call and update the base target with the cli-specified target
+    api_call = copy.deepcopy(API_CALL_BASE)
+    api_call["target_name"] = args.target.capitalize()
+
+    download(args.o, API_CALL_BASE, args.x)
```
hmacdope, line 31:
```suggestion
    download(args.o, api_call, args.x)
```

hmacdope (changes requested):
I think you need to use the copied object here but could be missing something, see my suggestion.

hmacdope (changes requested):
I think you need to use the copied object here but could be missing something, see my suggestion.

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
ijpulidos: Yes, totally! This is what happens when you try to make changes in the middle of a talk in a conference :)

hmacdope (approved):
LGTM!

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
```diff
@@ -26,6 +26,8 @@ def main():
 
     # Copy the base call and update the base target with the cli-specified target
     api_call = copy.deepcopy(API_CALL_BASE)
+    if args.target.lower() not in {"mpro", "mac1"}:
+        raise ValueError("Target must be one of [mpro, mac1].")
```
ijpulidos, line 30: I don't know if it's a good idea that we limit ourselves here, is there a special reason why we want to do this? I'm just trying to avoid having to remember adding every target manually here when we get them in fragalysis.

ijpulidos: When the target is not found on fragalysis an `HTTPError`will be raised from https://github.com/choderalab/asapdiscovery/blob/6ed0ed76ef830ab23a26dd58a7af759a557916b2/asapdiscovery-data/asapdiscovery/data/fragalysis.py#L49 just in case that was the reason

ijpulidos: This isn't a blocker for me, but just wanted to mention it.

kaminow: yeah it's a fair point. my motivation is just that the `HTTPError` doesn't give a reason for the failure, just that it fails. eg:
```python
  File "/home/kaminowb/miniconda3/envs/asapdiscovery_dev/bin/download-fragalysis-data", line 8, in <module>
    sys.exit(main())
  File "/lila/data/chodera/kaminowb/moonshot_ml_dev/asapdiscovery_dev/asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py", line 33, in main
    download(args.o, api_call, args.x)
  File "/lila/data/chodera/kaminowb/moonshot_ml_dev/asapdiscovery_dev/asapdiscovery-data/asapdiscovery/data/fragalysis.py", line 49, in download
    raise requests.HTTPError(
requests.exceptions.HTTPError: Post request to https://fragalysis.diamond.ac.uk/api/download_structures/ failed with 404 error code, using the following API call {'target_name': 'Sdf', 'proteins': '', 'event_info': False, 'sigmaa_info': False, 'diff_info': False, 'trans_matrix_info': False, 'NAN': False, 'mtz_info': False, 'cif_info': False, 'NAN2': False, 'map_info': False, 'single_sdf_file': True, 'sdf_info': True, 'pdb_info': False, 'bound_info': True, 'metadata_info': True, 'smiles_info': True, 'static_link': False, 'file_url': ''}.
```
as you say though, it might not be good to limit, especially since there are other targets on Fragalysis that we're not working with

ijpulidos: I see, I don't have a strong opinion about this, I think we can have this limit in the meantime. Just wanted to mention that we are returning the full API call used when we get the `HTTPError` hoping that users can debug it if needed.

On the other hand, I think we can have the option/choices limit directly using `ArgParser`, which I think it's a bit more elegant. I'll be making those changes soon.

ijpulidos: Done in https://github.com/choderalab/asapdiscovery/pull/255/commits/732ad26b9e6f79ac8e693f8dbf9b2fd58672bacd

ijpulidos (approved):
LGTM!

asapdiscovery-data/asapdiscovery/data/scripts/download_fragalysis_data.py
kaminow: oh perfect, thanks!

PR #261, opened 2023-05-08
Prototype schematization of repo

apayne97:
## Description
Write some ideas for how a schematized version of the repo would look in the top level `prototypes` directory.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


hmacdope:
Thanks for pushing, we will keep this as a prototype and use `schematization_v1` branch as our WIP branch so we can always refer back here without it always chopping and changing.

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+ligands = ligand_fr.load()
+targets = target_fr.load()
+
+lp: LigandPrepper = LigandPrepper(ligands)
+prepped_ligands: List[Ligand] = lp.prep_all()
+
+pp: ProteinPrepper = ProteinPrepper(targets)
+prepped_targets: List[Target] = pp.prep_all()
+
+ligand_selector: LigandSelector = PairwiseLigandSelector(prepped_ligand, prepped_target)
```
dotsdl, line 78: @hmacdope can you add a comment above noting that we would choose different `Selector` subclasses to use different selection schemes? In the case here, `PairwiseLigandSelector` does the simple product of the ligands x targets, but others could be a bit "smarter".

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+        prep_t = prep_target(ligand: Ligand)
+        prepped_targets.append(prep_t)
+
+    prepped_ligands = []
+    for ligand in ligand_fr.load():
+        prep_l = prep_ligand(ligand: Ligand)
+        prepped_ligands.append(prep_l)
+
+    # barrier: could make parallel here as well possibly as do not rely on collected results
+    ligand_selector: LigandSelector = PairwiseLigandSelector(prepped_ligands, prepped_targets)
```
dotsdl, line 47: You could do a:

```suggestion
    ligand_selector: LigandSelector = delayed(PairwiseLigandSelector)(prepped_ligands, prepped_targets)
```
 and then `ligand_selector` is also a `Delayed` object. Any subsequent calls on that object will also give back `Delayed` objects.

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+    docking_inputs: List[DockingInput] = ligand_selector.select()
+
```
dotsdl, line 49: In this example, you would pass your `docking_inputs` `Delayed` object to a `dask` executor, and get back actual `DockingInput` objects for the docking steps below. It's okay that this is a bottleneck, since the `Selector` determine how much work we'll actually do for the more expensive docking step below.

prototypes/mcs_docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+
+    prepped_ligands = []
+    for ligand in ligand_fr.load():
+        prep_l = prep_ligand(ligand: Ligand)
+        prepped_ligands.append(prep_l)
+
+    # true barrier as need to have all ligands and all targets prepped to choose pairs
+    # actually is this even true? We only need the list of TARGETS to know which are the best
+    ligand_selector: LigandSelector = MCSLigandSelector(prep_l, prep_t)
+    docking_inputs: List[DockingInput] = ligand_selector.select()
```
apayne97, line 49: the MCSS-based selector just needs to know the SMILES String of the reference and query ligands, so we wouldn't necessarily need to prep them first!

apayne97:
> Thanks for pushing, we will keep this as a prototype and use schematization_v1 branch as our WIP branch so we can always refer back here without it always chopping and changing.

yeah I like the idea of keeping this as a kind of drawing board!

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+lp: LigandPrepper = LigandPrepper(ligands)
+prepped_ligands: List[Ligand] = lp.prep_all()
+
+pp: ProteinPrepper = ProteinPrepper(targets)
+prepped_targets: List[Target] = pp.prep_all()
+
+ligand_selector: LigandSelector = PairwiseLigandSelector(prepped_ligand, prepped_target)
+docking_inputs: List[DockingInput] = ligand_selector.select()
+
+docker: Docker_OE = Docker_OE(docking_inputs)
```
hmacdope, line 81: type of docking could also be handled with a kwarg @ijpulidos

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+    # barrier: could make parallel here as well possibly as do not rely on collected results
+    ligand_selector: LigandSelector = PairwiseLigandSelector(prepped_ligands, prepped_targets)
+    docking_inputs: List[DockingInput] = ligand_selector.select()
+
+    docking_results = []
+    for docking_input in docking_inputs:
+        dr = docking_wf(docking_input)
+        docking_results.append(dr)
+
+    # blah blah blah
```
hmacdope, line 55:
```
docking_results.compute()
```

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
+from dask import delayed
+
+# inputs to the delayed functions are
+# probably length 1 lists of the objects
+
+@delayed
+def prep_targets(target: List[Target]):
```
hmacdope, line 7: Add caching for previously prepped targets? @apayne97

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+
+# main
+
+def main():
+    ligand_fr = FileReader(ligand_fn, type="ligand", format='sdf')
+    target_fr = FileReader(target_fn, type="target", format='pdb')
+
+    prepped_targets = []
+    for target in target_fr.load():
+        prep_t = prep_target(ligand: Ligand)
```
hmacdope, line 38: This takes 3-5 minutes, ruh, ruh.

prototypes/docking_wf_atomised.py
```diff
@@ -0,0 +1,82 @@
...
+target_fr = FileReader(target_fn, type="target", format='pdb')
+
+ligands = ligand_fr.load()
+targets = target_fr.load()
+
+lp: LigandPrepper = LigandPrepper(ligands)
+prepped_ligands: List[Ligand] = lp.prep_all()
+
+pp: ProteinPrepper = ProteinPrepper(targets)
+prepped_targets: List[Target] = pp.prep_all()
```
hmacdope, line 76: Serialisable `Target` can be used to cache the prepping.

ijpulidos: We should have a way to easily deserialize the objects to OEdesingUnits or the object we use in memory for docking.

PR #262, opened 2023-05-09
Refactor of the docking pipeline with far better schematization

kaminow:
## Description
Large rework of the docking pipeline/data structures to make heavy use of schema.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


apayne97:
Seems like we might want to split up the refactor into a few discrete goals and then have each of those goals be a PR into this branch

hmacdope:
> Seems like we might want to split up the refactor into a few discrete goals and then have each of those goals be a PR into this branch

Yes! I will come up with a strategy and we can debate it out recurring meeting I will set up.

PR #264, opened 2023-05-09
Fix issue 260

### LINKED ISSUE #260: Investigate why DGL 1.1 breaks our weight comparisons for constructing GAT model
Investigate why DGL 1.1 breaks our weight comparisons for constructing GAT model

kaminow:
Add new GAT weights that are compatible with the new version of `dgl`. Closes #260.

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/scripts/run_docking_oe.py
@@ -551,7 +551,7 @@ def main():
     logger.info(f"Loaded {n_mols} ligands, proceeding with docking setup")
 
     # Set up ML model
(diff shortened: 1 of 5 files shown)


hmacdope (changes requested):
Can you also change the use of the old models in testing and in the run-docking-oe script? Also just  add a comment to the models yaml file about DGL compatibility

kaminow:
oh yep, good call on both points

kaminow:
@hmacdope ok I think I've done everything correctly, and the tests are green now which is a good sign. I wasn't sure whether to just update the old weights file on S3 or add a new one so I just added a new one to be safe, but feel free to adjust as you see fit

hmacdope (approved):
LGTM, thanks!

hmacdope:
We need to fix the tolerance in some comparison, I'll do it one sec.

PR #238, opened 2023-04-24
Generalize protein prep and delete old protein prep scripts

apayne97:
## Description
Some improvements to generalize protein preparation

Towards #234 

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] make separate openeye biological assembly function
  - [x] move modeling functions from docking to modeling.modeling
  - [x] move du_to_complex to data.openeye
  - [x] make CrystalCompoundDataset a serializable class (to csv file)
  - [x] create separate script for parsing input data into CrystalCompoundDataset for protein preparation (prepare_protein_csv.py)
  - [x] make second generation protein prep script that reads csv file (prep_proteins_v2.py)
  - [x] get the `prepare-protein-csv` --> `prep-proteins-v2` pipeline to work
  - [x] add tests to repo
  - [x] #272 
  - [x] remove protein_only flag from prepare_protein_csv.py --> how do we make sure to include the OpenEye formatted site then?
  - [x] make single prep workflow that can be tested in pytest
  - [x] use superpose_molecule instead of align_receptor and delete align_receptor to avoid having the split_initial_complex functionality during alignment, also include the predicate selection within superpose molecule
  - [x] delete prep_receptor, prep_mp, and prep_fragalysis_dus, and prep_pdb_structures.py
  - [x] delete CrystalCompoundData properties: oligomeric_state, protein_chains, series
  - [x] rename CrystalCompoundData field `active_site` to `oe_active_site_str`
  - [x] delete old protein prep scripts
  - [x] #189 
  - [x] #251 
  - [x] add back logging 

## Notes for future schematization
- the ligand properties (SMILES, series, location of cached object, etc) should be stored within their own class

## Status
- [x] Ready to go

PR #272, opened 2023-06-01
Convert protein prep tests to be pythonic

apayne97:
## Description
Necessary to create a testing framework for finishing #238.

There are definitely some issues with the underlying code itself which are highlighted by the testing suite being a bit complicated but I want to create the framework for testing first and then use that to ensure the changes we make will help.

I'm also trying to focus more on the underlying functions rather than scripts, which has also helped clarified what changes we might want to make to the schema going forward.

## Todos
  - [x] deleted old test prep versions
  - [x] converted test to pytest
  - [x] added a number of resources to AWS to be fetched by Pooch
  - [x] add a function in the simulation module to test to make sure a forcefield can be generated for the prepared system
  - [x] enable saving results of test to local directory for further validation
  - [x] #274 
  - [x] move simulation test to its own useful function
  - [x] test the entire workflow for sars / mers as its own function

## To NOT Do in this PR:
  -  #275 

## Questions
- [x] Currently this test takes almost 5min because I'm actually prepping the proteins...in the future it might be nice to be able to disable these tests for PRs that have nothing to do with this module?

## Status
- [x] Ready to go


ijpulidos:
We need to add the tests on this subpackage in the `.github/workflows/CI.yaml`, so that would mean adding a line like 
```yaml
coverage run --append --source=./asapdiscovery-modeling/asapdiscovery/modeling/ -m pytest -v  --color=yes ./asapdiscovery-modeling/asapdiscovery/modeling/tests/
```

apayne97:
> We need to add the tests on this subpackage

Done!

ijpulidos:
Wait, why are the tests not getting run?

ijpulidos:
Oh, I see, this is for a non-main branch, that's why

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
```diff
@@ -0,0 +1,284 @@
...
+
+        dataset.structures[0].active_site_chain = "B"
+        assert dataset != xtal_dataset
+
+
+class TestProteinPrep:
+    def test_sars_protein_prep(
+        self, sars_xtal, ref, prepped_files, loop_db, ref_chain="A"
+    ):
+        xtal = sars_xtal
```
hmacdope, line 107: Similar idea here about easy decoupling of test and workflow, we want to test a function or script, If we have a currently optimal prep **script** you can use something like:

```python
@pytest.fixture()
def make_output_dir_and_cleanup():
    # create output dir
    os.makedirs("./outputs", exist_ok=True)
    yield
    # clean up
    shutil.rmtree("./outputs")


@pytest.mark.timeout(400)
@pytest.mark.script_launch_mode("subprocess")
def test_single_target_docking(
    script_runner,     make_output_dir_and_cleanup,
):
    args = [
        "prep-proteins",
        "-o",
        "output"
        arg2,
        arg3,
    ]
    ret = script_runner.run(*args)
    assert ret.success

```

Otherwise you can make this logic a function, test the function and use the function inside of a script if that makes sense?

So like 
```python 
# in module
def  protein_prep():

# in script 
...
protein_prep(...)
...

# in tests 
def  test_protein_prep():

```

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
```diff
@@ -0,0 +1,284 @@
...
+        save_openeye_sdf(lig, str(lig_fn))
+
+        for fn in [prot_fn, complex_fn, lig_fn]:
+            assert Path(fn).exists()
+            assert Path(fn).is_file()
+
+        # TODO: Add a test to make sure the ligand is in the active site
+        # TODO: Add a test to make sure the seqres has been added to the protein
+
+    def test_mers_protein_prep(
```
hmacdope, line 189: The sars and mers tests can be merged with a `parametrize()` decorator as they test the same functionality.

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
```diff
@@ -0,0 +1,284 @@
...
+        assert Path(prot_fn).is_file()
+
+    @pytest.mark.parametrize(
+        ("du_fn", "has_lig"),
+        [
+            ("Mpro-P2660_0A_bound-prepped_receptor_0.oedu", True),
+            ("rcsb_8czv-assembly1-prepped_receptor_0.oedu", False),
+        ],
+    )
+    def test_openeye_du_loading(self, du_fn, has_lig, prepped_files):
```
hmacdope, line 259: This tests these two specific files work, but would it be better to have a utility function to test the output of one of the prep_workflows themselves? Let me know if thats not clear.

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
```diff
@@ -0,0 +1,284 @@
...
+    )
+
+
+@pytest.fixture
+def xtal_dataset(sars_xtal, mers_xtal):
+    xtal_dataset = CrystalCompoundDataset(structures=[sars_xtal, mers_xtal])
+    return xtal_dataset
+
+
+class TestCrystalCompoundDataset:
```
hmacdope, line 82: These are really good just testing the simple functionality is very important.

hmacdope (changes requested):
Great work @apayne97! Testing this kind of stuff is not easy, but is super important. See my comments and we can work through them. :)

hmacdope (changes requested):
Great work @apayne97! Testing this kind of stuff is not easy, but is super important. See my comments and we can work through them. :)

hmacdope (changes requested):
Great work @apayne97! Testing this kind of stuff is not easy, but is super important. See my comments and we can work through them. :)

hmacdope:
Whoops double review.

hmacdope:
@apayne97 I forgot to respond to your questions:

 > In the future I'd like to change the tests a bit to compare the output to the file saved in AWS (i.e. if they are not the same, then the test fails) that way the tests can be decoupled more than they are now. Does this need to be done now or can it wait?

Yep thats a great idea. This is a good enough start I would say, but if you have a handy file go ahead. Another way is if you can programatically interrogate the result.

>  Currently this test takes almost 5min because I'm actually prepping the proteins...in the future it might be nice to be able to disable these tests for PRs that have nothing to do with this module?

I don't know if there is an easy way to do this as would require static analysis of all code paths to know what affects what. Prefer a long test suite than none at all :man_shrugging:

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
hmacdope: we discussed making this a `prep_sars` function.

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_protein_prep.py
hmacdope: and `prep_mers` here

PR #297, opened 2023-06-07
Speed up tests

mikemhenry:
## Description
Setup cache for micromamba and clean on merge.
Use `pytest-xdist` to run tests in parallel. 

## Status
- [ ] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -78,10 +78,10 @@ jobs:
         # run each package test suite; append to coverage file
         # Exit immediately if a command exits with a non-zero status.
         set -e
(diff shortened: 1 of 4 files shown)


mikemhenry:
This is ready for review! I found without `set -e` CI would show a green check even if pytest had a failure. I found that using xdist on the ml tests caused some issue with downloading, guessing that it had to do with some race condition when tests would run at the same time. I also was getting a weird issue with the fixture `make_output_dir_and_cleanup` and made one I think that is a bit better since it doesn't rely on hard-coding a path and will use a tmp_dir so it will be unique.

hmacdope:
@mikemhenry thanks so much, this is an awesome speedup (4-5x) and will help massively!

hmacdope (approved):

PR #246, opened 2023-04-27
Project support docking

hmacdope:
Fixes #234 

## Description
Adds a single reference prep+dock+MD+visualise workflow for an arbitrary structure. 

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [x] Clarify if MCS step needed 
  - [x] testing

PR #310, opened 2023-06-09
Add postera support for single-target docking wf. 

### LINKED ISSUE #258: Add pull and push to postera API on either end of docking pipeline.
For single target docking we should be able to push and pull to a postera molecule set.

hmacdope:
## Description

Fixes #258 
Branches off #246 

- Adds ability to push and pull from postera on either end of docking pipeline. 
- Adds filtering of up-loadable custom data to postera to a set of pre defined tags. 
- 

## Todos
  - [ ]  Now is one of the best times to fully validate SD tags and postera columns so lets try and get that sorted before we push this too much further.


asapdiscovery-data/asapdiscovery/data/postera/manifold_data_validation.py
```diff
@@ -0,0 +1,66 @@
...
+
+import pandas as pd
+
+
+class ManifoldAllowedColumns(Enum):
+    """
+    Enum of allowed columns for the P5 comp-chem team to update in postera.
+    """
+
+    SMILES = "SMILES"
```
jchodera, line 11: Do we actually refer to these columns via `ManifoldAllowedColumns.ASAP_VC_ID`, etc? If not, does it make sense to hard-code these ads defined values?

Alternatively, would a YAML file work for this, where instead of programmatically coding this, we have a flexible data-driven encoding of canonical names (`ASAP_VC_ID`) and SD tags (`asap_vc_id`), though I confess I'm not sure which name is supposed to be the SD tag and which is supposed to be the Python variable name that is extracted.

From a general philosophy standpoint, it's helpful to think about this as aiming to produce code that is reusable for other projects when possible, so that avoiding hard-coding project-specific choices in code that could instead be flexibly defined in configuration files (such as `project.yaml`) will make it easier for others to re-use the code.

Finally, I think we still need to align on SD tag names that parallel the CDD Vault protocol name structure, as well as are unique, relatively stable, extensible, and obvious to the chemists. For example, the chemists may not know what `chemgauss4` or `GAT` is. We may also want to use standard virus nomenclature---there is no `SARS2` or `MERS`, but `SARS-CoV-2`, `MERS-CoV`, `DENV`, `ZIKV`, `WNV`, etc. The target name may also be relevant. We're working on standardizing those target names [here](https://asapdiscovery.slack.com/archives/C03GYK0PHB4/p1684423929023879).

hmacdope: Thanks John! should be easy enough to change to YAML files, I'll raise an issue while we continue to work on the SD tag names themselves.

hmacdope:
Dropping from milestone until we sort out our SD tag and naming plan more consistently.

asapdiscovery-data/asapdiscovery/data/utils.py
```diff
@@ -1455,6 +1455,20 @@ def combine_files(paths: list[Union[Path, str]], output_file):
                 ofs.write(file_to_copy_fd.read())
 
 
+def target_names_from_common_names_and_crystals(name: str) -> str:
```
hmacdope, line 1458: This should probably be YAMLified (TM) as well.

asapdiscovery-data/asapdiscovery/data/postera/manifold_data_validation.py
JenkeScheen: also it should be noted here that we're not dealing with `ASAP_VC_ID`s, but instead with `UUID_POSTERA_API`s. `ASAP_VC_ID`s don't get assigned until designs are submitted to a `design set` in the postera portal, and that currently happens after any computational modelling has been done. If this seems confusing (it is) I'm happy to explain further.

hmacdope:
#348 will have to go in first.

hmacdope:
@jchodera would you be able to have another look here? Pulling from and pushing to a molecule set seems to be working as expected. 

I have added in a YAML config `manifold_data_dags.yaml` that  defines our tag spec. From that the tags of the form 

```
 <output>_<tool>_<unit?>_<organism>_<target>
```
are made programatically, where <unit?> is only included if the output has a unit.
for example the final tags look like:
```
- Docking_Score_POSIT_sars2_Mpro
- ML_Score_GAT_pIC50_sars2_Mac1
- MD_Pose_mers_Mpro
- MD_Interactions_mers_Mpro
- ...
```
Importantly these can be easily changed with a quick change to the YAML file. 

 I have built the tags adhering to the SD tag document with the exception of moving the units to after the tool rather than at the end per our discussion at last CCC meeting IIRC. 

Let me know what you think! Importantly any tweaks should be easy due to YAML.

asapdiscovery-docking/asapdiscovery/docking/docking_data_validation.py
```diff
@@ -0,0 +1,132 @@
...
+        ]
+        print("COLS: ", cols)
+        if not ManifoldFilter.all_valid_columns(cols):
+            raise ValueError(
+                f"Columns in dataframe {cols} are not all valid for updating in postera. Valid columns are: {ManifoldAllowedTags.get_values()}"
+            )
+        return cols
+
+
+def drop_non_output_columns(
```
dotsdl, line 67: It feels like a lot of this logic also is encoded in `ManifoldFilter`; consider ways to simplify.

asapdiscovery-docking/asapdiscovery/docking/docking_data_validation.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # drop all columns that are not in the output
+    df = df.drop(columns=[col for col in df.columns if col not in output_cols])
+    return df
+
+
+def rename_output_columns_for_target(
+    df: pd.DataFrame, target: str, manifold_validate: Optional[bool] = True
+) -> pd.DataFrame:
+    """
```
dotsdl, line 89: I don't understand what this docstring is saying.

dotsdl: Should include "inject target name" etc.

asapdiscovery-docking/asapdiscovery/docking/docking_data_validation.py
```diff
@@ -0,0 +1,132 @@
...
+            for col in cls
+            if ManifoldFilter.is_allowed_column(col.value + f"_{target}")
+        ]
+        print("COLS: ", cols)
+        if not ManifoldFilter.all_valid_columns(cols):
+            raise ValueError(
+                f"Columns in dataframe {cols} are not all valid for updating in postera. Valid columns are: {ManifoldAllowedTags.get_values()}"
+            )
+        return cols
+
```
dotsdl: These methods aren't used anywhere in the codebase. Either remove or use them to simplify code elsewhere.

dotsdl (approved):
Great work @hmacdope! Good to merge!

asapdiscovery-docking/asapdiscovery/docking/docking_data_validation.py
hmacdope: cleaned up by removing ManifoldFilter.

asapdiscovery-data/asapdiscovery/data/postera/molecule_set.py
```diff
@@ -261,3 +290,39 @@ def update_molecules(
...
+        if debug_df_path is not None:
+            df.to_csv(debug_df_path, index=False)
+
+        # make payload for postera
+        mol_update_list = MoleculeUpdateList.from_pandas_df(
+            df, smiles_field=smiles_field, id_field=id_field
+        )
+
+        # push updates to postera
+        return self.update_molecules(
```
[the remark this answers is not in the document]
hmacdope: I think might push ahead with out this for now but will raise an issue.

asapdiscovery-data/asapdiscovery/data/postera/manifold_data_validation.py
```diff
@@ -0,0 +1,227 @@
...
+    final_tags = sorted(final_tags)
+    return TagEnumBase.from_iterable("ManifoldAllowedTags", final_tags), final_tags
+
+
+ManifoldAllowedTags, _ = make_tag_combinations_and_combine_with_static(
+    target_tag_set, output_tag_set, static_tag_set
+)
+
+
+class ManifoldFilter:
```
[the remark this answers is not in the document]
hmacdope: I have simplified this a lot.

PR #332, opened 2023-06-21
Add SchNet pIC50 output

### LINKED ISSUE #257: Units and settings for Schnet inference
Currently we just have Schnet_score as the tag, we should use the proper units and/or audit the settings we use for the input.
@kaminow do you have an idea of what settings we should be using for the model? We can always do more than one.

hmacdope:
Fixes #257 

Adds correct kwargs for schnet inference to be in right units.


kaminow (approved):
looks great, thanks for taking care of this!

Issue #327, opened 2023-06-19
Each module should capture its own openeye logs. 


hmacdope:
Setting the openeye log output streams in multiple places means they trip over each other in ways that are hard to predict. 

If you need to capture openeye logs from a process in a file, it should be  the responsibility of that module to manage the logging resource RAII style.  This means flushing and closing the resource also. 

For example  

```python
def My_workflow():
 
    errfs = oechem.oeofstream("openeye_My_Workflow_log.txt")
    oechem.OEThrow.SetOutputStream(errfs)
    oechem.OEThrow.SetLevel(oechem.OEErrorLevel_Debug)
    oechem.OEThrow.Info(f"Starting docking for {logname}")
    oechem.OEThrow.Debug("Confirm that OE logging is working")
    
    ... # do stuff
    ... # we are done
   errfs.flush()
   errfs.close()
   return blahblah
```
Hopefully this helps get the logs to be more in one place and predictable.

hmacdope:
Impossible due to #281

PR #336, opened 2023-06-23
Allow openeye loading and saving functions to take str or Path, and improve typing. 

### LINKED ISSUE #315: make all loading / saving functions able to take a Path object as input
i.e. load_openeye_sdf, etc 

OpenEye has to use strings

hmacdope:
Fixes #315 

 - Allow openeye loading and saving functions to take str or Path
 - Improve type hinting.

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/openeye.py
+++ b/asapdiscovery-data/asapdiscovery/data/openeye.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from typing import Union, List, Optional # noqa: F401
 
 from openeye import oechem, oedepict, oedocking, oegrid, oeomega, oespruce  # noqa: F401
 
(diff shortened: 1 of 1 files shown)


asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -293,15 +311,39 @@ def load_openeye_sdfs(sdf_fn):
...
+
+    """
+    if not Path(du_fn).exists():
+        raise FileNotFoundError(f"{du_fn} does not exist!")
     du = oechem.OEDesignUnit()
-    retcode = oechem.OEReadDesignUnit(du_filename, du)
+    retcode = oechem.OEReadDesignUnit(str(du_fn), du)
     if not retcode:
-        raise RuntimeError(f"Unable to read design unit from {du_filename}")
+        oechem.OEThrow.Fatal(f"Unable to open {du_fn}")
```
apayne97, line 342: This is maybe its own issue / PR, but have you confirmed that all the scripts / code that uses this can handle the openeye error logging correctly? I guess this is #327

apayne97 (approved):
LGTM! If tests are passing then you're good. I love this, been wanting to do this for a while

asapdiscovery-data/asapdiscovery/data/openeye.py
hmacdope: This is only used in single target workflow so far, but its a good point. My instinct is that if you don't meddle with the OE streams (and I think we found a solution to do that anyway) this should be OK.

PR #338, opened 2023-06-23
Add schema for Ligands 

### LINKED ISSUE #333: Schema for Ligands
Create a schema that allows for ligand information, ids, coordinates and metadata to be easily passed around.

This will basically be a thin wrapper around a string of an SDF, which should then be manipulated as an OEMol.

hmacdope:
Fixes #333 

## Description
Adds a schema to ship small molecule data around, is basically a thin wrapper around an SDF that is designed to be manipulated as an OEMol and then re-constructed. 

@jchodera and @dotsdl it would be great to get some feedback on what other kinds of identifiers should be included in the `LigandIdentifiers` subschema. 

Please give your unvarnished feedback.  :)


jthorton:
Some of [these identifiers ](https://github.com/openforcefield/openff-qcsubmit/blob/9ddddc1041d3da0eeacc2fd974daed6fc77258cd/openff/qcsubmit/common_structures.py#L809) might be handy that we use in our quantum chemistry database. These allow you to do some filtering and searching without having to convert every molecule to OE/RDKit objects first which is great when dealing with a lot of molecules.

hmacdope:
> Some of [these identifiers ](https://github.com/openforcefield/openff-qcsubmit/blob/9ddddc1041d3da0eeacc2fd974daed6fc77258cd/openff/qcsubmit/common_structures.py#L809) might be handy that we use in our quantum chemistry database. These allow you to do some filtering and searching without converting every molecule to OE/RDKit objects, which is great when dealing with many molecules. We also found that the mapped smiles with explicit hydrogens are a good minimal representation of the molecules which can be reconstructed with a consistent atom ordering which is great if you also need to save the coordinates.

Thanks so much @jthorton! Very good suggestions here. I might also run a quick trial where we use openFF molecules to handle the interoperability and would love some feedback on that. I'll keep you posted.

hmacdope:
@JenkeScheen @dotsdl @kaminow @apayne97 @jchodera this is probably ready for a first proper review now 

Now has 

- SD tag manipulation in embedded SDF data 
- manifold_vc_id and manfold_api_id fields for tracking things through postera.  
-  UUID capacity for a prospective comp chem only database reference. 
- Thorough tests :)

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+    manifold_api_id: Optional[UUID] = Field(
+        None, description="Unique ID from Postera Manifold API"
+    )
+    manifold_vc_id: Optional[str] = Field(
+        None, description="Unique VC ID (virtual compound ID) from Postera Manifold"
+    )
+    compchem_id: Optional[UUID4] = Field(
+        None, description="Unique ID for P5 compchem reference"
+    )
```
apayne97, line 53: LOVE this.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,151 @@
...
+class Ligand(DataModelAbstractBase):
+    """
+    Schema for a Ligand
+    """
+
+    compound_name: str = Field(None, description="Name of compound")
+    ids: Optional[LigandIdentifiers] = Field(
+        None,
+        description="LigandIdentifiers Schema for identifiers associated with this ligand",
+    )
```
apayne97, line 65: I think we would want to make at least 1 ID a requirement?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,151 @@
...
+    data: str = Field(
+        "",
+        description="SDF file stored as a string to hold internal data state",
+        repr=False,
+    )
+    data_format: DataStorageType = Field(
+        DataStorageType.sdf,
+        description="Enum describing the data storage method",
+        allow_mutation=False,
+    )
```
apayne97, line 79: is there a reason these can't go into the baseclass?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def set_SD_data(self, key: str, value: str) -> None:
+        mol = sdf_string_to_oemol(self.data)
+        mol = set_SD_data(mol, key, value)
+        self.data = oemol_to_sdf_string(mol)
+
+    def set_SD_data_dict(self, data: dict[str, str]) -> None:
+        mol = sdf_string_to_oemol(self.data)
+        mol = set_SD_data_dict(mol, data)
+        self.data = oemol_to_sdf_string(mol)
```
apayne97, line 135: Can we simplify this by just always passing a dict?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,151 @@
...
+        mol = sdf_string_to_oemol(self.data)
+        return get_SD_data(mol, key)
+
+    def get_SD_data_dict(self) -> dict[str, str]:
+        mol = sdf_string_to_oemol(self.data)
+        return get_SD_data_dict(mol)
+
+    def print_SD_Data(self) -> None:
+        mol = sdf_string_to_oemol(self.data)
+        print_SD_Data(mol)
```
apayne97, line 147: I wonder if we should also have a method to update the SD tags with all the properties of the molecule? Basically a chemist-readable serialization of the Ligand object would have to include all the properties of the object itself, right? Not sure how useful that will be in practice but I think we should maybe stick to one serialization format that includes all the data? idk

apayne97 (approved):
So many new methods and tests!

apayne97 (changes requested):
I approved it but I guess I did leave some comments I think we should discuss so I'm rolling it back to request changes

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Seems reasonable, I will move `compound_name` to `LigandIdentifiers` for this purpose.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Yes because it is set to `DataStorageType.sdf` as a constant in this subclass, with `allow_mutation=False`. `Targets`  in #339 use the other enum member `DataStorageType.pdb` immutably in the same way. :)

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: yep!

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: Oh, right of course! Sorry late night brain. Does it make sense to make these two methods abstractmethods in the base class though?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,223 @@
...
+        write_SD_attrs : bool, optional
+            Whether to write out SD tags as attributes, by default True
+
+        """
+        if write_SD_attrs:
+            self.flush_attrs_to_SD_data()
+        # directly write out data
+        write_file_directly(filename, self.data)
+
+    def set_SD_data(self, data: dict[str, str]) -> None:
```
jthorton, line 170: Is there a reason to store these in the sdf string rather than the `Ligand` object under some field like `tags` or `properties`? This would mean that you don't have to convert to openeye mol every time you read and add a tag, you can also still put all of these tags into the molecule when you do convert to oemol as well by extending the `flush_attrs_to_SD_data` method. It just feels a little strange to have a schema to store the info in but the sdf is still the primary storage format. I don't have a feel for what data goes into these SD tags so I could be missing something but could common ones also be added as fields which can be validated?

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
```diff
@@ -0,0 +1,61 @@
...
+    def from_dict(cls, dict):
+        return cls.parse_obj(dict)
+
+    @classmethod
+    def from_json(cls, json_str):
+        return cls.parse_obj(json.loads(json_str))
+
+    # check_fields required to be false as `data` not defined for baseclass
+    @validator("data", check_fields=False)
+    def data_must_not_be_empty(cls, v):
```
jthorton, line 47: Pydantic has this validation built in, if you change any required fields to a default value of `...` (ellipsis) pydantic will raise a 'ValidationError' if you create a model with a missing required field.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,223 @@
...
+
+# Ligand Schema
+
+
+class LigandIdentifiers(DataModelAbstractBase):
+    """
+    Identifiers for a Ligand
+    """
+
+    moonshot_compound_id: Optional[str] = Field(
```
jthorton, line 42: Is there some structure to these ids? if so it might be good to add some validation in the form of regex to make sure invalid ids don't sneak through.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def to_SD_tags(self) -> dict[str, str]:
+        """
+        Convert to a dictionary of SD tags
+        """
+        data = self.dict()
+        return {str(k): str(v) for k, v in data.items() if v is not None}
+
+
+class Ligand(DataModelAbstractBase):
```
jthorton, line 63: Would it be useful to have some creation provenance here as well? We could store data depending on the class method used to make the molecule for example the function name and the input data?  Something like:
```python
# from smiles
provenance = Provenance(method="from_smiles", input="CCC", version="asap.__version__", date=datetime)
# from sdf
provenance = Provenance(method="from_sdf", input="path/to/file", version="asap.__version__", date=datetime)

```

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: Yeah 10/10 agree with this one! I wonder if we even want to strip the `data` representation of SD Tags to help alleviate the problem you mention below regarding ligands with different SD tags being considered unequal. Then we when we write out the SDF file we can add the SD Tags? idk good thing to argue about on Wednesday :P

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: I'm sure @kaminow can cook some nice regex up for us!

We used to use REGEX to parse all the IDs out of the file names so this definitely seems possible!

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
apayne97: is there a way to make a method required in the base class but changeable in the subclasses? I think the problem here is that `data` is defined in the subclasses but not the base class. if we make it an abstract method and make it required in the base class will that requirement hold if we overwrite the method in subclasses?

hmacdope: Yep no way to make it required but you just have base class raise NotImplementedError and force the child classes to implement it. 

I'll make the changes and push tommorow morn.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: ah not really because it isn't functional, don't think it adds anything moving to baseclass as would need to re-define to get correct description.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: This has been implemented now but will follow up in discussion with @jthorton below.

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
hmacdope: I have just removed this as @jthorton suggests.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: I am going to push provenance to another PR, issue #343

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
```diff
@@ -0,0 +1,61 @@
...
+        if not v:
+            raise ValueError("Data field can not be empty")
+        return v
+
+    @property
+    def size(self) -> ByteSize:
+        """Size of the resulting JSON object for this class"""
+        return ByteSize(utf8len(self.json())).human_readable()
+
+    def data_equal(self, other: DataModelAbstractBase) -> bool:
```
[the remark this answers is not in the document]
hmacdope: We can just drop this if we don't think useful?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Raised an issue #344

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Basically this comes down to the possibly tricky idea that we are trying to have an SDF file as a first class object here, ie can reconstruct the whole object from one with the correct tags. I think we probably need to make a clear choice one way or another. 

IMO:

**Pros**
-  The chemists work with SDFs not JSON, so if we give them something it can come back with all the info it left with
- No need to carry around other metadata when working with SDFs as will re-serialise to the same object.
- No need to worry about attributes becoming out of sync with underlying sdf str as the underlying data **is** the attributes.  

**Cons**
- A little bit clumsy with custom serialisation logic. 
- Painful roundtripping

**Cons of dropping SDF as serialisable format** 
- Once you write an object to an SDF it will not re-serialize to the same object without also injecting exactly the same metadata. 


If we only want to work with JSON thats fine we can drop attributes as tags all together and things will be much simpler, but the flipside is that once you write out an SDF it is no longer round-trippable. 

I am leaning on the side of keeping SDFs as first class, but we can discuss on Wednesday. 

For reference here is what an `Ligand` full of attributes serializes to as an SDF. 

```

  -OEChem-06282316122D

  6  5  0     0  0  0  0  0  0999 V2000
    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.5000    0.8660    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.5000    0.8660    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    3.0000    1.7321    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
    4.0000    1.7321    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  2  3  1  0  0  0  0
  3  4  1  0  0  0  0
  4  5  1  0  0  0  0
  5  6  1  0  0  0  0
M  END
> <compound_name>
'MyLig'

> <ids>
{'moonshot_compound_id': '1234', 'manifold_api_id': 'd12508af-afb8-44df-b084-8b536e3f11e2', 'manifold_vc_id': 'BLAH', 'compchem_id': '8a6dc4d2-45ba-48b6-841a-fde29eef3fbc'}

> <experimental_data>
{'compound_id': 'blah2', 'racemic': 'False', 'achiral': 'False', 'absolute_stereochemistry_enantiomerically_pure': 'False', 'relative_stereochemistry_enantiomerically_pure': 'False'}

> <experimental_data_values>
{'pIC50': 1.0}

> <name_jeff>
'42'

$$$$

```

hmacdope: Lets discuss in full on Wed.

hmacdope: As a partial improvement I have added the `tags` attribute that you can assign to and will be written as SD tags.

hmacdope: @JenkeScheen any thoughts? We will probably discuss tommorow.

JenkeScheen: I think we should focus on making the object round-trippable at all times. Let's discuss in more detail on Thursday.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
kaminow: if the `Ligand` object will only ever have an SDF file in the `data` object, would it make sense/is it possible to just make the `data_format` a constant rather than a `Field`?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,240 @@
...
+        compound_name : Optional[str], optional
+            Name of the compound, by default None
+        read_SD_attrs : bool, optional
+            Whether to read in SD tags as attributes, by default True, overrides kwargs
+        """
+        # directly read in data
+        sdf_str = read_file_directly(sdf_file)
+        lig = cls(data=sdf_str, compound_name=compound_name, **kwargs)
+        if read_SD_attrs:
+            lig.pop_attrs_from_SD_data()
```
kaminow, line 159: this method doesn't actually pop anything, right? I could definitely be wrong about this, but in my mind "pop" means "remove from source and return"

kaminow (changes requested):
this is looking great, really happy with the progress! barring our discussion soon and the couple of minor requests/suggestions, I think it's good to go

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,240 @@
...
+            Whether to write out attributes as SD tags, by default True
+
+        """
+        if write_SD_attrs:
+            self.flush_attrs_to_SD_data()
+        # directly write out data
+        write_file_directly(filename, self.data)
+
+    def set_SD_data(self, data: dict[str, str]) -> None:
+        self.tags.update(data)
```
hmacdope, line 182: @apayne97 Try json.parse to have embedded subdicts

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,240 @@
...
+            Path to the SDF file
+        write_SD_attrs : bool, optional
+            Whether to write out attributes as SD tags, by default True
+
+        """
+        if write_SD_attrs:
+            self.flush_attrs_to_SD_data()
+        # directly write out data
+        write_file_directly(filename, self.data)
+
```
hmacdope, line 180: Clear SD tags on read, the underlying .data should never have tags.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: we decided to not move this, instead we'll have a validator check to make sure that at least compound_name or one of the ids is set

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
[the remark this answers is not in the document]
apayne97: we can have `==` mean just compare positions (just the string in `data`)

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Done!

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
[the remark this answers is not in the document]
hmacdope: Done

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -0,0 +1,223 @@
...
+        data = self.dict()
+        return {str(k): str(v) for k, v in data.items() if v is not None}
+
+
+class Ligand(DataModelAbstractBase):
+    """
+    Schema for a Ligand
+    """
+
+    compound_name: str = Field(None, description="Name of compound")
```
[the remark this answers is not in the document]
hmacdope: See comment about validator for one of `name` or a value in `LigandIds`

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: The field should be const now with `allow_mutation=False`, `const=True`

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Done.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Does now as discussed :)

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: > @apayne97 Try json.parse to have embedded subdicts

I am going to push this to another PR, this is good enough for now.

hmacdope: Issue raised #354

hmacdope:
@apayne97 @kaminow  @JenkeScheen @dotsdl @jthorton this is probably ready for a final review. 

I know I have pushed a few things to separate issues, but I think we should try and move on here. What we have should work for now and we can refine as problems arise.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -62,7 +72,31 @@ def to_SD_tags(self) -> dict[str, str]:
 
 class Ligand(DataModelAbstractBase):
     """
-    Schema for a Ligand
+    Schema for a Ligand.
+
+    Has first class serialization support for SDF files as well as the typical JSON and dictionary
+    serialization.
+
+    Note that equality comparisons are done on the SDF data in the underlying .data field, not the other fields.
```
apayne97, line 80:
```suggestion
    Note that equality comparisons are done on the chemical structure data found in the `data` field, not the other fields or the SD Tags in the original SDF
```

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -62,7 +72,31 @@ def to_SD_tags(self) -> dict[str, str]:
...
+    This means you can change the other fields and still have equality, but changing the SDF data will change
```
apayne97, line 81:
```suggestion
    This means you can change the other fields and still have equality, but changing the chemical structure data will change
```

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: Not sure if this is better but I think "SDF data" is a bit too vague?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
apayne97: Not sure if this is better but I think "SDF data" is a bit too vague?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -62,7 +72,31 @@ def to_SD_tags(self) -> dict[str, str]:
...
+    compound_name : str, optional
+        Name of compound, by default None
+    ids : Optional[LigandIdentifiers], optional
+        LigandIdentifiers Schema for identifiers associated with this ligand, by default None
+    experimental_data : Optional[ExperimentalCompoundData], optional
+        ExperimentalCompoundData Schema for experimental data associated with the compound, by default None
+    tags : dict[str, str], optional
+        Dictionary of SD tags, by default {}
+    data : str, optional, private
+        SDF file stored as a string to hold internal data state, by default ""
```
apayne97, line 97:
```suggestion
        Chemical structure data from the SDF file stored as a string ""
```

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -159,36 +252,65 @@ def to_sdf(self, filename: Union[str, Path], write_SD_attrs: bool = True) -> Non
...
 
     def set_SD_data(self, data: dict[str, str]) -> None:
-        mol = sdf_string_to_oemol(self.data)
-        mol = set_SD_data(mol, data)
-        self.data = oemol_to_sdf_string(mol)
+        """
+        Set the SD data for the ligand, uses an update to overwrite existing data in line with
+        OpenEye behaviour
+        """
+        self.tags.update(data)
```
apayne97, line 270: what do we want to happen if you try to update the SD tags with a reserved field?

asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py
```diff
@@ -57,5 +50,36 @@ def size(self) -> ByteSize:
...
+    def __ne__(self, other: DataModelAbstractBase) -> bool:
+        return not self.data_equal(other)
+
     class Config:
         validate_assignment = True
+        # can't use extra="forbid" because of the way we use
+        # kwargs to skip root_validator on some fields
+
+
+def schema_dict_get_val_overload(obj: dict | BaseModel):
```
apayne97, line 67: you C++ m*****f***** with your "overloaded" functions. 
love it.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema_v2.py
```diff
@@ -141,41 +182,50 @@ def test_ligand_sdf_rountrip_data_only(
...
+
+@pytest.mark.parametrize("tags", [{"test_key": "test_value"}, {}])
 @pytest.mark.parametrize("exp_data_vals", [{"pIC50": 5.0}, {}])
 @pytest.mark.parametrize("moonshot_compound_id", ["test_moonshot_compound_id", None])
 @pytest.mark.parametrize("manifold_vc_id", ["ASAP-VC-1234", None])
 @pytest.mark.parametrize("manifold_api_id", [uuid4(), None])
 @pytest.mark.parametrize("compchem_id", [uuid4(), None])
-@pytest.mark.parametrize("compound_name", ["test_name", None])
+@pytest.mark.parametrize("compound_name", ["test_name"])
 def test_ligand_sdf_rountrip_SD(
```
apayne97, line 229: can we also add a test to make sure we can check if all of the data of two ligands is the same / different? I think using `flush_attrs_to_SD_data`

apayne97 (changes requested):
Mostly edits to documentation / testing! the only comment is on updating the SD tags - should we add a check to make sure someone isn't trying to update SD tags that are a restricted attribute? Phenomenal job with this! Excited to see this come together!

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
hmacdope: Good spotting, added a check for that and a test.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema_v2.py
hmacdope: I have added a `full_equal` method to check for every-field equality.

hmacdope:
@apayne97 hopefully have addressed all your comments.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema_v2.py
apayne97: nice, thanks!

apayne97 (approved):
LGTM!

PR #339, opened 2023-06-26
Add Schema for Targets

hmacdope:
Adds a schema to ship a protein structure around data around, is basically a thin wrapper around a PDB  that is designed to be manipulated as an OEMol and then re-constructed.

Branches off #338 which should be merged first.


hmacdope:
This is ready for review but we should probably finish up on #338 first.

asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -522,3 +522,289 @@ def openeye_copy_pdb_data(
...
+    -------
+    str
+        PDB string representation of the input OEDesignUnit
+    """
+    oms = oechem.oemolostream()
+    oms.SetFormat(oechem.OEFormat_PDB)
+    oms.openstring()
+    oechem.OEWriteDesignUnit(oms, du)
+    molstring = oms.GetString().decode("UTF-8")
+    return molstring
```
apayne97, line 810: What gets written out here? a string representation of the binary OEDU file?

Or just a PDB representation of the protein-ligand system that is in the OEDU?

Can you reconstruct a DU using this string? I don't see any tests yet

asapdiscovery-data/asapdiscovery/data/schema_v2/dynamic_properties.py
```diff
@@ -0,0 +1,7 @@
+import pkg_resources
+
+from ..dynamic_enum import make_dynamic_enum
+
+target_type_yaml = pkg_resources.resource_filename(__name__, "target_type.yaml")
+
+TargetType = make_dynamic_enum(target_type_yaml, "TargetType")
```
apayne97, line 7: this is a cool setup and I am going to steal it in future projects :P

asapdiscovery-data/asapdiscovery/data/openeye.py
apayne97: We have other methods for getting PDB files from DesignUnits and I thought they required a few more hoops to jump through.

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    Schema for a Target
+    """
+
+    target_name: str = Field(None, description="The name of the target")
+
+    ids: Optional[TargetIdentifiers] = Field(
+        None,
+        description="TargetIdentifiers Schema for identifiers associated with this ligand",
+    )
```
apayne97, line 49: same comment for the Ligand schema, I think we might want to enforce the existence of at least 1 ID.

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -0,0 +1,110 @@
...
+    data: str = Field(
+        "",
+        description="PDB file stored as a string to hold internal data state",
+        repr=False,
+    )
+    data_format: DataStorageType = Field(
+        DataStorageType.pdb,
+        description="Enum describing the data storage method",
+        allow_mutation=False,
+    )
```
apayne97, line 60: Same comment for ligand schema, do we want to make this a part of the base class?

Or I guess if we keep it manually consistent then that works I guess

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -0,0 +1,110 @@
...
+    def from_pdb_via_openeye(
+        cls, pdb_file: Union[str, Path], target_name: Optional[str] = None, **kwargs
+    ) -> "Target":
+        # directly read in data
+        pdb_str = read_file_directly(pdb_file)
+        # NOTE: tradeof between speed and consistency with `from_pdb` method lines below will make sure that the pdb string is
+        # consistent between a load and dump by roundtripping through and openeye mol but will slow down the process significantly.
+        mol = pdb_string_to_oemol(pdb_str)
+        pdb_str = oemol_to_pdb_string(mol)
+        return cls(data=pdb_str, target_name=target_name, **kwargs)
```
apayne97, line 80: what is openeye doing that makes this method different? ie what fields are lost

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Therefore the user is responsible for reading and outputting an OEMol or OEGraphMol of the OEDesignUnit
+    component in question.
+
+    eg.
+
+    L = Ligand.from_pdb('complex.pdb')
+    prepped_oemol = prep_oemol(L.to_oemol())
+    L2 = Ligand.from_oemol(prepped_oemol)
+    """
```
apayne97, line 110: I think we would like to be able to pass a DU around along with this schema, whatever that takes. As mentioned its pretty nontrivial to go DU --> Mol --> DU so if we can pass a DU directly (even if that means some godawful binary string) I think that would be preferabl

asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
```diff
@@ -0,0 +1,104 @@
...
+    assert t1 == t2
+
+
+def test_oemol_roundtrip_via_openeye(
+    moonshot_pdb,
+):  # test that a pdb file can be read in and out consistently via roundtrip through openeye
+    t1 = Target.from_pdb_via_openeye(moonshot_pdb)
+    mol = t1.to_oemol()
+    t2 = Target.from_oemol(mol)
+    assert t1 == t2
```
apayne97, line 104: can you add a test that shows that this is necessary ? 
i.e. an expected failure due to differences between reading / loading with / without openeye?

apayne97 (changes requested):
More questions and thoughts! Overall love where this is going and exited to see the progress!

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
hmacdope: I'm going to investigate DUs seperately as part of #355

asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
hmacdope: Should no longer be nessecary. :)

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
hmacdope: I think I have fixed this now  by changing the `OEIFlavour` to `OEIFlavour_PDB_DATA` but the only caveat is the `MASTER ` record sometimes ends up in a different place. Hence I have excluded it from the `__eq__` comparison.

hmacdope:
@apayne97 this should be ready for another look over I think. I will push investigating DUs to an additional PR see #355

hmacdope:
@JenkeScheen @apayne97 @kaminow follow up question, do we want to be able to re-serialise from a PDB file analogous to SDFs in #338? I would say possibly less important, but we can discuss.

hmacdope:
@apayne97 any further feedback here?

kaminow (approved):
I think this looks great! very excited with the progress being made

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -0,0 +1,129 @@
...
+                ):
+                    raise ValueError(
+                        "At least one identifier must be provide, or target_name must be provided"
+                    )
+        return v
+
+    @classmethod
+    def from_pdb(
+        cls, pdb_file: Union[str, Path], target_name: Optional[str] = None, **kwargs
+    ) -> "Target":
```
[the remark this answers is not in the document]
hmacdope: Delayed annotation of class type hints is still `future` and doesnt play nice with python 3.9. 

See this PEP https://peps.python.org/pep-0563/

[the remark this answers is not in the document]
kaminow: oh interesting. learn something new every day

PR #363, opened 2023-07-06
fix gif viz kinks

JenkeScheen:
Makes the gif viz so that the view is *just* the provided orientation. We were depending on zooming on ligand, this removes that function. Also hides atoms that are outside of frame to slightly speed up pymol rendering. 

Also added slight smoothing across frames, not sure why this was disabled.


JenkeScheen:
@hmacdope can you review and merge? I don't think I'll be checking this during my holiday.

hmacdope (approved):
LGTM!

PR #378, opened 2023-07-18
PreppedTarget DesignUnit Schema

### LINKED ISSUE #355: Investigate just embedding DesignUnit in schema using Bytes, MsgPack or Bson
We have been trying to get around storing a DesignUnit in a Schema by moving it to another format.

Following some insightful discussions with @jchodera and @dotsdl we may be able to store the DU directly with 

1. `BytesIO`
2. `BSON`
3. `MsgPack`

We should investigate as this would simplify our life considerably.

hmacdope:
Fixes #355 

Adds a schema to embed OEDUs as b64 encoded bytes in a PreppedTarget. I can't believe this actually works.


asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
```diff
@@ -24,15 +25,10 @@ def sars2_spruced_pdb():
...
-def test_targettype_init(ttype):
-    tt = TargetType(ttype)
-    assert tt.value == ttype
-
-
-def test_targettype_init_bad_name():
-    with pytest.raises(ValueError):
-        _ = TargetType("bad_name")
+@pytest.fixture(scope="session")
+def oedu():
```
kaminow, line 29: might be good to change this to `oedu_fn` or something so that it's clear this is a file and not an actual `OEDesignUnit` object

asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
```diff
@@ -108,10 +110,61 @@ def test_oemol_roundtrip(
...
+
+
+def test_preppedtarget_from_oedu_file_at_least_one_id(oedu):
+    with pytest.raises(ValidationError):
+        # neither id is set
+        PreppedTarget.from_oedu_file(oedu)
+
+
+def test_preppedtarget_from_oedu_file_at_least_one_target_id(oedu):
+    _ = PreppedTarget.from_oedu_file(oedu, ids=TargetIdentifiers())
```
kaminow, line 138: don't we want this to raise an error?

kaminow (changes requested):
a couple of mostly minor things, but this is looking good! one thing that I don't think is addressed (but I'm not sure if it's something that we've reached a decision on) is how we're going to serialize a `PreppedTarget` object (ie #355). will this serialize properly if we just do `pt.json()`?

asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
hmacdope: good idea.

asapdiscovery-data/asapdiscovery/data/tests/test_target_schema_v2.py
hmacdope: Yes! good pickup

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -127,3 +135,63 @@ def __eq__(self, other: Any) -> bool:
...
+
+    @root_validator(pre=True)
+    @classmethod
+    def _validate_at_least_one_id(cls, v):
+        # simpler as we never need to pop attrs off the serialised representation.
+        ids = v.get("ids")
+        compound_name = v.get("target_name")
+        # check if all the identifiers are None
+        if compound_name is None:
+            if ids is None or all([v is None for v in ids]):
```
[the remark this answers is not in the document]
hmacdope: Good spotting :)

hmacdope:
> a couple of mostly minor things, but this is looking good! one thing that I don't think is addressed (but I'm not sure if it's something that we've reached a decision on) is how we're going to serialize a `PreppedTarget` object (ie #355). will this serialize properly if we just do `pt.json()`?

Yes the JSON serialisation will work. I have added a test to show that.

hmacdope:
@kaminow thanks for the review! Hopefully have addressed everything.

hmacdope:
@kaminow thanks for the fixes, hopefully good to go.

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -1,10 +1,18 @@
...
+    oedu_to_bytes64,
+    oemol_to_pdb_string,
+    pdb_string_to_oemol,
+    save_openeye_design_unit,
+)
 from pydantic import Field, root_validator
 
-from .dynamic_properties import TargetType
+from ..postera.manifold_data_validation import TargetTags
 from .schema_base import (
```
apayne97, line 16: is there a good reason to use relative imports here? don't we want to stick to `asapdiscovery.data...` as much as possible?

apayne97 (changes requested):
This is so cool! Glad to see you got the bytes round trip encoding working. I think I'll need to use the code to have any more specific feedback but the tests convince me that it must do at least most of what it says it does! Just a minor nit

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
hmacdope: Yep sure, can fix.

hmacdope: Done!

Issue #379, opened 2023-07-21
HTML renders need new NGLView version


hmacdope:
HTML docked poses renders are not currently working 

Change needed  on ngl load to `<script src="https://unpkg.com/ngl@2.1.1/dist/ngl.js" type="text/javascript"></script>`

hmacdope:
Seems to be Google Chrome specific. Try firefox for older artifacts.

Issue #371, opened 2023-07-13
Make Viz-targets enum follow the new same convention as new postera tags with proper virus names.


hmacdope:
Make Viz-targets enum follow the new same convention as new postera tags with proper target and virus tag names 

current 

```python
from enum import Enum


# enum for allowed targets
# TODO make configurable from YAML file
class VizTargets(Enum):
    viz_sars2_mpro = "sars2_mpro"
    viz_mers_mpro = "mers_mpro"
    viz_7ene_mpro = "7ene_mpro"
    viz_272_mpro = "272_mpro"
    viz_sars2_mac1 = "sars2_mac1"
...
```

Desired: Targets and viruses should follow the new assay names convention with proper virus taxonomy names, eg:
```
SARS-CoV-2-Mac1
# special views should be a postfix 
**SARS-CoV-2-Mac1_myspecialview**
```

PR #380, opened 2023-07-21
Fix GIF and HTML angles + rework VizTargets + spruce up dataviz modules. 

### LINKED ISSUE #376: GIF poses seem to be broken.
GIF poses seem to be broken at the moment post #348  I have added back in

```python
         # p.cmd.zoom("resn UNK", buffer=4)  # zoom to ligand
```

In #375 

For now, but this means the pose may not be the exact same angle. We will work towards fixing this again, not sure exactly what went wrong with #348

hmacdope:
Fixes #376 #371 #379 + others

- Changes the target enums for visualisation to work better
-  fixes up some angles and renders for GIFS + HTML 
- re-organises some of the internal logic of the Viz classes and modules to be neater and more consistent. 
- Adds standalone script for HTML render testing 
- Adds info on how to make HTML angles.


asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
```diff
@@ -155,17 +144,12 @@ def get_html_footer(self):
...
-            return colour_mers_mpro + orient_tail_272_mpro
-        elif self.target == "sars2_mac1":
-            return colour_sars2_mac1 + orient_tail_sars2_mac1
-        else:
-            raise ValueError(
-                f"Target {self.target} does not have an HTML visualiser element implemented."
-            )
+        # colour uses target name
+        protein_name = VizTargets.get_target_name(self.target, underscore=True)
+        colour = getattr(HTMLBlockData, f"colour_{protein_name}")
```
jthorton, line 149: two functions on the `HTMLBlockData` might work better here one something like `get_pocket_colour` and `get_orient_tail` which just takes the targets name and handles all of the underscore matchings internally might be easier to follow and allow you to reuse a single matching logic.

asapdiscovery-dataviz/asapdiscovery/dataviz/tests/test_gif_viz.py
```diff
@@ -0,0 +1,47 @@
...
+    "logger",
+    [
+        None,
+        FileLogger(
+            "gif_to_viz", path="./", stdout=True, level=logging.DEBUG
+        ).getLogger(),
+    ],
+)
+def test_html_viz(traj, top, logger, target, tmp_path):
+    gif_visualiser = GIFVisualizer(
```
jthorton, line 35: It might be worth considering a factory pattern for the GIF and HTML maker that way you can configure the general settings and just pass in individual targets to some run method to actually do the work?

hmacdope: This is a good idea, but I think i will push to another PR if thats ok?

hmacdope: Just because the fixes to visualisation angles are needed urgently.

PR #384, opened 2023-07-24
Change cudatoolkit pin for lilac

### LINKED ISSUE #342: Pin cudatoolkit to <=11.7 for lilac
Pin cudatoolkit to <=11.7 for lilac. Lilac does not have the correct driver versions for cudatoolkit=11.8 (the default.)

Lets pin in the main env or at least make another env file with a working cudatoolkit version.

hmacdope:
Fixes #342

Using cudatoolkit > 11.7 will cause lilac nodes to complain about PTX versioning due to using old drivers. Lets just pin for simplicity.


hmacdope:
@mikemhenry is there an easy work around for no cudatoolkit on MacOS?

mikemhenry:
The issue is that because there is no cudatoolkit on MacOS, you need to not have that explicitly listed as a dependency and instead let the solver pull in what it needs.

I think this highlights an issue, is the env yaml for CI or for someone to create an env for production? We really shouldn't be doing both with the same yaml IMHO for this reason

PR #396, opened 2023-07-28
Put adaptive back in to deal with failing workers and align correctly to traj. 

hmacdope:
Workers are not spinning back up, best way to fix this is with scheduled migrations and adaptive scaling. 

## Status
- [ ] Ready to go


hmacdope:
@jthorton can I get a quick review here. :)

PR #400, opened 2023-08-01
Use pydantic settings and pydantic >=2.0

### LINKED ISSUE #388: Consider pydantic settings managment
We could consider making a pydantic settings class to cover all of the settings [here](https://github.com/choderalab/asapdiscovery/blob/a4bda414a8920ca161e303399e2ae4ce3026db82/asapdiscovery-docking/asapdiscovery/docking/scripts/dock_small_scale_end_to_end.py#L390-L422) as I am sure other work flows may need to reuse these like uploading the results of FEC? Pydantic can handle gathering the settings from the environment automatically see [here](https://docs.pydantic.dev/2.0/usage/pydantic_settings/#settings-management). We can then add validation methods which checks they are all set.

hmacdope:
Fixes #388 

Adds in pydantic settings to easily make interfaces to services.


hmacdope:
@jthorton I think this is blocked upstream by some pydantic pins in openFF, let me investigate fully.

hmacdope:
@jthorton do you have some hack for getting pydantic-settings with pydantic <=2.0.0

PR #404, opened 2023-08-03
Add stereo enumerator script

### LINKED ISSUE #403: Add stereo enumerator script.
Add capacity to enumerate stereo easily and quickly.

hmacdope:
Fixes #403


hmacdope:
This should become a workflow class, but lets start here and build it out when we need it next.

hmacdope:
@jthorton this is now a workflow class that we can convert to a workflow element down the track.

JenkeScheen:
what are your ideas on how we should be naming the enumerated molecules?

hmacdope:
> what are your ideas on how we should be naming the enumerated molecules? `warts` are okay-ish but can be problematic downstream, especially if you add additional enumerations that `wart` (e.g. enumeration of protonation states).

Let me think about this and do some research, as it must be a common problem. Perhaps @jthorton has some ideas also.

PR #413, opened 2023-08-07
Dask jobqueue Cluster factories to make clusters with correct configurations

### LINKED ISSUE #399: Move dask configurations to a factory function that takes a schema
Currently we are relying on a loose coupling between a yaml file that defines the dask-jobqueue spec, we should have a function that just makes you the right jobqeue with arguments at runtime.

hmacdope:
Fixes #399


hmacdope:
Confirmed that this works on lilac.

hmacdope:
@dotsdl @jthorton could I grab a review here :)

asapdiscovery-data/asapdiscovery/data/dask_utils.py
```diff
@@ -0,0 +1,111 @@
...
+
+class DaskCluster(BaseModel):
+    class Config:
+        allow_mutation = False
+        extra = "forbid"
+
+    name: str = Field("dask-worker", description="Name of the dask worker")
+    cores: int = Field(1, description="Number of cores per job")
+    memory: str = Field("20 GB", description="Amount of memory per job")
+    death_timeout: int = Field(
```
jthorton, line 52: Can we add the units to the description? I guess its seconds?

asapdiscovery-data/asapdiscovery/data/dask_utils.py
```diff
@@ -0,0 +1,111 @@
...
+}
+
+_LILAC_GPU_EXTRAS = {
+    GPU.GTX1080TI: [
+        '-R "select[hname!=lt16]"'
+    ],  # random node that has something wrong with its GPU
+}
+
+
+def _walltime_to_h(walltime: str) -> int:
```
jthorton, line 37: use this as a validator, that way you can have the walltime in the correct format on the schema might be easier to track that way?

asapdiscovery-data/asapdiscovery/data/dask_utils.py
```diff
@@ -0,0 +1,111 @@
...
+    shebang: str = Field("#!/usr/bin/env bash", description="Shebang for the job")
+    queue: str = Field("cpuqueue", description="LSF queue to submit jobs to")
+    project: str = Field(None, description="LSF project to submit jobs to")
+    walltime: str = Field("1:00", description="Walltime for the job")
+    use_stdin: bool = Field(True, description="Whether to use stdin for job submission")
+    job_extra_directives: Optional[list[str]] = Field(
+        None, description="Extra directives to pass to LSF"
+    )
+
+    def to_cluster(self, exclude_interface: Optional[str] = "lo"):
```
jthorton, line 67: add the return type

jthorton (approved):
LGTM! This will be a nice way of tracking the dask settings used once everything is in schema format. I have a couple of small changes not blocking though!

asapdiscovery-data/asapdiscovery/data/dask_utils.py
hmacdope: fixed by removing mismatch between `24:00` and `24h` syntax altogether

hmacdope: Just using `h` notation from now on.

hmacdope:
I just need to do some final testing on lilac itself to make sure this is all good.

hmacdope:
Confirmed that this works on lilac.

PR #406, opened 2023-08-03
Better loss tracking

kaminow:
Change the stats that are tracked logged during training to include the compound and label associated with loss values, rather than just the loss values themselves. This will make it easier to figure out what's going on in training and isolate any specific molecules that might be difficult for the model.

Also some other minor changes that coincided with this, mainly a rearrangement of the code in `train.py` to be smarter about `wandb` initialization and letting it play better with CLI args.

PR #419, opened 2023-08-07
Improve loss tracking

kaminow:
Redo of #406, should have all suggestions there addressed.

Change the stats that are tracked logged during training to include the compound and label associated with loss values, rather than just the loss values themselves. This will make it easier to figure out what's going on in training and isolate any specific molecules that might be difficult for the model.

Also some other minor changes that coincided with this, mainly a rearrangement of the code in train.py to be smarter about wandb initialization and letting it play better with CLI args.


hmacdope:
Just cycling CI here.

PR #422, opened 2023-08-08
Fix for GIF viz so that pocket is coloured correctly

### LINKED ISSUE #393: GIF colouring seems to be off.
The colouring of MAC1 GIFs doesn't seem to be working as indened.
trajectory

hmacdope:
Fixes #393 

Adds in `keepIds=True` so the right residue numbers are written to PDB files.


hmacdope:
@JenkeScheen this should be a quick review.

JenkeScheen (approved):
LGTM, I don't see why this would break stuff for MERS/Sars2 Mpro but we'd notice soon enough.

PR #429, opened 2023-08-10
Add FragalysisFactory schema

### LINKED ISSUE #383: FragalysisFactory for generation of prep input Target and Ligand Schema from Fragalysis
FragalysisFactory for generation of prep input Target and Ligand Schema from a fragalysis file system
This will looks something like the below in use.
```python
from asapdiscovery.data.schema_v2 import *

targets: List[Target], ligs: List[Ligand] = FragalysisFactory(fragalysis_path) 
```
OR we can also combine these into a combined schema, which may be neater.

```python
from asapdiscovery.data.schema_v2 import *

complexs: List[Complex] = FragalysisFactory(fragalysis_path)

### PR DESCRIPTION
**Pull request by kaminow**
## Description
Add a `FragalysisFactory` schema to contain all complexes in a Fragalysis database. Also adds a `Complex` schema which will be stored in `FragalysisFactory`.

Closes #383.

## Todos
  - [x] Add `Complex` schema
  - [x] Test `Complex` schema
  - [x] Add `FragalysisFactory` schema
  - [x] Test `FragalysisFactory` schema

## Status
- [x] Ready to go

### CODE REVIEW

**Comment by kaminow** [comment]
@hmacdope can you take a look through this when you get a chance? wondering what you think, and also what else would be useful to have here

File: asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+        if len(df) == 0:
+            raise ValueError("metadata.csv file is empty.")
+
+        if (xtal_col not in df.columns) or (compound_col not in df.columns):
+            raise ValueError(
+                "metadata.csv file must contain a crystal name column and a "
+                "compound name column."
+            )
+
+        # Dict mapping crystal name to compound name
```
hmacdope, line 71: I might try and see if we can use more pandas ops here and below.

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
```diff
@@ -0,0 +1,117 @@
...
+                continue
+
+            try:
+                c = Complex.from_pdb(
+                    pdb_fn,
+                    target_kwargs={"target_name": d},
+                    ligand_kwargs={"compound_name": compound_name},
+                )
+            except Exception:
+                print(f"Unable to parse PDB file for {d}.", flush=True)
```
hmacdope, line 101: Should be an error, and could also have a `fail_missing` switch as to whether to skip or raise.

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
```diff
@@ -0,0 +1,117 @@
...
+                continue
+
+            complexes.append(c)
+
+        return cls(parent_dir=parent_dir, complexes=complexes)
+
+    @root_validator(pre=True)
```
hmacdope, line 108: This can just be a normal validator as it only takes a single field argument. Root validators are normally used to compare *between* arguments in the whole class input. 

Something like 

```python
    @validator("parent_dir")
    @classmethod
    def _validate_parent_dir(cls, v): #v is a single value when using `validator("xxx")`
        if not v.exists():
            raise ValueError("Given parent_dir does not exist.")

        if not v.is_dir():
            raise ValueError("Given parent_dir is not a directory.")

        return v
```

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
```diff
@@ -0,0 +1,117 @@
...
+        complexes = []
+        for d in all_xtal_dirs:
+            compound_name = xtal_compound_dict[d]
+            pdb_fn = parent_dir / "aligned" / d / f"{d}_bound.pdb"
+            if not pdb_fn.exists():
+                print(f"No PDB file found for {d}.", flush=True)
+                continue
+
+            try:
+                c = Complex.from_pdb(
```
hmacdope, line 95: Awesome, this is going to be so so useful.

asapdiscovery-data/asapdiscovery/data/tests/test_complex_schema_v2.py
```diff
@@ -0,0 +1,84 @@
...
+        Complex.from_pdb(complex_pdb, target_kwargs={"target_name": "test"})
+
+    with pytest.raises(ValidationError):
+        Complex.from_pdb(complex_pdb, ligand_kwargs={"compound_name": "test"})
+
+    with pytest.raises(ValidationError):
+        Complex.from_pdb(complex_pdb)
+
+
+def test_complex_dict_roundtrip(complex_pdb):
```
hmacdope, line 52: Great!

asapdiscovery-data/asapdiscovery/data/tests/test_complex_schema_v2.py
```diff
@@ -0,0 +1,84 @@
...
+        complex_pdb,
+        target_kwargs={"target_name": "test"},
+        ligand_kwargs={"compound_name": "test"},
+    )
+    c2 = Complex.from_dict(c1.dict())
+
+    assert c1 == c2
+
+
+def test_complex_json_roundtrip(complex_pdb):
```
hmacdope, line 63: Great again!

asapdiscovery-modeling/asapdiscovery/modeling/modeling.py
```diff
@@ -577,19 +577,24 @@ def split_openeye_mol(
         Molecule filter object that contains the filter criteria.
     prot_cutoff_len : int, default=10
         Minimum number of residues in a protein chain required in order to keep
+    keep_one_lig : bool, default=True
+        Only keep one copy of the ligand in lig_mol
 
     Returns
     -------
-    namedtuple
```
hmacdope, line 583: Was the named tuple not working for the new design?

asapdiscovery-modeling/asapdiscovery/modeling/modeling.py
```diff
@@ -577,19 +577,24 @@ def split_openeye_mol(
...
     # These are the objects that will be returned at the end
     lig_mol = oechem.OEGraphMol()
     prot_mol = oechem.OEGraphMol()
     water_mol = oechem.OEGraphMol()
     oth_mol = oechem.OEGraphMol()
 
-    if type(molecule_filter) == str:
+    if molecule_filter is None:
+        molecule_filter = MoleculeFilter()
+    elif type(molecule_filter) == str:
```
hmacdope, line 597:
```suggestion
    elif type(molecule_filter) is str:
```
or  can use `isinstance` , happy with either, personally I think type(x) is ok for elementary types like str. 

https://www.flake8rules.com/rules/E721.html

asapdiscovery-modeling/asapdiscovery/modeling/modeling.py
```diff
@@ -577,19 +577,24 @@ def split_openeye_mol(
...
     Returns
     -------
-    namedtuple
-        A namedtuple containing the protein, ligand, and water molecules.
+    dict
+        A dict containing the protein ("prot"), ligand ("lig"), water atoms ("wat"),
+        and all other atoms ("oth")
     """
     # These are the objects that will be returned at the end
     lig_mol = oechem.OEGraphMol()
```
hmacdope, line 590: These are all 2D right? perhaps at some point we will need to preserve the 3D information. Lets think about how to address this best. Perhaps its as simple as changing to `OEMol`

asapdiscovery-modeling/asapdiscovery/modeling/tests/test_modeling_utils.py
```diff
@@ -15,74 +13,88 @@
...
+def test_ligand_splitting(target, local_path, ligand_chain, oemol_dict):
+    """
+    Test splitting when we just care about ligand.
+    """
+
+    oemol = oemol_dict[target]
+    molfilter = MoleculeFilter(ligand_chain=ligand_chain)
+    split_dict = split_openeye_mol(oemol, molfilter)
+
+    print(type(split_dict["lig"]), flush=True)
```
hmacdope, line 47:
```suggestion
```

hmacdope (changes requested):
Looking great @kaminow, amazing progress. Check out comments. Something we will need to address is preserving 3D coords from `split-openeye-mol`

asapdiscovery-modeling/asapdiscovery/modeling/modeling.py
kaminow: the `OEGraphMol` class does have 3D coordinates. it's just a specific `OEBaseMol` subclass that's used for single-conformation molecules. all information for each atom should be preserved from the original complex molecule to the split molecules

asapdiscovery-modeling/asapdiscovery/modeling/modeling.py
kaminow: previously only the protein molecule was getting returned, which is why I ended up reworking the function. I think `namedtuple` should be fine, I've just never used them before so I defaulted to `dict` lol. can switch it around if you think it would be better

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
kaminow: oh cool, didn't know that was an option lol

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
kaminow: good call

asapdiscovery-data/asapdiscovery/data/schema_v2/complex.py
```diff
@@ -0,0 +1,61 @@
...
+        ligand_kwargs={},
+    ):
+        # First load full complex molecule
+        complex_mol = load_openeye_pdb(pdb_file)
+
+        # Split molecule into parts using given chains
+        mol_filter = MoleculeFilter(
+            protein_chains=target_chains, ligand_chain=ligand_chain
+        )
+        split_dict = split_openeye_mol(complex_mol, mol_filter)
```
hmacdope, line 45: Add similar code path to `Target.from_pdb` to force there to be no ligands in a regular target as well.

hmacdope (commented):
Write Fragalysis factory tests, other than that looks great!

asapdiscovery-data/asapdiscovery/data/schema_v2/fragalysis.py
kaminow: this ended up being a great idea for helping to make sure that Factories loaded from the same directory have the complexes in the same order (previously relied on the `os.listdir`, which I think is not the same every time it's called)

kaminow:
@hmacdope thanks for the suggestions! I think everything should be ready for a second review whenever you get a chance

hmacdope (approved):
Amazing work @kaminow! Thanks so much for hitting this, is a big step on the way to getting schema into everything! Super pleased with this.

PR #442, opened 2023-08-21
Make targets build from PDB strictly protein only.

### LINKED ISSUE #434: Openeye mol splitting code path to `Target.from_pdb` to force there to be no ligands in a regular target as well.
Add similar code path to `Target.from_pdb` to force there to be no ligands in a regular target as well.

_Originally posted by @hmacdope in https://github.com/choderalab/asapdiscovery/pull/429#discussion_r1296496875_

hmacdope:
Fixes #434

Target schema built directly from a PDB should exclude the ligand, instead this is handled by a `Ligand` and combined in a Complex

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
+++ b/asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
@@ -4,12 +4,16 @@ from typing import Any, Dict, Optional, Tuple, Union  # noqa: F401
 from asapdiscovery.data.openeye import (
     bytes64_to_oedu,
     load_openeye_design_unit,
+    load_openeye_pdb,
     oechem,
     oedu_to_bytes64,
     oemol_to_pdb_string,
     pdb_string_to_oemol,
     save_openeye_design_unit,
 )
+from asapdiscovery.modeling.modeling import split_openeye_mol
+from asapdiscovery.modeling.schema import MoleculeFilter
+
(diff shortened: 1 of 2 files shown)


asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -91,34 +93,30 @@ def _validate_at_least_one_id(cls, v):
...
 
     def to_pdb(self, filename: Union[str, Path]) -> None:
         # directly write out data
         write_file_directly(filename, self.data)
 
     @classmethod
-    def from_oemol(
-        cls, mol: oechem.OEMol, target_name: Optional[str] = None, **kwargs
-    ) -> "Target":
+    def from_oemol(cls, mol: oechem.OEMol, **kwargs) -> "Target":
```
kaminow, line 116: is there a reason that we're getting rid of the `target_name` arg other than that it's not really necessary to have since it can just get lumped in `kwargs`?

kaminow (approved):
looks great! definitely an important improvement

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
hmacdope: Yeah its messing with dumping via kwargs as can sometimes get duplicate kwarg as you say.

PR #443, opened 2023-08-23
Refactor interactive pose viewer backend to 3Dmol

JenkeScheen:
We're switching to a `3DMol` backend instead of `NGL`. Beside having a more '`PyMol`-like' atom/bonds representation and controls, this package is _much more flexible_ and will allow us to:
- show interactions in a modular way 
- show bond orders from SDF

Overall the code should become somewhat cleaner/more interpretable. 

Main points to implement:
- [x] use `airium` to write HTML rather than just slapping strings together
- [x] add view orientation for all targets (possibly remove 272)
- [x] feed in color_dict for both `subpockets` and `fitness` coloring methods (so remove bfactor workaround)
- [x] add interactions by default. may need to use `PLIP` for this because `ProLIF` seems to sensitive to protonation variations
- [x] finalize bond orders and protonation representation for both ligand and protein
- [x] make sure all components work in the workflow
- [x] remove all redundant code 

Preview of new view:
image


dotsdl:
@JenkeScheen is this ready for review, or is there more to hit on your checklist first?

JenkeScheen:
Thanks, this is still WIP. I'd like to get the bypass in place, then get a review in before I attempt to remove the current flow.

dotsdl:
Sounds good! Ping me when you'd like a review!

chrisiacovella:
Have you considered using templates for the html generation part?   I am a big fan of [jinja](https://jinja.palletsprojects.com/en/3.1.x/ ), since, in addition to just straight up find/replace, it allows you to define formatting macros and loops within the templates, minimizing the amount of processing/preparation you need to do inside the python code.

JenkeScheen:
noting down here so I don't forget to update the conda env.yml: we're adding a dependency to `PLIP`, of which the `openbabel` dependency is a bit fiddly. This ended up doing the trick:
```
conda install -c conda-forge openbabel lxml 
conda install -c conda-forge --no-deps plip
```

JenkeScheen:
`PLIP` conda install has `abi` issues on running `my_mol.analyze()`:
```
libc++abi: terminating with uncaught exception of type swig::stop_iteration
Abort trap: 6
```
for now, going with `singularity` -> load `PLIP` docker image approach.

JenkeScheen:
This is almost ready for review, I just need to resolve some issues with protonation. I'll check in with @kaminow for this, looks like it's mostly just `OE` being difficult. 

@dotsdl, I'm mostly unsure about https://github.com/choderalab/asapdiscovery/pull/443/commits/73cfad4187c1b5c9368ca88b64d1230098369a1d. It works on my end, but I'm not sure that `conda`-installing `plip` will result in a `plip` bin for all users, if not then https://github.com/choderalab/asapdiscovery/commit/ecdf23db1443406e48a22a6be44bf45363a0ec4f?diff=unified#diff-8ad06bf1f92be6cc4e927ac68d370da58af3abfef7c2fa0e9c4ceac4a12c8756R278 would fail. Can you check that after install, that CLI call works correctly?

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
```diff
@@ -128,36 +137,326 @@ def write_html(html, path):
...
+                    + "].includes(atom.resi)){ \n return '"
+                    + color
+                    + "' \n "
+                )
+
+        # start writing the HTML doc.
+        a("<!DOCTYPE HTML>")
+        with a.html(lang="en"):
+            with a.head():
+                a.meta(charset="utf-8")
```
dotsdl, line 355: Think I agree with @chrisiacovella that using `jinja` plus a template may be less painful than building up the script below. I think you could also use a combination if you prefer, using `airium` to build up the overall HTML doc but using `jinja` to populate the JS script.

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
```diff
@@ -128,36 +137,326 @@ def write_html(html, path):
...
+
+        # load the XML produced by PLIP that contains all the interaction data.
+        intn_dict_xml = xmltodict.parse(
+            ET.tostring(ET.parse("tmp/report.xml").getroot())
+        )
+
+        intn_dict = {}
+        intn_counter = 0
+        # wrangle all interactions into a dict that can be read directly by 3DMol.
+        for _, data in intn_dict_xml["report"]["bindingsite"]["interactions"].items():
```
dotsdl, line 294: Tests currently choke on this: looks like `intn_dict_xml["report"]["bindingsite"]` is a `list`, with more than one item. Not sure what the correct thing to do here is; do we expect more than one binding site in many cases?

dotsdl (changes requested):
Looking improved over previous version @JenkeScheen! A few changes needed before this is good to merge:

- CI currently failing. Looks like you'll need to remove some of the test items since this PR removes things like `view_coords_MERS_CoV_Mpro_7ene`. Also, there is a failure in testing `get_interactions_plip`: it looks like there may be more than one bindingsite in many cases?

In a future PR, we'll probably want to:
- use `jinja` templates for at least the JS components of the generated HTML artifacts; this is likely to be simpler than building up the strings hardcoded into the module itself.
- move target specs in `viz_targets`, and `view_coords`, `pocket_dict`s, and `color_dict`s in `_gif_blocks`, and `orient`s in `_html_blocks` to a set of YAML files instead of directly in the module code. We can take advantage of similar patterns @hmacdope has done elsewhere in `asapdiscovery`

dotsdl:
Running the tests also leaves a bunch of log files in the working directory; can we disable these?

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
JenkeScheen: I agree. I'll discuss with @hmacdope and @mikemhenry before starting a PR to introduce `jinja` throughout the codebase.

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
JenkeScheen: ah, that might be happening if we have two ligands bound to a dimer (in case of e..g Mpro). In principle that's fine, we can show interactions on both monomers. I'll change this to account for that.

JenkeScheen:
> CI currently failing. Looks like you'll need to remove some of the test items since this PR removes things like view_coords_MERS_CoV_Mpro_7ene

ah, I'll have a look. I haven't been paying attention to CI as it was breaking because of an issue with `fragalysis`, thanks for spotting.

dotsdl:
@JenkeScheen looks like CI is broken in a different spot now in `get_interactions_plip`. Can you address this?

JenkeScheen:
ugh

JenkeScheen:
@dotsdl looks like CI still fails after my bugfixes, but on something to do with `fragalysis`.

JenkeScheen:
looks like CI is still failing stochastically on `fragalysis` stuff. `python 3.10` succeeded this time, so this PR is good to go. @dotsdl did you want to have one last look?

hmacdope:
> looks like CI is still failing stochastically on `fragalysis` stuff. `python 3.10` succeeded this time, so this PR is good to go. @dotsdl did you want to have one last look?

Yeah sorry about this known issue #463

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
```diff
@@ -278,20 +279,34 @@ def get_interactions_plip(self, pose) -> dict:
...
         # wrangle all interactions into a dict that can be read directly by 3DMol.
-        for _, data in intn_dict_xml["report"]["bindingsite"]["interactions"].items():
+        binding_site_data = intn_dict_xml["report"]["bindingsite"]
+        if isinstance(binding_site_data, list):
+            # this can happen if multiple binding sites (e.g. dimer) exist
+            interaction_data = {}
+            [
+                interaction_data.update(binding_site["interactions"])
+                for binding_site in binding_site_data
+            ]
```
dotsdl, line 304: Is the intention here to just overwrite the dict with the values of the last `'interactions'` dict in `binding_site`? Is this guaranteed to always be what we want to do?

dotsdl (approved):
Still seeing lots of `log.txt` files polluting the working directory when running the tests. Will raise an issue for this; library code generally shouldn't be creating side-effect files in the working directory by default.

Otherwise, good to merge once comments addressed.

asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
JenkeScheen: keys here are unique, we're enumerating interactions when we get them from PLIP. so in this way, no interaction will be overwritten at any point.

PR #447, opened 2023-08-24
Add postera factory to make ligands. 

### LINKED ISSUE #446: Make `PosteraFactory` to make ligand schema from postera `MoleculeSets`
Make `PosteraFactory` to make ligand schema from postera `MoleculeSets`

hmacdope:
Fixes #446

Makes a factory that can create Ligand Schema from postera molecule sets. Queried with the Postera people how the VCIDs appear in the API if at all.


kaminow (approved):
don't know as much about the PostEra API, but this seems good! might be good to add some tests, but I'm assuming everything else is tested already

PR #449, opened 2023-08-25
Add handling for Cheng-Prusoff values in mtenn.PIC50Readout class

kaminow:
As of mtenn v0.3.0, the `PIC50Readout` class takes values for the Cheng-Prusoff equation. Update our code to interface properly with that.


asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -250,6 +258,8 @@ def build_model(
             config = {}
     elif (type(config) is str) or isinstance(config, Path):
         config = parse_config(config)
+    elif type(config) is not dict:
```
hmacdope, line 261: Long term may be an idea to move these configs to .... you guessed it, schema. But no need in this PR.

asapdiscovery-ml/asapdiscovery/ml/utils.py
```diff
@@ -1722,9 +1745,15 @@ def update_loss_dict(split, compound_id, target, in_range, uncertainty, pred, lo
 
         # Stop if loss has gone to infinity or is NaN
         if (
-            np.isnan(epoch_val_loss)
+            np.isnan(epoch_train_loss)
```
hmacdope, line 1748: Move to a function called `check_valid` or something like that this is a few too many `or`.

hmacdope (changes requested):
Looking good, see comments. :)

asapdiscovery-ml/asapdiscovery/ml/utils.py
kaminow: yep that's a good idea. opened #450 to track this

hmacdope (approved):
Looks good to me @kaminow :)

PR #451, opened 2023-08-30
Clean up environment files and split CI into linux and MacOS components.

### LINKED ISSUE #342: Pin cudatoolkit to <=11.7 for lilac

### LINKED ISSUE #432: Clean up older conda envs
There are a bunch of old conda envs in the `devtools` directory that we can get rid of.

hmacdope:
Fixes #432 
Fixes #342


.github/workflows/CI.yaml
```diff
@@ -45,14 +45,23 @@ jobs:
         df -h
         ulimit -a
 
+    - name: Determine Conda env file
```
mikemhenry, line 48: This could be simplified by making the file names the same as the `matrix.os` variable like: `environment-file: devtools/conda-envs/asapdiscovery-${{ matrix.os }}.yaml`

mikemhenry (approved):
Looks good, I think there is a way to make it a bit more simple but what you have looks fine!

.github/workflows/CI.yaml
hmacdope: good idea, change made.

PR #461, opened 2023-09-05
Add schematised docking workflow class 

### LINKED ISSUE #452: Make schematised workflow class for actual docking run.
Make schematised workflow class for actual docking run, can link with work done in #430 and share a base class with that.

hmacdope:
Fixes #452


asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,272 @@
...
+            for inp in inputs:
+                out = dask.delayed(self._dock)(inputs=[inp])
+                outputs.append(out[0])  # flatten
+        else:
+            outputs = self._dock(inputs=inputs)
+        return outputs
+
+    @abc.abstractmethod
+    def provenance(self) -> dict[str, str]:
+        ...
```
hmacdope, line 78: Add Score Type enum

hmacdope:
@ijpulidos @apayne97 @kaminow @jthorton would yall be able to review for content whil I try and debug failing CI runner.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+    NONE = oedocking.OEPoseRelaxMode_NONE
+
+
+class POSITDocker(DockingBase):
+    """
+    Docker class for POSIT.
+    """
+
+    type: Literal["POSITDocker"] = "POSITDocker"
+    score_type: Literal[SCORE_TYPE.CHEMGAUSS4] = SCORE_TYPE.CHEMGAUSS4
```
jthorton, line 114: Would it make sense to separate the scoring and the pose generation? To make the workflow as modular as possible this would allow you to have a new pose generation method like docking program x and still get the CHEMGAUSS4 score with minimal changes as the workflow would take the final geometries from docking and apply a list of score functions to them?

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+
+    relax: POSIT_RELAX_MODE = Field(
```
jthorton, line 116: Similar to the score does it make sense to separate the relaxation method, I am not familiar with posit so maybe these are very specific to this method and this makes sense here but more generally it might be nice to have other relaxation methods like they hybrid ML/MM stuff we have in FEGrow.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+        POSIT_RELAX_MODE.NONE,
+        description="When to check for relaxation either, 'clash', 'all', 'none'",
+    )
+    posit_method: POSIT_METHOD = Field(
+        POSIT_METHOD.ALL, description="POSIT method to use"
+    )
+    use_omega: bool = Field(True, description="Use omega to generate conformers")
+    num_poses: PositiveInt = Field(1, description="Number of poses to generate")
+    log_name: str = Field("run_docking_oe", description="Name of the log file")
```
jthorton, line 125: We could probably drop this and just absorb the log file into the result object in a field like stdout?

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+    )
+    allow_low_posit_prob: bool = Field(False, description="Allow low posit probability")
+    low_posit_prob_thresh: float = Field(
+        0.1,
+        description="Minimum posit probability threshold if allow_low_posit_prob is True",
+    )
+    allow_final_clash: bool = Field(
+        False, description="Allow clashing poses in last stage of docking"
+    )
+    output_dir: Path = Field(
```
jthorton, line 137: I like to keep the work logic and file writing separate so these classes would just return the result object and then the workflow this fits into would handle file writing where needed.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+    and a Ligand object containing the docked pose.
+    Also contains the probability and chemgauss4 score of the docked pose.
+    """
+
+    input_pair: DockingInputPair = Field(description="Input pair")
+    posed_ligand: Ligand = Field(description="Posed ligand")
+    probability: Optional[PositiveFloat] = Field(description="Probability")
+    score_type: SCORE_TYPE = Field(description="Docking score type")
+    score: float = Field(description="Docking score")
+    provenance: dict[str, str] = Field(description="Provenance")
```
jthorton, line 38: We could have a settings field where just dict the docking method used to preserve the settings as well.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+    Base class for docking.
+    """
+
+    type: Literal["DockingBase"] = "DockingBase"
+
+    @abc.abstractmethod
+    def _dock() -> list[DockingResult]:
+        ...
+
+    def dock(
```
jthorton, line 70: Awesome love this!

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+class DockingResult(BaseModel):
+    """
+    Schema for a DockingResult, containing both a DockingInputPair used as input to the workflow
+    and a Ligand object containing the docked pose.
+    Also contains the probability and chemgauss4 score of the docked pose.
+    """
+
+    input_pair: DockingInputPair = Field(description="Input pair")
```
jthorton, line 33: Just want to check I understand is the input pair the receptor and a Ligand object and then the posed ligand is the ligand again but with a different geometry?

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: Yep sounds good

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: The input `pair` is a combination of:

- A query `Ligand` to dock to the receptor.
- A `PreppedComplex` that has the receptor + bound ligand (if present) as an `oedu` AND a `Ligand` that represents the compound originally bound to the receptor (if present). This allows us to do matching against the receptor for chemoinf easily. 

Yes the `posed_ligand` is then the same query ligand again in a different conformation.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
kaminow: I believe the docking score is added to the mol object by OpenEye during docking, but it's also possible to rescore molecules after docking: https://docs.eyesopen.com/toolkits/python/dockingtk/OEDockingClasses/OEScore.html

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+    posed_ligand: Ligand = Field(description="Posed ligand")
+    probability: Optional[PositiveFloat] = Field(description="Probability")
+    score_type: SCORE_TYPE = Field(description="Docking score type")
+    score: float = Field(description="Docking score")
```
kaminow, line 37: I think some way of having multiple scores would be nice, maybe making these lists and having them correspond 1:1? or some kind of enum that has a member variable (not sure if this is actually possible)

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
kaminow, line 114: following on with my comment in the `DockingResult` class, this could be a list of the scores we want to calculate

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -0,0 +1,283 @@
...
+
+                        combined_oemol = docking_result.to_posed_oemol()
+                        save_openeye_pdb(combined_oemol, output_pdb_file)
+
+            else:
+                pass
+
+        return docking_results
+
+    def provenance(self) -> dict[str, str]:
```
kaminow, line 278: do we want to also include the docking parameters/method used? or is that something for the future?

kaminow (changes requested):
so far looking really good! I left some comments but I assume we'll also probably discuss this a bit in the dev sync tomorrow

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: Seperated into a scorer, but it deosn't seem easy to rescore for `probability` so that will have to come out at the time the docking is done.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: Lets push adding relaxation blocks to the to another PR, as adding relaxation as a seprrate step here isn't really possible as baked into OpenEye

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: Moved it to a staticmethod, so now needs to be done in workflow but is contained withing class still.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: issue raised #527

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: Future, I will work on all the provenance stuff down the track.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: I am going to go back through and sort out all the provenance.

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: There is now a `MetaScorer` class to do this in this PR. See `scorers_v2`

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
hmacdope: I think this is solved in the stuff i just pushed. Just need to figure out the right way to melt the resulting dataframe.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -369,3 +372,11 @@ def pop_attrs_from_SD_data(self) -> None:
 
 class ReferenceLigand(Ligand):
     target_name: Optional[str] = None
+
+
+def compound_names_unique(ligands: list[Ligand]) -> bool:
```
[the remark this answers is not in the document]
hmacdope: I just grouped it with the other ligand associated stuff, its fine here for now, we can move if we change our minds.

hmacdope:
@ijpulidos CI seems to be working again even when i tried to torture it (see `1ed7dff`) perhaps issues were related to https://www.githubstatus.com/incidents/tpykx175y6jt. I will revert that change, perhaps a self hosted runner is not needed right now. Sorry for the red herring, we may need to revisit the idea at some point though, lets just see how it goes.

hmacdope:
@jthorton @kaminow would you mind giving this another look.

asapdiscovery-docking/asapdiscovery/docking/scorer_v2.py
```diff
@@ -0,0 +1,237 @@
...
+        results = []
+        for inp in inputs:
+            schnet_score = self.inference_cls.predict_from_oemol(inp.to_posed_oemol())
+            results.append(
+                Score.from_score_and_docking_result(schnet_score, self.score_type, inp)
+            )
+        return results
+
+
+class MetaScorer(ScorerBase):
```
jthorton, line 222: This is awesome!

kaminow (commented):
couple more minor things, but this is great! I really like the addition of the `Scorer` classes

asapdiscovery-docking/asapdiscovery/docking/scorer_v2.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Score from a combination of other scorers
+    """
+
+    scorers: list[ScorerBase] = Field(..., description="Scorers to score with")
+
+    def _score(self, inputs: list[DockingResult]) -> list[Score]:
```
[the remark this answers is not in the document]
hmacdope: good point, will re-work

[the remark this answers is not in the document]
hmacdope: Done

kaminow (approved):
other than whatever pre-commit/CI stuff comes up, I think this looks good to go!

PR #474, opened 2023-09-15
Pull errors and tracebacks from errored tasks from alchemiscale

ijpulidos:
## Description
Add feature to extract errors and tracebacks from simulations run with alchemiscale.

The idea is to extend the `status` command to query the errored tasks in the network and return their key with the error (python exception) and the traceback, if specified in the CLI.

## Todos
  - [ ] Extend the `status` command with options or flags to get errors and tracebacks
  - [ ] Add the API point for this purpose in `AlchemiscaleHelper` class
  - [ ] Write tests

## Status
- [ ] Ready to go


ijpulidos:
Closing. Superseded by https://github.com/choderalab/asapdiscovery/pull/477

PR #475, opened 2023-09-16
Added `restart` CLI point to `asap-alchemy`

### LINKED ISSUE #468: restarting (automatically) failed simulations

dotsdl:
## Description
Closes #468

## Status
- [ ] Ready to go

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
@@ -278,5 +278,37 @@ def status(network: str):
     client.network_status(planned_network=planned_network)
 
 
+@cli.command()
+@click.option(
+    "-n",
+    "--network",
+    type=click.Path(resolve_path=True, readable=True, file_okay=True, dir_okay=False),
+    help="The name of the JSON file containing a planned FEC network.",
+    default="planned_network.json",
+    show_default=True,
+)
+@click.option(
+    "-v",
+    "--verbose",
+    count=True,
+    help="Verbosity of output; one `-v` will give ScopedKeys of restarted Tasks",
+)
+def restart(network: str, verbose):
+    """Restart all errored Tasks for the given FEC network.
+
+    """
+    from .schema.fec import FreeEnergyCalculationNetwork
+    from .utils import AlchemiscaleHelper
+
+    client = AlchemiscaleHelper()
+    planned_network = FreeEnergyCalculationNetwork.from_file(network)
+
+    restarted_tasks = client.restart_tasks(planned_network)
+    if verbose == 1:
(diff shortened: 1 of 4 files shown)


JenkeScheen:
should there be a way to store and query which tasks have been retried and subsequently failed/succeeded? This would be useful for users that want to see whether their previous `restart` has resulted in any more `success` runs. Seeing that the same number of simulations failed after `restart` would prompt the user to look for a systemic issue, rather than just `restart`ing again.

dotsdl:
@JenkeScheen I think the information a user would need to decide whether to `restart` again or not will be obtained from the approach implemented in #477. That should give for each currently 'error'ed `Task` the number of failed `ProtocolDAGResult`s that exist (corresponding to the number of retries), and for each of those the exception(s)/traceback(s) raised.

dotsdl:
We could make this command take individual `Task` `ScopedKey`s as input if a user wants to do targeted restarts for just a subset. Would this be useful?

ijpulidos:
> We could make this command take individual `Task` `ScopedKey`s as input if a user wants to do targeted restarts for just a subset. Would this be useful?

I think that could be useful, since as far as I can see some errored tasks can be restarted right away (some NaN'ed sims for example), whereas others might need better protein preparation or some manual intervention before attempting the restart (for example with template amino acids not being matched). From #477 it should be easy to extract the `ScopedKey`s for errored `Task`s.

JenkeScheen:
Yes! And in a future PR we could potentially add a way for users to prioritize `restart`ing sure-shot errors.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
```diff
@@ -278,5 +278,35 @@ def status(network: str):
...
+    type=click.Path(resolve_path=True, readable=True, file_okay=True, dir_okay=False),
+    help="The name of the JSON file containing a planned FEC network.",
+    default="planned_network.json",
+    show_default=True,
+)
+@click.option(
+    "-v",
+    "--verbose",
+    count=True,
+    help="Verbosity of output; one `-v` will give ScopedKeys of restarted Tasks",
```
JenkeScheen, line 294: would multiple `v`s do anything here? If not, should we add multiple levels?

dotsdl: Realized that if we're only using one count, kinda silly to make this count-based. Switched it to a simple flag.

dotsdl:
> Yes! And in a future PR we could potentially add a way for users to prioritize `restart`ing sure-shot errors.

Added ability to specify as arguments `ScopedKey`s of `Task`s to limit restarts to.

JenkeScheen:
just tested this as well, works like a treat too!

Issue #465, opened 2023-09-07
Consider a separator between `{self.dataset_name}` and `{system_a.name}` in FEC 


hmacdope:
Consider a separator between `{self.dataset_name}` and `{system_a.name}`.

_Originally posted by @dotsdl in https://github.com/choderalab/asapdiscovery/pull/424#discussion_r1300848297_

PR #476, opened 2023-09-16
Add check during creation of FEC dataset that there are no duplicate ligands

### LINKED ISSUE #469: add check during setup that all ligands are unique, i.e. make sure there are no duplicates in the series

dotsdl:
## Description
Closes #469 and #465.

## Status
- [x] Ready to go

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
@@ -1,5 +1,6 @@
 from typing import Literal, Optional
 
+from collections import Counter
 import gufe
 import openfe
 from alchemiscale import ScopedKey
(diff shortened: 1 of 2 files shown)


dotsdl:
Needs an explicit test yet.

JenkeScheen:
would it be possible to signal to the user *which* ligands are duplicates? Would make it easier to debug the input.

dotsdl:
@JenkeScheen done! This is now ready for review!

JenkeScheen (approved):
LGTM! The only drawback to this approach is if we were to model two binding poses for one ligand. Currently we're not doing that (and we're not planning to in the near future). We can revisit/extend this functionality when we hit that.

dotsdl:
@JenkeScheen I believe for same ligand, different binding poses the `SmallMoleculeComponent`s will hash differently, since they use the conformer information as part of this calculation. So we might be good to go there.

dotsdl:
Test failures appear unrelated; merging!

PR #477, opened 2023-09-18
Pull errors and tracebacks from errored tasks from alchemiscale

ijpulidos:
## Description

Add feature to extract errors and tracebacks from simulations ran with alchemiscale.

The idea is to extend the status command to query the errored tasks in the network and return their key with the error (python exception) and the traceback, if specified in the CLI.

## Questions
- [x] What do we want the output to be?

## Status
- [x] Ready to go


ijpulidos:
@JenkeScheen For now this is just printing the errors or full tracebacks. When we call something like `asap-alchemy -n network.json --errors` (only errors) or `asap-alchemy -n network.json --errors --with-traceback` to get the full tracebacks as well.

Is there a preference on how to get them? Maybe we serialize them to a json file or something similar. I'd appreciate your thoughts on this. Thanks!

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
```diff
@@ -175,3 +176,30 @@ def collect_results(
...
+        network_key = planned_network.results.network_key
+        errored_tasks = self._client.get_network_tasks(network_key, status="error")
+
+        error_data = defaultdict(dict)
+        for task in errored_tasks:
+            for err_result in self._client.get_task_failures(task):
+                for failure in err_result.protocol_unit_failures:
+                    error_data[str(task.gufe_key)]["errors"] = failure.exception
+                    if with_traceback:
+                        error_data[str(task.gufe_key)]["traceback"] = failure.traceback
```
dotsdl, line 204: These lines as written will only retain the last `protocol_unit_failure` exception/traceback for the last failed `ProtocolDAGResult` returned by `AlchemiscaleClient.get_task_failures`. This isn't necessarily complete information, since each `Task` can have many failed `ProtocolDAGResult`s (one for each attempt), and each of these could feature many failed units (eventually...with the way we currently run on `alchemiscale`, this isn't possible).

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
```diff
@@ -259,23 +266,44 @@ def gather(network: str, allow_missing: bool):
...
     # load the network
     planned_network = FreeEnergyCalculationNetwork.from_file(network)
     # check the status
     client.network_status(planned_network=planned_network)
+    # Output errors
+    if errors:
+        task_errors = client.collect_errors(
+            planned_network, with_traceback=with_traceback
+        )
+        print(json.dumps(task_errors, indent=4))
```
dotsdl, line 308: Instead of this, would recommend some human-readable structured output, such as separated, multiline text for each `Task`, the exceptions/tracebacks for each `ProtocolDAGResult`, as the default. 

Can then do `click.echo` of JSON form as you are doing here if  a `--json` flag is given. That would be suitable for redirection into a file from STDOUT.

dotsdl (commented):
Looking good so far @ijpulidos!

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
ijpulidos: Oh yes, you are right, let me think about it since that's definitely the general case we want to support. Good catch, thanks!

ijpulidos: One way is to just add another level to the nested object that has the protocol unit key, since a task can have many different failing units. I don't know if this dictionary is becoming too big and if so, would it be best to have our own custom object to store these results?

JenkeScheen:
> Is there a preference on how to get them? Maybe we serialize them to a json file or something similar. I'd appreciate your thoughts on this. Thanks!

For now let's stick to stdout, let's let @jthorton decide on how he wants to port this once he's back in a couple weeks.

ijpulidos:
@dotsdl @JenkeScheen I think this one should be ready for review. There's probably still room for improvement in terms of the output of the cli command, but I think this can be reiterated once we decide on what the `restart` command should be handling in #475 . But I think this is a good starting point right now.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
```diff
@@ -165,9 +165,8 @@ def submit(network: str, organization: str, campaign: str, project: str):
         project: The name of the project this network should be submitted under.
     """
     from alchemiscale import Scope
-
-    from .schema.fec import FreeEnergyCalculationNetwork
-    from .simulation.utils import AlchemiscaleHelper
```
dotsdl, line 170: Rationale for switching back to absolute imports here? Not exactly opposed, but usually prefer relative imports in library code.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
```diff
@@ -250,6 +249,14 @@ def gather(network: str, allow_missing: bool):
     network_with_results.to_file("result_network.json")
 
 
+def validate_traceback_flag(ctx, param, value):
+    """Validate traceback flag --with-traceback is only used in conjunction with --errors flag."""
+    if value:
+        if not ctx.params.get("errors"):
+            raise click.UsageError("--with-traceback requires --errors flag to be set.")
+    return value
```
dotsdl, line 257: Perhaps instead of forcing users to use both `-et`, `-t` could imply `-e`? In that sense, `-t` is a stronger form of `-e`?

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/conftest.py
```diff
@@ -27,3 +28,16 @@ def tyk2_fec_network():
     """Create a FEC planned network from file"""
     fec_network = fetch_test_file("tyk2_small_network.json")
     return FreeEnergyCalculationNetwork.from_file(fec_network)
+
+
+@pytest.fixture()
```
dotsdl, line 33: Haha looks like we had the same thought. Did something similar in #475; can sort out the dead once we merge them both.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -394,3 +395,34 @@ def create_fec_dataset(
...
+
+    error: tuple[str, tuple[Any, ...]] = Field(
+        tuple(), description="Exception raised and associated message."
+    )
+    traceback: str = Field(
+        "", description="Complete traceback associated with the failure."
+    )
+
+
+class AlchemiscaleFailure(_BaseFailure):
```
dotsdl, line 413: Nice!

dotsdl (approved):
Looks great @ijpulidos! Just a comment or two, but otherwise good to merge in my view!

JenkeScheen:
This is great! I just tested this on a network that had some errors and it worked a treat. Two things:
1. I agree with @dotsdl that the separate `-e` and `-t` are perhaps not the most convenient. Ideally it would be something like `-v` (for just the error) and -`vv` (for error + traceback), but technically we're not using an 'error flag' in that form. Perhaps `-e` and `-ee`? Or is that obscene?
2. Currently there's no (easy) way of finding out the ligand names from the error. For example:
```
Task:
Task-401a06d54cf24d9588771592d359067d-asap-jscheen-mac1_phosphate_run2
Error:
('SimulationNaNError', ('Propagating replica 10 at state 10 resulted in a NaN!\nThe state of the system and integrator before the error were saved in /scratch/dotson/19606012-25/shared/ProtocolDAG-0efe5d85d20aac22ba954a50bd47590b/shared_RelativeHybridTopologyProtocolUnit-a7392625a30142e0899bf2e787e3d0fa_attempt_0/nan-error-logs',))
```
.. would also grabbing the ligand names (and showing them in that box) for the edge be in scope for this PR?

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
ijpulidos: This was just my personal preference, but I don't have a strong opinion on either. I just noticed than in the other subpackages we were using absolute imports for the most part. I think it's okay to mix them to make things more readable.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli.py
ijpulidos: Yeah, I think that makes sense, I'll make this change.

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/conftest.py
ijpulidos: Yeah, I did notice that as well haha. I prefer yours since it's making some assertions that seem useful to have.

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
```diff
@@ -175,3 +176,34 @@ def collect_results(
...
+            planned_network: Network to get failed tasks from.
+
+        Returns:
+            List of AlchemiscaleFailure objects with errors and tracebacks for the failed tasks in the network.
+        """
+        network_key = planned_network.results.network_key
+        errored_tasks = self._client.get_network_tasks(network_key, status="error")
+
+        error_data = []
+        for task in errored_tasks:
```
dotsdl, line 197: > .. would also grabbing the ligand names (and showing them in that box) for the edge be in scope for this PR?

The shortest path to getting ligand names would be, for each `Task` `ScopedKey`, doing `self._client.get_task_transformation(task_sk).name`. With the way [`Transformation`s are built up](https://github.com/choderalab/asapdiscovery/blob/f18e5dfb35328190bc346023c457e0312eebbcd1/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py#L333-L339), this should give sufficient information to track down which ligands are involved.

JenkeScheen:
@ijpulidos is this WIP or can it be merged?

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
ijpulidos: While this is possible, it gets a bit intricate in terms of having to inspect transformations -> ChemicalSystems -> Components and filter the ligands. I think I prefer to leave this one for improvements in a future PR, if we really end up needing this.

ijpulidos:
@JenkeScheen @dotsdl I think this should be ready to be reviewed and merged as it is right now. 

I was thinking about adding the timestamps to the error reports and the json serialization of the errors, but I think I prefer to discuss those in separate issues such that we can give this a test ride as soon as possible and receive the feedback that can help with those future features. If that makes sense. I'll be raising the issues.

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
dotsdl: I don't think it's as intricate as you're making it in this case. Because we build up the `Transformation` names in a particular way, they already include information on the ligands involved in the failed transformation. There is no need to introspect further down into the `ChemicalSystems`, though you could choose to if you prefer since you would have those already.

Something I'd like to do to make this way more performant is to add methods like `get_transformation_name`, `get_chemicalsystem_name`, etc. for getting back the names of these objects without having to pull down a whole chain of `gufe` objects and put them back together. 

Also, perhaps methods `get_chemicalsystem_components` and `get_component_name` would make it quick to get more granular information without full `gufe` objects as well.

JenkeScheen: can tackle this in a next PR!

JenkeScheen:
looking good @ijpulidos , minimal changes since last review. feel free to merge in.

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
ijpulidos: Yes, I agree, I'm unsure how good it is to rely on the transformation names for this. Maybe that's robust enough, but for the sake of moving forward with this PR I think we could discuss it in its own issue.

ijpulidos:
I'm merging this one as a way to have a minimal functionality available as soon as possible. Future improvements/developments are already reflected in some of the issues. Thanks!

Issue #482, opened 2023-09-22
reduce test burden of `test_gif_viz`


JenkeScheen:
`test_gif_viz` currently takes ~30min on my machine.. looks like we're running it on a full-size trajectory, whoops. It's also not named correctly, i.e. it's currently named `test_html_viz`, should be `test_gif_viz`: https://github.com/choderalab/asapdiscovery/blob/main/asapdiscovery-dataviz/asapdiscovery/dataviz/tests/test_gif_viz.py#L34-L47

JenkeScheen:
looks like most of the tests that are slowing down CI are prep-related:
![image](https://github.com/choderalab/asapdiscovery/assets/43140137/47c54942-36c8-4f79-84fa-ddd3d350d71a)

PR #483, opened 2023-09-22
Fix gif test

JenkeScheen:
fixes https://github.com/choderalab/asapdiscovery/issues/482

We were using a single test trajectory + topology, then iterating over all targets. For MERS, this resulted in *very* long test times (up to 5 mins), even though it doesn't test that we're actually generating a GIF. Just testing GIF generation on SARS2 Mpro (which is faster) proves already that we're generating GIFs succesfully.


JenkeScheen:
hmm, looks like we're only winning 3-5 mins. Oh well.

hmacdope (approved):
LGTM @JenkeScheen I will merge when this train of CI fixes is in.

Issue #478, opened 2023-09-20
GIFs are not looping


JenkeScheen:
GIFs are not looping; may be due to increased size (see e.g. https://stackoverflow.com/questions/10867789/animated-gif-only-loops-once-in-chrome-and-firefox). This is not ideal because now users need to refresh the webpage to keep seeing dynamics. 

We should be able to add this functionality back in with `gifsicle`, by changing [the current optimization code](https://github.com/choderalab/asapdiscovery/blob/main/asapdiscovery-dataviz/asapdiscovery/dataviz/gif_viz.py#L321) to:
```
from pygifsicle import gifsicle
gifsicle(
    sources=str(path)
    optimize=True, 
    colors=256,
    options=["--loopcount"]
)
```

PR #484, opened 2023-09-22
swap to using high-level gifsicle api

JenkeScheen:
fixes https://github.com/choderalab/asapdiscovery/issues/478

ez pz, this doesn't break anything.


hmacdope (approved):
Also LGTM @JenkeScheen I will merge when CI fixes are in.

PR #493, opened 2023-10-02
Separate workflows/CI for subpackages

### LINKED ISSUE #265: Separate workflows for the different packages?
As a way of getting the CI tests to be run faster, we could try to have different workflows, that is, one per each package (`data`, `ml`, `docking`, etc... ). This would allow for tests to run in parallel and have separate and smaller environments for each, even though we wouldn't have a single environment file for all the packages, that would be the tradeoff.

### LINKED ISSUE #480: Segmentation fault in spruce protein test in CI
We have been consistently getting a segmentation fault for the CI tests. Namely the [`test_spruce_protein[mers]` test](https://github.com/choderalab/asapdiscovery/actions/runs/6242281217/job/16945951393#step:7:1028).

It is only happening on the Linux CI tests. I cannot reproduce this locally using my linux workstation, all tests are running fine there.

ijpulidos:
## Description
Testing if separating workflows for packages fixes our CI.

In this case it is testing `macos` in a single workflow for all subpackages. This is possible since the "offending" tests were skipped for this OS.

On the other hand, we are separating all subpackages tests for `linux`. It is going to be a bit annoying when we need to update them, but I think it's better to have a single approach (everything separated and "flexible") than having to think what is separated and what isn't. We would need to study if we hit a limit with the amount of workflows or workers that we can run in the project but I think we are fine that way as far as I have tested.

Resolves #265 
Resolves #480

### DIFF
--- a/.github/workflows/alchemy.yaml
+++ b/.github/workflows/alchemy.yaml
@@ -1,0 +1,99 @@
+name: alchemy-ci
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+    branches:
+      - "main"
+  schedule:
+    # Nightly tests run on main by default:
+    #   Scheduled workflows run on the latest commit on the default or base branch.
+    #   (from https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule)
+    - cron: "0 0 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  test:
+    name: Test on ${{ matrix.os }}, Python ${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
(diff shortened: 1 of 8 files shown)


ijpulidos:
Apparently testing docking needs every other asapdiscovery subpackage. I think we probably need to come up with more unitary tests for the docking software.

hmacdope:
Legitimate failure #496 in single-target

ijpulidos:
I guess another advantage of separating the workflows like this is that it should be easier to know where the failure is (which subpackage tests are failing)

hmacdope:
@ijpulidos just while I think of it also need `asapdiscovery-alchemy`

ijpulidos:
@hmacdope Oh definitely! haha, I missed that one. Thanks!

Simulation and docking tests should fail (simulation isn't testing anything and docking has a legitimate fail).

hmacdope (approved):
Seems great! Feel free to merge @ijpulidos.

ijpulidos:
Merging to main, this fixes the segmentation fault issue with the costs of having many separate workflows to maintain now.

We can probably review this in the future if needed, if it becomes too inconvenient to have this amount of granularity.

The macos test that is failing was also failing before these changes, so it's something unrelated to these changes. We probably need to skip that one in macos as well.

PR #502, opened 2023-10-03
Fix alchemy restart testing with non-deterministic order

### LINKED ISSUE #501: testing issue for restarts in asap-alchemy as ordering of returned ScopedKeys is non-deterministic (expected))
A test is failing in `asap-alchemy` as the ordering of ScopedKeys being returned is tested. In general the order in which ScopedKeys is returned is non-deterministic by design.

hmacdope:
Fixes #501


hmacdope:
Tagging @jthorton for your info, but I think I will merge this so we can get through all the small changes held up by CI issues. We can discuss/fix post-hoc?

dotsdl (approved):
Yes, I think this is the right move given how `AlchemiscaleHelper.restart_tasks` is written. It could admittedly be rewritten to make the order more deterministic, but not sure that's really needed here.

Thanks @hmacdope for addressing this!

PR #503, opened 2023-10-03
Remove annoying files and logs

### LINKED ISSUE #481: `asapdiscovery-dataviz` classes produce logfiles in working directory by default
When running the test suite for `asapdiscovery-dataviz`, several `*-log.txt` files are produced, polluting the working directory the test suite is ran from. This appears to stem from default behavior of e.g. `GifVisualizer` and `HTMLVisualizer`, in which not specifying a `logger` results in a `FileLogger` being created instead.

Can we change the default behavior for `logger=None` to be a logger streaming only to STDOUT? If `logger=<Path>` could then be used to add a `FileHandler` pointing to the given `Path`, and `logger=<Logger>` would be the current behavior of just using the given logger?

hmacdope:
Fixes #493
Fixes #481 
Remove annoying logfiles created during tests and DGL cache made upon inference.

### DIFF
--- a/.github/workflows/docking.yaml
+++ b/.github/workflows/docking.yaml
@@ -45,6 +45,10 @@ jobs:
         df -h
         ulimit -a
 
+    - name: Set Swap Space
+      uses: pierotofy/set-swap-space@master
+      with:
+        swap-size-gb: 10
 
     # More info on options: https://github.com/mamba-org/provision-with-micromamba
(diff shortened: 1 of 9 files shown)


hmacdope:
@JenkeScheen @apayne97 @kaminow could I grab a review here, should be simple.

JenkeScheen (approved):
LGTM!

PR #511, opened 2023-10-04
Add fitness interactions

JenkeScheen:
Adjust interaction depiction in the case of a `fitness` HTML viz. Here we're going to color interactions by fitness (of the residue it hits) instead of by interaction type (which can be inferred easily anyway). This will enable the viewer to quickly see if a posed compound is interacting with mutable residues or not. New depiction with interactions colored by fitness:

image

We're adding:
- color interactions the same way as residues (white=no fit mutants, all the way to deep red which = many fit mutants)
- color interactions green if they hit the protein backbone (this is a good tactic and should be encouraged by highlighting it a separate color)

One issue is that light-red and white are difficult to distinguish. This also applies to residue surface coloring, so should be handled in a separate PR. We may want to rethink how we're coloring residues because even if there is a single fit mutant this means the virus *will* mutate to it, so it should be highlighted more clearly compared to straight white.

### DIFF
--- a/asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
+++ b/asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
@@ -254,7 +254,6 @@ class HTMLVisualizer:
         backbone_atoms = [ at.GetIdx() for at in self.protein.GetAtoms(oechem.OEIsBackboneAtom()) ]
 
         # with oe, iterate over atoms until this one's found. then use oechem.OEIsBackboneAtom
-        import sys
         is_backbone = False
         for idx, res_coords in self.protein.GetCoords().items():
             # round to 3 because OE pointlessly extends the coordinates float.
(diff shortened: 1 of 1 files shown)


asapdiscovery-dataviz/asapdiscovery/dataviz/html_viz.py
```diff
@@ -242,13 +242,67 @@ def get_interaction_color(intn_type) -> str:
             return "#8C4099"
         else:
             raise ValueError(f"Interaction type {intn_type} not recognized.")
+        
+    def is_backbone_residue(self, x, y, z) -> bool:
```
JenkeScheen, line 246: this function is gnarly, if you can think of a more elegant way of doing this we should implement it. In any case, currently this is very fast (practically instant) and is catching all interactions.

hmacdope (approved):
LGTM, I will make an issue to enhance gnarly function but lets not block on this.

PR #516, opened 2023-10-05
Use alchemiscale repeats not serial openFE

jthorton:
## Description
This PR updates the default options on the FEC schema to use parallel repeats on alchemiscale and not serial repeats with in openFE which should speed up FEC estimates. cc @JenkeScheen 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


dotsdl (approved):
Looks great @jthorton! Should be good to go with this!

You probably need to wait until the `alchemiscale-client` `conda-forge` build for 0.2.0 completes before using the changes in the env files here.

PR #518, opened 2023-10-06
Remove OE logging in docking

hmacdope:
A re-occurence of https://github.com/openforcefield/openff-toolkit/issues/1615 is interfering in openmm making systems when Omega is being called, under the hood, removing the openeye logging manipulation fixes it. This is despite the initial fix detailed in the issue :(

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/docking.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/docking.py
@@ -118,14 +118,6 @@ def run_docking_oe(
         logger.warning(f"No logfile with name '{logname}' exists, using stdout instead")
     logger.info(f"Running docking for {compound_name}")
 
-    # Set up OEThrow logging
-    # this can sometimes interfere with OEOmega see https://github.com/openforcefield/openff-toolkit/issues/1615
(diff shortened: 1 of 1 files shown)


apayne97:
Can we make this optional? This is only relevant if you need to run simulations, right? and having the logging when just running a bunch of docking is pretty useful

kaminow:
I think that makes sense personally, maybe just a `--verbose` CLI flag or something

apayne97 (dismissed):
(doing the official review thing)
Could you make OE logging an option via a --verbose flag?

hmacdope:
This will segfault if OEOmega is used by **any package** after the docking is run. This includes things like `openmmforcefields` which is triggering this error or but a whole bunch of stuff in our stack uses Omega under the hood. I think possibly triggering a segfault with the verbose flag will be unintuitive (it took some serious digging to even find out this was the issue). For this reason having it pushed to stdout rather than a file I think is preferable, and I would prefer to just remove it.  Sorry to be a pest.

hmacdope:
@kaminow @apayne97 I think I am going to push ahead with this, as is required for #520 which is in turn required to fix some project support issues. We can re-visit this once these are resolved.

PR #528, opened 2023-10-11
Add ability to run inference directly on OEMol

### LINKED ISSUE #253: Allow building of Schnet and e3nn inference tensors directly from OpenEye Mol
I was thinking last night that writing all these temporaries is probably quite inefficient, this is fine for a 0.1.0 but if we could go straight from a OpenEye molecule to the `pose_dict` tensors that would be good.

_Originally posted by @hmacdope in https://github.com/choderalab/asapdiscovery/pull/250#discussion_r1183664688_

kaminow:
Add ability to run inference directly on OEMol for the `StructuralInference` class. Once the new docking stuff is merged we can also add a method to predict from `DockingResult` object if we want. Closes #253

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/docking.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/docking.py
@@ -14,7 +14,6 @@ from asapdiscovery.data.openeye import (
     load_openeye_sdf,
     oechem,
     oedocking,
-    save_openeye_pdb,
     save_openeye_sdf,
 )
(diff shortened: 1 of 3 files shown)


hmacdope (changes requested):
Awesome! Amazing work. Can you also change the call site in the docking code?

kaminow:
oh yep good point, will do

kaminow:
@hmacdope should be good to go once CI finishes

hmacdope (approved):
Thanks @kaminow

PR #531, opened 2023-10-13
Ligand sdf round trip

### LINKED ISSUE #530: Make ligand schema round trip to SDF easily
While developing the Epik state expander I noticed that our internal Ligand model does not round trip to SDF easily. Consider this example case where I want to create a single SDF file with a list of ligands with their attributes such as `compound_name` added as SD tags.

```python
from asapdiscovery.data.openeye import save_openeye_sdfs, load_openeye_sdfs, oemol_to_sdf_string
from asapdiscovery.data.schema_v2.ligand import Ligand

ligands = [Ligand.from_smiles("CC", compound_name="ethane", tags={"parent":"me"}, experimental_data={"compound_id":"1"}), Ligand.from_smiles("CO", compound_name="methanol", ids={"moonshot_compound_id": "randomID"})]

# flush the data
for ligand in ligands:
    # sync the state internal and sdf data
    ligand.data = ligand.flush_attrs_to_SD_data()

oe_mols = [ligand.to_oemol() for ligand in ligands]

# write to sdf
save_openeye_sdfs(oe_mols, "test.sdf")

# load back
loaded_mols = load_openeye_sdfs("test.sdf")
# we need each mol to be an sdf to read with attrs
new_ligands = []
for ligand in loaded_mols:
    with open("ligand.sdf", "w") as out:
        out.write(oemol_to_sdf_string(ligand))
    new_ligands.append(Ligand.from_sdf("ligand.sdf", read_SD_attrs=True))

for ligand in new_ligands:
    print(ligand)
```

What I suggest is that we make the model round trip to and from openeye by default so all tags are included and parsed everytime we convert to and from oemols. I would also only store the graph and coords data in the internal data field and have the sd tags added automatically everytime we convert to an oemol to keep the models in sync as its easy to forget to flush the data before writing to file.

cc @hmacdope

jthorton:
## Description
This PR fixes #530 by allowing our internal Ligand model to round trip to SDF files using SD tags to store extra information. We now store pydantic models as JSON strings in SD tags which even allows for nested models and is general so that any new models added to ligand will automatically survive the round trip. This will allow us to SDFs through the workflow without losing any important information. 

One notable change is that the internal copy of the sdf string we keep in the data field now has no SD tags, these are now added when ever we convert to openeye mols to keep the internal state of the molecule consistent. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [x] Ready to go

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/openeye.py
+++ b/asapdiscovery-data/asapdiscovery/data/openeye.py
@@ -148,6 +148,44 @@ def load_openeye_pdb(
         oechem.OEThrow.Fatal(f"Unable to open {pdb_fn}")
 
 
+def load_openeye_smi(smi_fn: Union[str, Path]) -> oechem.OEGraphMol:
+    """
+    Load an OpenEye SMILES file containing a single molecule and return it as an
+    OpenEye OEGraphMol object.
+    Parameters
+    ----------
+    smi_fn : Union[str, Path]
+        Path to the SMILES file to load.
+    Returns
+    -------
+    oechem.OEGraphMol
+        An OpenEye OEGraphMol object containing the molecule data from the SDF file.
+    Raises
+    ------
+    FileNotFoundError
+        If the specified file does not exist.
+    oechem.OEError
+        If the CIF file cannot be opened.
+    Notes
+    -----
+    This function assumes that the SDF file contains a single molecule. If the
+    file contains more than one molecule, only the first molecule will be loaded.
+    """
+
+    if not Path(smi_fn).exists():
+        raise FileNotFoundError(f"{smi_fn} does not exist!")
+
+    ifs = oechem.oemolistream()
+    ifs.SetFlavor(oechem.OEFormat_SMI, oechem.OEIFlavor_SMI_DEFAULT)
+    if ifs.open(str(smi_fn)):
+        coords_mol = oechem.OEGraphMol()
+        oechem.OEReadMolecule(ifs, coords_mol)
+        ifs.close()
+        return coords_mol
+    else:
+        oechem.OEThrow.Fatal(f"Unable to open {smi_fn}")
+
+
 def load_openeye_cif1(cif1_fn: Union[str, Path]) -> oechem.OEGraphMol:
     """
     Loads a biological assembly file into an OEGraphMol object.
@@ -658,9 +696,9 @@ def oemol_to_smiles(mol: oechem.OEMol) -> str:
     return oechem.OEMolToSmiles(mol)
 
 
-def oemol_to_inchi(mol: oechem.OEMol) -> str:
+def oe_smiles_roundtrip(smiles: str) -> str:
     """
-    InChI string of an OpenEye OEMol
+    Canonical SMILES string of an OpenEye OEMol
 
     Paramers
     --------
@@ -670,12 +708,40 @@ def oemol_to_inchi(mol: oechem.OEMol) -> str:
     Returns
     -------
     str
+       SMILES string of molecule
+    """
+    mol = smiles_to_oemol(smiles)
+    return oemol_to_smiles(mol)
+
+
+def oemol_to_inchi(mol: oechem.OEMol, fixed_hydrogens: bool = False) -> str:
+    """
+    InChI string of an OpenEye OEMol
+
+    Paramers
+    --------
+    mol: oechem.OEMol
+        OpenEye OEMol
+    fixed_hydrogens: bool
+        If a fixed hydrogen layer should be added to the InChI, if `True` this will result in a non-standard inchi
+        which can distinguish tautomers.
+
+    Returns
+    -------
+    str
        InChI string of molecule
     """
-    return oechem.OECreateInChI(mol)
+    if fixed_hydrogens:
+        inchi_opts = oechem.OEInChIOptions()
+        inchi_opts.SetFixedHLayer(True)
+        inchi = oechem.OEMolToInChI(mol)
+    else:
+        inchi = oechem.OEMolToSTDInChI(mol)
+
+    return inchi
 
 
-def oemol_to_inchikey(mol: oechem.OEMol) -> str:
+def oemol_to_inchikey(mol: oechem.OEMol, fixed_hydrogens: bool = False) -> str:
     """
     InChI key string of an OpenEye OEMol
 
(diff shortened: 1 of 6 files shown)


jthorton:
@hmacdope I think this is ready to go and should allow me to finally finish the expanders I hope!

asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -148,6 +148,44 @@ def load_openeye_pdb(
...
+    Load an OpenEye SMILES file containing a single molecule and return it as an
+    OpenEye OEGraphMol object.
+    Parameters
+    ----------
+    smi_fn : Union[str, Path]
+        Path to the SMILES file to load.
+    Returns
+    -------
+    oechem.OEGraphMol
+        An OpenEye OEGraphMol object containing the molecule data from the SDF file.
```
hmacdope, line 162:
```
        An OpenEye OEGraphMol object containing the molecule data from the SDF file.
```

asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -148,6 +148,44 @@ def load_openeye_pdb(
         oechem.OEThrow.Fatal(f"Unable to open {pdb_fn}")
 
 
+def load_openeye_smi(smi_fn: Union[str, Path]) -> oechem.OEGraphMol:
```
hmacdope, line 151: Single smiles only? probably better to have it return all the ones in a file, I know there is a bit of a historical artifact tho this, but still would probably prefer loading all the mols.

asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -148,6 +148,44 @@ def load_openeye_pdb(
...
+        An OpenEye OEGraphMol object containing the molecule data from the SDF file.
+    Raises
+    ------
+    FileNotFoundError
+        If the specified file does not exist.
+    oechem.OEError
+        If the CIF file cannot be opened.
+    Notes
+    -----
+    This function assumes that the SDF file contains a single molecule. If the
```
hmacdope, line 171: See above.

asapdiscovery-data/asapdiscovery/data/openeye.py
```diff
@@ -728,7 +804,7 @@ def _set_SD_data_repr(mol: oechem.OEMol, data: dict[str, Any]) -> oechem.OEMol:
         OpenEye OEMol with SD data set
     """
     # NOTE: use repr to ensure re-reading the SD data will give the same value
-    mol = set_SD_data(mol, {k: repr(v) for k, v in data.items()})
```
hmacdope, line 731: Change comment above if changing this, might also want to wrap in try/except for string cast, unless guaranteed that all incoming objects will be castable.

asapdiscovery-data/asapdiscovery/data/schema_v2/experimental.py
```diff
@@ -47,13 +47,13 @@ class ExperimentalCompoundData(Model):
...
-        """
-        Convert to a dictionary of SD tags
-        """
-        data = self.dict()
-        exp_data = data.pop("experimental_data")
-        # cannot use a nested dict in SD tags, so flatten to two tags
-        data = {str(k): str(v) for k, v in data.items() if v is not None}
-        exp_data = {str(k): float(v) for k, v in exp_data.items() if v is not None}
-        return data, exp_data
+    # def to_SD_tags(self) -> tuple[dict[str, str], dict[str, float]]:
```
hmacdope, line 50: remove commented out bits.

asapdiscovery-data/asapdiscovery/data/schema_v2/identifiers.py
```diff
@@ -0,0 +1,43 @@
...
+        None, description="Unique ID from Postera Manifold API"
+    )
+    manifold_vc_id: Optional[str] = Field(
+        None, description="Unique VC ID (virtual compound ID) from Postera Manifold"
+    )
+    compchem_id: Optional[UUID4] = Field(
+        None, description="Unique ID for P5 compchem reference, unused for now"
+    )
+
+    # def to_SD_tags(self) -> dict[str, str]:
```
hmacdope, line 38: Same here remove commented out bits.

asapdiscovery-data/asapdiscovery/data/schema_v2/identifiers.py
```diff
@@ -0,0 +1,43 @@
+from typing import Optional
```
hmacdope, line 1: If this is going to be a seperate file, move `TargetIdentifiers` here also.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -112,12 +73,12 @@ class Ligand(DataModelAbstractBase):
...
 
     data: str = Field(
-        "",
+        ...,
         description="SDF file stored as a string to hold internal data state",
         repr=False,
     )
-    data_format: DataStorageType = Field(
-        DataStorageType.sdf,
+    data_format: Literal["sdf"] = Field(
```
hmacdope, line 80: Can we keep the enum? It should cast easily to a string anyway, but we can use `Literal`

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -311,60 +298,87 @@ def clear_SD_data(self) -> None:
...
-            )
-        # get reserved attribute names
-        reser_attr_names = [attr.name for attr in self.__fields__.values()]
-        # push all non reserved attribute names to tags
-        data["tags"].update(
-            {k: v for k, v in data.items() if k not in reser_attr_names}
-        )
-        # reinitialise object
-        self.__init__(**data)
+    # def _clear_internal_SD_data(self) -> None:
```
hmacdope, line 301: Remove commented

hmacdope (changes requested):
Looking really good! Just a few things to query. I think this is a much better solution than the one I had.

asapdiscovery-data/asapdiscovery/data/openeye.py
jthorton: Good catch I think everything from the ligand side should be a string as we are enforcing it via typing and using json models so we should be safe to change the type hint to expect dicts of str only.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
jthorton: Found a way to do it just had to run it through JSON to get it to work!

jthorton:
continued in #534

PR #535, opened 2023-10-16
add option to split_sdf script to flatten molecules to 2d

apayne97:
## Description
More ligand sdf tinkering. This was the easiest way I could think to flatten the molecules to make sure that the inclusion of 3d coordinates for the ligands isn't what is messing up hybrid vs shapefit.

## Todos
- [x] add an option to the split sdf code to generate 2d coordinates from the 3d mols

## Status
- [x] complete

At some point we will probably want to replace this with something that happens in the docking code. Doing it this way is kinda annoying bc the default calculated RMSDs are going to be wrong and will need to be redone


hmacdope (approved):
LGTM, just fix the pre-commit complaints.

PR #540, opened 2023-10-18
Adjust molecule filtering to make more sense

### LINKED ISSUE #539: Handle semi-quant data filtering differently
Currently the molecule filtering process treats keeping semi-quant data the same as keeping achiral/racemic/enantiopure molecules. I think the filter for semi-quant should be a lower priority, ie if a molecule passes the other filters, then we check for semi-quant, but we shouldn't include molecules that don't pass the other filters but are semi-quant (which is the current behavior).

kaminow:
Changes filtering molecules that were downloaded from CDD so that data being semi-quantitative is no longer a criteria for inclusion, but only for exclusion if `retain_semiquantitative_data == False`. Fixes #539.


hmacdope (approved):
LGTM :+1:

PR #543, opened 2023-10-19
Fix case in `setLevel` for logging

ijpulidos:
## Description
Quick fix to change the case when setting the level of logging in some cases.

## Status
- [x] Ready to go


hmacdope (approved):
LGTM thanks for fixing

Issue #525, opened 2023-10-10
Check and fix GIF and HTML angles. 


hmacdope:
Check and fix GIF and HTML angles, some off which are broken.

Issue #524, opened 2023-10-10
on prep, align to a representative structure


JenkeScheen:
with `fragalysis` uploads broken, currently a large part of crystal structures that we're pulling in from a google drive folder (ugh) are not aligned to each other. This messes up a whole bunch of html and gif viz viewing angles. We should, for each target, pick a representative complex and when prepping a structure align to that reference.

kaminow:
alignment to a reference structure is already built in to the protein prep pipeline I believe: https://github.com/choderalab/asapdiscovery/blob/main/asapdiscovery-modeling/asapdiscovery/modeling/modeling.py#L45

JenkeScheen:
perhaps we're just not calling it in end2end, @hmacdope ?

PR #547, opened 2023-10-23
fix HTML/GIF view once and for all

JenkeScheen:
Fixes https://github.com/choderalab/asapdiscovery/issues/525 and https://github.com/choderalab/asapdiscovery/issues/524

Viewing angles of HTML pose and MD GIF are wrong often, which causes us to have to do some manual work to find the correct pose angle coordinates. The main reason this is happening is because `fragalysis` uploads are broken and we're pulling crystal structures from a google drive that are not aligned, in some cases. 

Instead, we'll use a `master structure` per target. These master structures will have been 'aligned' to each other so that the ligand + rotation is the same between these. Because of this we'll be able to use _a single viewing angle_ per viz type. 

We'll need to make sure that for whatever new target comes into P3/5 we include a master structure in `metadata`.


JenkeScheen:
@hmacdope this should be ready for review - tested with a Mac1 structure and all looks in order, both for HTML and GIF.

hmacdope (approved):
LGTM!

PR #552, opened 2023-10-25
Make docs and RTD docs build

### LINKED ISSUE #458: Create an API docs page somewhere
It would be super nice to have an API reference (like ReadTheDocs or something) now that we're building out the schematization and everything. Happy to help with this, although I don't have any experience working with these kinds of packages. Tagging @hmacdope @mikemhenry @ijpulidos because I feel like y'all might have done this at some point for other packages.

hmacdope:
Fixes #458

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/_autosummary/
 
 # PyBuilder
 target/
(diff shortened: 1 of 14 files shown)


hmacdope:
@ijpulidos @mikemhenry once again here we are getting workers crashing on tests unrelated to these changes. Any ideas? I will turn off parallelism in data CI, but still very strange.

PR #553, opened 2023-10-26
Fix segfaulting test in openeye CIF read

hmacdope:


hmacdope:
possibly due to unpinning openeye toolkit, investigating.

hmacdope:
Yep confirming this is a segfault - causing regression in `2023.1.1`

hmacdope:
I have removed attempts to load CIF/CIF1 with pdb reading code.

PR #557, opened 2023-10-26
Ml import restructure

### LINKED ISSUE #555: Clean up ML imports to not import from toplevel
The ML subpackage doesn't use the same convention as the rest of the subpackages, where there are no recursive imports and everything must be imported with an absolute import. We should make these the same all round.

hmacdope:
Fixes #555 

@kaminow you should check that all your workflows, including training etc work as expected and don't rely on the toplevel imports  that I haven't caught. Feel free to just push any fixes as you find them.


hmacdope:
@kaminow would you be able to check that all the ML training stuff works as expected? Ill let you review and then if all good approve. :)

kaminow:
@hmacdope would it make sense to just move all these "subpackages" that are just one file up to the top level? that would accomplish (I think) the same goal, but wouldn't (I think) require any changes to any import statements

hmacdope:
@kaminow that is a good idea, a lot of `from inference.inference import inference`. I will make the change.

hmacdope:
@kaminow to finish this one off.

kaminow (approved):
going to approve this since I think it's good to go, but @hmacdope please take one last pass through before merging

PR #568, opened 2023-11-01
Try and fix dask CPU workers

hmacdope:
Adds ability use dask-workers on CPUqueue using suggested fix from MSK HPC.

### DIFF
--- a/asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
+++ b/asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
@@ -124,7 +124,7 @@ class LargeScaleDockingInputs(BaseModel):
     )
 
     dask_cluster_max_workers: PositiveInt = Field(
-        40, description="Maximum number of workers to use for Lilac dask cluster"
+        100, description="Maximum number of workers to use for Lilac dask cluster"
     )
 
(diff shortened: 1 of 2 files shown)


hmacdope:
@JenkeScheen confirmed that this works on lilac, and should also fix the issue of core dumps on workers. Just add `ulimit -c 0` in your supervisory job also.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -124,7 +124,7 @@ class LargeScaleDockingInputs(BaseModel):
     )
 
     dask_cluster_max_workers: PositiveInt = Field(
-        40, description="Maximum number of workers to use for Lilac dask cluster"
+        200, description="Maximum number of workers to use for Lilac dask cluster"
```
JenkeScheen, line 127: why 200?

JenkeScheen (approved):
LGTM!

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
hmacdope: More workers = more fun.

PR #575, opened 2023-11-02
Add example for how to run the full ML pipeline

kaminow:
Add example for how to run the full ML pipeline, starting from downloading the raw data, and ending with training a models.


examples/ml_pipeline.md
```diff
@@ -0,0 +1,208 @@
...
+environment variable, or in a file that is then passed to the script. By default, the
+script will download all Moonshot compounds that have fluorescence values for SARS-Cov-2
+Mpro, but other searches can be specified using the CLI args.
+
+As an example, we will download the default search, keeping all molecules that are
+achiral or enantiopure and including semiquantitative data. We assume in this case that
+the `CDDTOKEN` env variable has been set appropriately. This call also saves the
+unfiltered and unprocessed file as a cache file to avoid repeated CDD download calls:
+```bash
+python /path/to/data/scripts/download_moonshot_data.py \
```
hmacdope, line 24: Are these not package scripts that you can just run like `download_moonshot_data`? Longterm these could become part of the mega cli

examples/ml_pipeline.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Run Docking
+The final step in preparing the structural data is to run docking. There are a number of
+options that can be configured, but this a minimal example that will dock each compound
+to its top match from the MCS search. This step is again parallelized (`-n 32`), which
+can be adjusted based on your machine. Note that the single quotes in the `-r` option
+are required to prevent the shell from expanding the wildcard.
+```bash
+mkdir -p docking_results
+python /path/to/docking/scripts/run_docking_oe.py \
```
hmacdope, line 136: Can we make this step use new docking or is it to convoluted atm?

hmacdope (commented):
Great start, two quick queries.

examples/ml_pipeline.md
kaminow: yep, good point. I updated all the script calls to just use the CLI package call. would definitely be good to add everything to `click` at some point, but that's something for later I think

examples/ml_pipeline.md
kaminow: I think this'll need the large scale docking since it's for the whole CDD. I didn't think we had that updated yet, but if we do I'm happy to rewrite to use it

hmacdope: We do have large scale docking ready but they are probably not compatible if relying on file names etc. Good enough for now and we can iterate.

hmacdope:
As a future example @kaminow and as part of #601 (or extension thereof) it would be good to be able to do the whole pipline in-memory if that makes sense? That way a notebook could be self contained etc and the ML pipeline will be more robust wrt FS and execution location etc.

hmacdope:
@kaminow is this outdated now? If so can you close :)

kaminow:
@hmacdope yep, this is superseded by #601. re your earlier comment (that totally slipped through, sorry), I agree, and I'll open a new issue to address

PR #580, opened 2023-11-03
add mamba requirement to main readme

apayne97:
## Description
I was completely unable to build the repo with conda, so I think we should make that clear in the main readme.

## Status
- [x] Ready to go


hmacdope (approved):
Ta!

PR #585, opened 2023-11-05
Improve docking efficiency

### LINKED ISSUE #584: Improve efficiency of docking: initial steps
* We should make retries in the docking code configurable and remove the chemgauss 4 scoring, that is done later by the scorer. 
* We should also make the use of Omega default False in the large-scale docking workflow, to improve efficency.

hmacdope:
Fixes #584 

@JenkeScheen 
@apayne97  importantly changes the default to not do oeomega enumeration for each compound before docking in large scale workflow


apayne97:
just to be clear - this is basically just an update to the cli (with some minor naming changes)? we've had the capability to do this for a while, right?

hmacdope:
Yep.

PR #591, opened 2023-11-07
Fix protein prep loop db filepath 

jthorton:
## Description
This PR fixes an issue where the loop db file path is set to `str(None)` when not supplied by mistake causing a fatal error with openeye. 

## Todos
  - [X] Add test for the case.

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


jthorton:
Note that if I revert the fix to the loop db path the test dies with exit code 1.

hmacdope (changes requested):
Pending clarification on structure release

hmacdope (approved):

PR #593, opened 2023-11-07
Add minimal redundant networks

jthorton:
## Description
This PR adds support for the new minimal redundant network planning method in openfe and makes it the default in `asap-alchemy`.

## Todos
  - [X] add test for the network type

## Questions
- [ ] 

## Status
- [ ] Ready to go


jthorton:
blocked by alchemiscale conda recipe which is pinned to openfe 13 
https://anaconda.org/conda-forge/alchemiscale-client/files

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/test_fec_schema.py
```diff
@@ -70,6 +70,7 @@ def test_mapper_provenance(mapper, programs):
         pytest.param("radial", id="Radial"),
         pytest.param("maximal", id="Maximal"),
         pytest.param("minimal_spanning", id="Minimal Spanning"),
+        pytest.param("minimal_redundant_network", id="Minimal redundan"),
```
JenkeScheen, line 73:
```suggestion
        pytest.param("minimal_redundant_network", id="Minimal redundant"),
```

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/network.py
```diff
@@ -102,6 +104,8 @@ def _get_network_plan(self):
             return openfe.ligand_network_planning.generate_radial_network
         elif self.network_planning_method == "maximal":
             return openfe.ligand_network_planning.generate_maximal_network
+        elif self.network_planning_method == "minimal_redundant_network":
+            return openfe.ligand_network_planning.generate_minimal_redundant_network
```
JenkeScheen, line 108: you should probably expose `mst_num` as it can be useful to increase it to >2 in some cases.

JenkeScheen:
> blocked by alchemiscale conda recipe which is pinned to openfe 13

@dotsdl

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/network.py
jthorton: Ah thanks I missed that was an option, I think I'll have to rework the `network_planning_method` field to be more like the `atom_mapping_engine` field so each planning method can have options as well.

JenkeScheen:
may be good to also update cinnabar to 0.4.1 for this update.

jthorton:
@JenkeScheen see [2abd12a](https://github.com/choderalab/asapdiscovery/pull/593/commits/2abd12a9755dfc46574d5b9d8b6b69f0e376fff1) for the proposed support for options in the planning method. The only issue is that this is a breaking change to the schema so all old planned networks can no longer be loaded from without being updated first. As we haven't run too many calculations yet we should be okay to make this change and we could include something to update the schema either a CLI or the from file method handle updates.

PR #594, opened 2023-11-07
Enable simpler docking workflow using composable parts

apayne97:
## Description
I wanted to make a simpler workflow in order to test multi-receptor docking for #590 , which I found difficult because a lot of code I wanted to use in large_docking_workflow wasn't easily reusable in a new workflow. This PR is an attempt to split out some of the stuff that I think should be base options.

## Todos
  - [x] DockingInputsBase class
  - [x] POSITInputs class
  - [x] move POSIT specific stuff to docking.openeye
  - [x] move some other stuff to DockingBase
  - [x] put POSITDocker inputs back in the class
  - [x] rename / create / move DockingInputs to WorkflowInputs
  - [x] create `cross_docker_workflow.py` that's basically copy-paste from the large-scale-docking-workflow
  - [x] add `cross_docking_workflow` cli
  - [x] (add to a future issue / PR on abstracting Workflow objects) --> #596 

## Questions
- [x] decide if we want to continue with having input schema for posit functions or not --> the idea is that anytime we have a class that is handling the functionality, we can let that class handle the validation. but for inputs to functions, having a separate inputs class is helpful. (although I guess I'd argue that in this case we might as well just convert the function and the inputs schema into its own class)

## Status
- [x] Ready to go


asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    Enum for POSIT relax modes
+    """
+
+    CLASH = oedocking.OEPoseRelaxMode_CLASHED
+    ALL = oedocking.OEPoseRelaxMode_ALL
+    NONE = oedocking.OEPoseRelaxMode_NONE
+
+
+class POSITInputs(DockingInputsBase):
```
hmacdope, line 58: We have moved away from using input Schema as separate classes/objects. It is a bit too decoupled, and makes more sense for openeye because of C++ stuff/  Instead these can just be attributes of the workflow class itself. You can share via inheritance as you have done here also.

asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,375 @@
...
+        False, description="Allow clashing poses in last stage of docking"
+    )
+    allow_retries: bool = Field(
+        True,
+        description="Allow retries with different options if docking fails initially",
+    )
+
+
+class POSITDocker(DockingBase):
+    posit_inputs: POSITInputs = Field(..., description="POSIT inputs")
```
hmacdope, line 104: See above

asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def provenance(self) -> dict[str, str]:
+        return {
+            "oechem": oechem.OEChemGetVersion(),
+            "oeomega": oeomega.OEOmegaGetVersion(),
+            "oedocking": oedocking.OEDockingGetVersion(),
+        }
+
+
+class POSITDockingResults(DockingResult):
```
hmacdope, line 329: I like this distinction. Makes sense.

asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,375 @@
...
+                    raise ValueError(
+                        f"docking failed for input pair with compound name: {pair.ligand.compound_name}, smiles: {pair.ligand.smiles} and target name: {pair.complex.target.target_name}"
+                    )
+                else:
+                    raise ValueError(f"Unknown error handling option {error}")
+
+        return docking_results
+
+    @staticmethod
+    def write_docking_files(
```
hmacdope, line 262: Should probably be a method of the base class, thats my original omission.

asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,375 @@
+"""
+This module contains the inputs, docker, and output schema for using POSIT
```
hmacdope, line 2: Makes sense to seperate into its own module.

asapdiscovery-docking/asapdiscovery/docking/openeye.py
hmacdope: I will say the exception to this is as an argument to a function that needs a tonne of validation.

apayne97: I am confused because you have a LargeScaleDockingInputs class (which is where I modeled this from):
https://github.com/choderalab/asapdiscovery/blob/3839851460c93f0555f3044fdfd0c92f27b94a12/asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py#L37-L250

I don't have strong opinions either way but I think we should be consistent with implementation?

apayne97: I think if we have results schema it makes some sense to have inputs schema

asapdiscovery-docking/asapdiscovery/docking/openeye.py
apayne97: done here: https://github.com/choderalab/asapdiscovery/pull/594/commits/e38cd3cc7db87b5fbedfad964adfd39e58435464

asapdiscovery-docking/asapdiscovery/docking/openeye.py
hmacdope: Ah okay perhaps the naming and API design is a bit confusing, lets discuss tommorow. But basically:

The LargeScaleDockingInputs thing is to validate a workflow, ie collect a tonne of information, and validation to run a **function** that does the whole workflow, stopping a situation like this. 

```python
workflow_func(a=1, b=2, c=3, d=42, e=... # takes a lot of information that then needs to be reconciled. )
```
Instead with 

```python
workflow_func(input_schema) 
```
The idea is that the function doesn't have to do any sanitation of the inputs. 

For **classes** that run steps (components), I don't see the need to separate the settings and object itself, as the object can easily contain its own settings if that makes sense? All the current component classes use the second pattern below roughly, but perhaps I need to formalize this in a document.

The only difference between 

```python
class DockingInputs 
    a=1
    b=2
    # use a, b validation in class directly
```
 and 

```python
class Docker
    inputs = DockingInputs()
   ...
   # use inputs.a, inputs.b validation in DockingInputs
```

Is that there are two classes instead of one and its more nested. Both could work though to be perfectly clear

asapdiscovery-docking/asapdiscovery/docking/workflows/workflows.py
```diff
@@ -0,0 +1,129 @@
...
+
+    dask_cluster_max_workers: PositiveInt = Field(
+        200, description="Maximum number of workers to use for Lilac dask cluster"
+    )
+
+    n_select: PositiveInt = Field(
+        5, description="Number of targets to dock each ligand against."
+    )
+
+    # TODO: make this an "abstractmethod" kinda thing that needs to be re-named in subclasses
```
hmacdope, line 72: I think you can just leave out `"docking_workflow_base"` as default param and the model can't be instantiated.

eg 

```python
    logname: str = Field(..., description="Name of the log file.")


```

asapdiscovery-docking/asapdiscovery/docking/tests/cli/test_docking_cli.py
```diff
@@ -132,4 +132,35 @@ def test_large_docking_cli_pdb_file(ligand_file, pdb_file, du_cache, tmp_path):
     assert click_success(result)
 
 
+@pytest.mark.skipif(
+    os.getenv("RUNNER_OS") == "macOS", reason="Docking tests slow on GHA on macOS"
+)
+def test_cross_docking_cli_structure_directory_du_cache(
```
hmacdope, line 138: Nice!

asapdiscovery-docking/asapdiscovery/docking/openeye.py
```diff
@@ -0,0 +1,298 @@
...
+        description="Allow retries with different options if docking fails initially",
+    )
+
+    @staticmethod
+    def to_result_type():
+        return POSITDockingResults
+
+    @root_validator
+    @classmethod
+    def _output_dir_write_file(cls, values):
```
hmacdope, line 96: My issue, but this can be removed, as `write_file` is no longer an attr.

asapdiscovery-docking/asapdiscovery/docking/cli.py
```diff
@@ -131,5 +135,81 @@ def large_scale(
...
+@structure_dir
+@gen_cache
+@cache_dir
+@cache_type
+@dask_args
+@output_dir
+@input_json
+def cross_docking(
+    target: TargetTags,
+    use_omega: bool = False,
```
hmacdope, line 170: Make default here same as in base class as will be overridden. 

We need to be careful about being overidden by CLI args also.

hmacdope (changes requested):
Few small changes, but looking great.

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
```diff
@@ -0,0 +1,234 @@
...
+    # write docking results
+    logger.info("Writing docking results")
+    POSITDocker.write_docking_files(results, output_dir / "docking_results")
+
+    # add chemgauss4 scorer
+    scorers = [ChemGauss4Scorer()]
+
+    # score results
+    logger.info("Scoring docking results")
+    scorer = MetaScorer(scorers=scorers)
```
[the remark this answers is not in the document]
apayne97: understood, I think I'm just gonna keep it meta in case I want to use others in the future

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
```diff
@@ -0,0 +1,234 @@
...
+        logger.info(f"Caching prepped complexes to {cache_path}")
+        for cache_type in inputs.cache_type:
+            prepper.cache(prepped_complexes, cache_path, type=cache_type)
+
+    # define selector and select pairs
+    # using dask here is too memory intensive as each worker needs a copy of all the complexes in memory
+    # which are quite large themselves, is only effective for large numbers of ligands and small numbers of complexes
+    logger.info("Selecting pairs for docking based on MCS")
+    selector = MCSSelector()
+    pairs = selector.select(
```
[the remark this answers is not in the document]
apayne97: Pairwise! Thank you!

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
```diff
@@ -0,0 +1,234 @@
...
+    )
+
+    n_pairs = len(pairs)
+    logger.info(f"Selected {n_pairs} pairs for docking")
+
+    del prepped_complexes
+
+    # dock pairs
+    logger.info("Running docking on selected pairs")
+    docker = POSITDocker(use_omega=inputs.use_omega, allow_retries=inputs.allow_retries)
```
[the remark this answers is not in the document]
apayne97: yep, in a future PR we'll do what Ben said but for now I'll just pass all the available options to posit

asapdiscovery-docking/asapdiscovery/docking/workflows/workflows.py
apayne97: ahhh i didn't know you could do that, that's cool

asapdiscovery-docking/asapdiscovery/docking/cli.py
apayne97: bah yeah this is a bit of a concern. should be ok though bc the overriding goes in the direction that makes sense (cli --> workflow --> class).

hmacdope: Perhaps @jthorton knows a design that circumvents this?

hmacdope:
@apayne97 large scale docking WF inputs should also subclass WorkflowInputsBase, ill make a seperate PR.

PR #595, opened 2023-11-08
Rework small scale docking

hmacdope:
FIxes #510

PR #596, opened 2023-11-08
[DNM] add docking workflow abstraction rough draft

### LINKED ISSUE #597: Convert Workflow functions into classes
to enable inheritance between workflows

apayne97:
## Description
Depends on #594 #595 and a discussion about how to proceed but I think abstracting away some of the code in the docking workflows would be nice to do!

Would close #597 

## Todos
  - [ ] TODO 1

## Questions
- [ ] What's the best way to create a recipe of functions to run?

## Status
- [ ] Ready to go


hmacdope:
@apayne97 at this stage we are pretty set on the way things are (for better or worse). Just as part of cleaning up PRs perhaps we could close this. It will live on in closed, but just looking to clean up things we likely won't merge.

apayne97:
Yeah, fair enough! We can come back to it if needed

PR #608, opened 2023-11-09
Protein prep updates

jthorton:
## Description
This PR updates the protein prep tool to write out the prepped complex as json oedu and pdb and sdf for visualisation. This helps downstream with inputs into the asap-alchemy prep cli.

## Todos
  - [x] make sure the cache still works

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -171,3 +172,17 @@ def from_oedu_file(cls, oedu_file: Union[str, Path], **kwargs) -> "PreppedTarget
...
+
+        Parameters
+        ----------
+        filename: The name of the pdb file the target should be writen to.
+
+        """
+        oe_receptor = oechem.OEGraphMol()
+        oedu = self.to_oedu()
+        oedu.GetProtein(oe_receptor)
+        save_openeye_pdb(oe_receptor, pdb_fn=filename)
```
apayne97, line 188: in the past it's been necessary to "re-perceive" residues in the protein (after adding hydrogens) otherwise CONECT records and atom ordering gets wonky. This is the function we've used to do that:
https://github.com/choderalab/asapdiscovery/blob/c48443f9632a84f57df7f9fa7ede16174cdc49c3/asapdiscovery-data/asapdiscovery/data/openeye.py#L526

I don't think it would hurt to add and it could definitely help, esp with CONECT records for later simulations.

in the past we've used this function to split an oedu into the protein / ligand components (although not much different functionality than what you've done here, just wanted to make sure you saw it).
https://github.com/choderalab/asapdiscovery/blob/c48443f9632a84f57df7f9fa7ede16174cdc49c3/asapdiscovery-modeling/asapdiscovery/modeling/modeling.py#L732

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -78,32 +78,44 @@ def provenance(self) -> dict[str, str]:
...
+            # if type == CacheType.DesignUnit:
+            #     du_name = pc.target.target_name + ".oedu"
+            #     du_path = dir / du_name
+            #     if not du_path.exists():
+            #         pc.target.to_oedu_file(du_path)
+            # elif type == CacheType.JSON:
+            #     json_name = pc.target.target_name + ".json"
+            #     json_path = dir / json_name
+            #     if not json_path.exists():
+            #         pc.to_json_file(json_path)
```
apayne97, line 112: so just to be clear - now every time we prep something, instead of deciding how we want to cache it (and only being able to pick one at a time), now we will cache it as all the types? I like this a lot actually

apayne97 (commented):
Not gonna approve yet bc I think there are still some things you're gonna add but I like where this is going

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
jthorton: Correct, I think the JSON should be the default as all other models are easily made from this and it's complete and usable downstream. We found loading from oedu was missing some data like a target name that made automation of later steps difficult. The pdb and sdf are mainly just for easy vis and could be made optional though.

asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
jthorton: Ah thanks, thats good to know about the perceive residues, I wonder if would there be any harm in doing this by default in the `save_openeye_pdb` function? No I missed the split function I'll switch to using that!

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -78,32 +78,44 @@ def provenance(self) -> dict[str, str]:
     def cache(
         prepped_complexes: list[PreppedComplex],
         dir: Union[str, Path],
-        type=CacheType.DesignUnit,
+        # type=CacheType.DesignUnit,
```
hmacdope, line 81: Clean up commented

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -78,32 +78,44 @@ def provenance(self) -> dict[str, str]:
...
-                du_path = dir / du_name
-                if not du_path.exists():
-                    pc.target.to_oedu_file(du_path)
-            elif type == CacheType.JSON:
-                json_name = pc.target.target_name + ".json"
-                json_path = dir / json_name
-                if not json_path.exists():
-                    pc.to_json_file(json_path)
+            # create a folder for the complex data
+            complex_folder = dir.joinpath(pc.target.target_name)
```
hmacdope, line 92: We may need to be careful here about targets with the same name.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -78,32 +78,44 @@ def provenance(self) -> dict[str, str]:
...
+            pc.to_json_file(complex_folder.joinpath(pc.target.target_name + ".json"))
+            pc.target.to_oedu_file(
+                complex_folder.joinpath(pc.target.target_name + ".oedu")
+            )
+            pc.target.to_pdb_file(
+                complex_folder.joinpath(pc.target.target_name + ".pdb")
+            )
+            pc.ligand.to_sdf(complex_folder.joinpath(pc.ligand.compound_name + ".sdf"))
+
+            # if type == CacheType.DesignUnit:
```
hmacdope, line 103: Clean up commented.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
hmacdope: You could specify cache type multiple times but I agree this is just simpler overall at the slight cost of some disk space.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -78,32 +78,44 @@ def provenance(self) -> dict[str, str]:
...
+            #     if not json_path.exists():
+            #         pc.to_json_file(json_path)
 
     @staticmethod
     def load_cache(
         complexes: list[Complex],
         cache_dir: Union[str, Path],
-        cache_type: CacheType,
+        # cache_type: CacheType,
         fail_missing_cache: bool = False,
```
hmacdope, line 119: Clean commented

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -112,43 +124,21 @@ def load_cache(
...
-                if du_path.exists():
-                    prepped_target = PreppedTarget.from_oedu_file(
-                        du_path,
-                        ids=complex.target.ids,
-                        target_name=complex.target.target_name,
-                        ligand_chain=complex.ligand_chain,
-                    )
-                    pc = PreppedComplex(target=prepped_target, ligand=complex.ligand)
-                    prepped_complexes.append(pc)
+        for complex_target in complexes:
```
hmacdope, line 128: Ok yeah makes sense to just always load from JSON.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -183,7 +173,7 @@ class ProteinPrepper(ProteinPrepperBase):
     cache_dir: Optional[Path] = Field(
         None, description="Path to a directory where design units are cached"
     )
-    cache_type: CacheType = Field(CacheType.DesignUnit, description="Type of cache")
+    # cache_type: CacheType = Field(CacheType.DesignUnit, description="Type of cache")
```
hmacdope, line 176: Clean commented.

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
```diff
@@ -81,9 +79,9 @@ class ProteinPrepInputs(BaseModel):
         description="Path to a directory where generated prepped complexes should be cached",
     )
 
-    cache_type: Optional[list[str]] = Field(
-        [CacheType.DesignUnit], description="The types of cache to use."
-    )
+    # cache_type: Optional[list[str]] = Field(
```
hmacdope, line 82: Commented

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
```diff
@@ -112,7 +110,7 @@ class ProteinPrepInputs(BaseModel):
 
     dask_cluster_n_workers: PositiveInt = Field(
         10,
-        description="Number of workers to use as inital guess for Lilac dask cluster",
+        description="Number of workers to use as initial guess for Lilac dask cluster",
```
hmacdope, line 113: Ta

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
```diff
@@ -154,13 +152,13 @@ def check_inputs(cls, values):
...
         return values
 
-    @validator("cache_type")
-    @classmethod
-    def check_cache_type(cls, v):
-        # must be unique
-        if len(v) != len(set(v)):
-            raise ValueError("cache_type must be unique")
-        return v
+    # @validator("cache_type")
```
hmacdope, line 155: Clean commented.

hmacdope (approved):
Agree, this is going to makwe things a lot simpler, only question is duplicate target names result in same file (pre-existing) problem also, we may want to address.

hmacdope (changes requested):
Whoops sorry meant to request changes.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
jthorton: Good point we could change this to `YYYY-MM-DD-target-name`?

jthorton: Or is there a unique reproducible hash we could use with the target name maybe sha256 of the pdb file, I guess this depends under what conditions are two prepared complexes regarded the same?

jthorton: The current method in [8c5e275](https://github.com/choderalab/asapdiscovery/pull/608/commits/8c5e27590ab870e639d8703a878dd2ef291af298) uses a hash based on the input target to check for matches in the cache.

jthorton:
I have no idea why but its pulling in changes not related to this PR again! Ill open a new clean one to replace this one.

PR #621, opened 2023-11-13
Fix `eq` bug in `schema_base`

### LINKED ISSUE #619: Fix not-equal bug in DataModelAbstractBase
The not-equal method of DataModelAbstractBase should just be `not _eq__`, right? Otherwise if your class that subclasses it doesn't have a `data` method, it throws an error.

https://github.com/choderalab/asapdiscovery/blob/671776df4f1cbb8f24a200b175478f554e26299d/asapdiscovery-data/asapdiscovery/data/schema_v2/schema_base.py#L79-L89

hmacdope:
Fixes #619 

Thanks @apayne97 for spotting.


apayne97 (approved):
LGTM :)

PR #624, opened 2023-11-14
Alchemy cinnabar update

jthorton:
## Description
This PR updates the asap-alchemy cinnabar interface to use the new API which allows us to create absolute and relative data frames. 

## Todos
  - [x] add a test using computed tyk2 results which converts to an fe map and calculates the dataframes

## Questions
- [ ] Do we want to extend the absolute dataframe and add 2D pictures of the ligands and write it to an interactive HTML?

## Status
- [ ] Ready to go


JenkeScheen:
@jthorton when adding measurements, `Measurement` now requires `labelA` and `labelB`, even when it's an absolute measurement, in which case there is no A/B. Am I missing something?

jthorton:
Ah yes you have to use a special flag for the ligand state A [see](https://github.com/OpenFreeEnergy/cinnabar/blob/95cca4e64bb64df5e9fc4b83bc3c34cfaaf0a4a6/cinnabar/femap.py#L171C1-L178C14).

hmacdope:
@jthorton is this still WIP or ready for review.

hmacdope (approved):
LGTM bossman.

Issue #571, opened 2023-11-02
Make prepped structure caching more "cache" like


hmacdope:
Currently the way structure caching works is more like a dump than a cache, requiring all structures that need to be prepped be present in the cache.

Instead the cache should act as a an actual "cache ", ie searching for if the target has been prepped before by a name based match, and IF NOT prepping it and adding to the cache. 

One question here is how we avoid collisions, eg with structures that have unclear names, eg `protein.pdb`. We should probably also add some heuristics to avoid overwrrites and collisions.

kaminow:
adding on to this, it seems like the `protein_prep_workflow` quietly deletes and replaces whatever `output_dir` you give it, which might be better to have as an option

apayne97:
Also the current workflow waits until all the prepped complexes are made to save them - should be done in parallel otherwise if something bad happens you lose everything: https://github.com/choderalab/asapdiscovery/blob/0432f69fc99c85a923a498bf6ba6f117d3e703f3/asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py#L260-L271

hmacdope:
I guess my idea was that saving to disk should be seperate of the computation, ie you can cache prepped complexes any time, not just when they are being prepped.

Would a partially complete "cache" not be misleading? You would then have to remember that it was only partially finished (and most probably remove it). This is also partially due to the limitations in current caching which treats a "cache" (inverted commas) as a fully complete dump of the prepped structures.

hmacdope:
Fixed

PR #627, opened 2023-11-15
Protein-prep updates v2

### LINKED ISSUE #574: Save metadata alongside DU cache
Related to #571 

When we cache DUs, it would be nice if we could store some metadata about what has been prepped alongside them. I had previously done this using the `modeler.schema.PreppedTargets.to_json` method using the Pydantic ability to convert schema objects to json.

jthorton:
## Description
This PR supersedes #608 in updating the protein prep CLI. The CLI will now write out the prepared receptor and ligand in all possible formats JSON, oedu and pdb + sdf for vis. The cache is also updated to reuse prepared structures where possible this is based on a unique complex hash based on the pdb file used for the target and the inchikey of the bound ligand.

Fixes #574  , #571 

## Todos
  - [X] test that cached structures can be reused via cli

## Questions
- [ ] 

## Status
- [X] Ready to go


asapdiscovery-data/asapdiscovery/data/schema_v2/target.py
```diff
@@ -138,6 +146,11 @@ class PreppedTarget(DataModelAbstractBase):
         description="Enum describing the data storage method",
         allow_mutation=False,
     )
+    target_hash: str = Field(
```
hmacdope, line 149: Could this also be done for `Target` not just `PreppedTarget` just so the interface is consistent?

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -74,77 +84,40 @@ def provenance(self) -> dict[str, str]:
...
-                du_path = dir / du_name
-                if not du_path.exists():
-                    pc.target.to_oedu_file(du_path)
-            elif type == CacheType.JSON:
-                json_name = pc.target.target_name + ".json"
-                json_path = dir / json_name
-                if not json_path.exists():
-                    pc.to_json_file(json_path)
+            # create a folder for the complex data
+            complex_folder = dir.joinpath(f"{pc.target.target_name}-{pc.hash()}")
```
hmacdope, line 97: could we make this logic a method of PreppedComplex or Complex? We will need to re-use it in the docking workflow probably. 

Something like `pc.to_unique_folder()` naming up to you.

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
```diff
@@ -249,6 +231,35 @@ def protein_prep_workflow(inputs: ProteinPrepInputs):
         target_kwargs={"target_name": "ref"},
         ligand_kwargs={"compound_name": "ref_ligand"},
     )
+
+    # setup the path to cached structures
+    cache_path = output_dir / inputs.gen_cache
+
+    # load the cache
```
hmacdope, line 238: Is it possible to fold the cache checking logic inside the ProteinPrep class or related, the reason this is important is so that other workflows that read cache OR prep, eg large or small scale docking don't have to duplicate this validation logic. I could be misunderstanding how this works though.

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
```diff
@@ -249,6 +231,35 @@ def protein_prep_workflow(inputs: ProteinPrepInputs):
...
+    try:
+        cached_complexs = ProteinPrepper.load_cache(cache_dir=cache_path)
```
hmacdope, line 240: With load cache now outside the function other workflows now need to do same right? might be easier to hide behind the class.

hmacdope (changes requested):
I think we need to think about hiding caching detail inside the class, happy to chat about this at some point to work out best way forward. LMK if I am not understanding correctly.

hmacdope:
@hmacdope to test tommorow to confirm caching working with project support workflows.

asapdiscovery-modeling/asapdiscovery/modeling/workflows/protein_prep.py
hmacdope: Fixed

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -76,78 +159,43 @@ def provenance(self) -> dict[str, str]:
...
-                du_path = dir / du_name
-                if not du_path.exists():
-                    pc.target.to_oedu_file(du_path)
-            elif type == CacheType.JSON:
-                json_name = pc.target.target_name + ".json"
-                json_path = dir / json_name
-                if not json_path.exists():
-                    pc.to_json_file(json_path)
+            # create a folder for the complex data
+            complex_folder = cache_dir.joinpath(pc.unique_name())
```
hmacdope, line 175: Skip if filename present.

jthorton:
@hmacdope in [cdbe817](https://github.com/choderalab/asapdiscovery/pull/627/commits/cdbe8172d9849437beaf0672b2d8389bd330cc86) I removed the root validator check for `cache_dir` and `save_to_cache` from `DockingWorkflowInputsBase` so that if `save_to_cache` is True and `cache_dir=None` then nothing happens instead of an error. Feel free to change this back if you want but the default value of `save_to_cache` would have to become false.

jthorton:
Turns out #663 might have been caused by me in this PR I have added a fix in [45d9227](https://github.com/choderalab/asapdiscovery/pull/627/commits/45d92274d7cb40beb17d0f0a67c9d90ceff0ed4b). What would be a good test for this measure the ligand distances to the protein before and after prep and make sure they match?

JenkeScheen:
> measure the ligand distances to the protein before and after prep and make sure they match?

yeah that would work! can round to 0.1A or something just to make sure we don't fail the test by a rounding error.

hmacdope:
> Turns out #663 might have been caused by me in this PR I have added a fix in [45d9227](https://github.com/choderalab/asapdiscovery/pull/627/commits/45d92274d7cb40beb17d0f0a67c9d90ceff0ed4b). What would be a good test for this measure the ligand distances to the protein before and after prep and make sure they match?

Great find, you legend. linking #633 as being closed by this PR.

hmacdope (approved):
@jthorton  @JenkeScheen I am going to approve here.  Thanks so much for doing this! Up to you if you want to add test for align-ligand proximity here or push to an issue.

PR #601, opened 2023-11-08
[WIP] Schematize ML components

kaminow:
Convert all the parts of the ML workflow to use nice pydantic schema. Closes #450.

~~Adding this here so it doesn't get lost: I changed the `mtenn` pin in the conda env files so the CI will run. THIS SHOULD BE CHANGED BACK BEFORE MERGING THIS PR~~
This has been done. In a future PR, we should change the `mtenn` dep from a github install to a `conda-forge` install now that it's up there. Should be quick, but not doing it here just for the sake of compartmentalization.

PR #641, opened 2023-11-20
Add new models trained with better learning rate

kaminow:
Add new SchNet model trained with a lower learning rate to avoid the weights collapsing. These models were trained using `mtenn v0.4.0`, so probably shouldn't be used with previous versions. This is the version that's pinned in the `asapdiscovery` conda env files, so it shouldn't cause any issues.

### DIFF
--- a/asapdiscovery-ml/asapdiscovery/ml/asap_models.yaml
+++ b/asapdiscovery-ml/asapdiscovery/ml/asap_models.yaml
@@ -201,7 +201,7 @@ asapdiscovery-SARS-CoV-2-Mpro-schnet-2023.11.20:
     sha256hash: 18bc1633f15aa892de48d691356aad63894bfc6135f92e9ec7ac27a6c36edb91
   config:
     resource: asapdiscovery-SARS-CoV-2-Mpro-schnet-2023.11.20.json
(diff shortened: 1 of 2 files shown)


hmacdope (approved):
LGTM!

hmacdope:
@kaminow need to fix for CI, seems like we were checking that the pIC50 > 0 but is ~ -2000, for our test case. Is this expected?

kaminow:
no it absolutely is not lol, I'm looking into it

kaminow:
this should work now. it turns out there was a config option that wasn't getting passed around the way I thought it was, so I've included it explicitly now in the model `config.json` file. as a side note, this is something that won't be an issue once #601 is finished, since it was just a matter of me being dumb and not properly keeping track of all the model config options

kaminow:
@hmacdope looks like everything is green now! I'll leave the merging up to you in case you want to take one last look

hmacdope:
> @hmacdope looks like everything is green now! I'll leave the merging up to you in case you want to take one last look

I am just going to fix that test that you had to change so we dont have to re-fix it all the time but other than that thanks so much for this, looks great.

hmacdope (approved):

PR #652, opened 2023-11-26
Large scale upload all results to Postera + HTML viz + enhance ability to join with existing postera data

### LINKED ISSUE #626: `large docking` workflow should push results to all molecules
Now that we have an end-to-end _large_ docking workflow that is able to pull/push results to Manifold, we need to adjust how we push. Currently we filter to `top n`, then only push to results to those `n` molecules in Manifold. This ends up looking like this when the moleculeset in question has `>n` designs:
image

which is okay, but there's a lot of data missing. There are several use cases where we or Medchemists may want to see negative data (e.g. "this design that we had high hopes for isn't in the `top n`, what happened?"): so we should be pushing all results to enable this. This would require an extra column/tag that signals to the moleculeset user that a design is a hit, perhaps a bool `in_silico_<target>_docking_hit_msk`?

Additionally, for the `top n` hits we want to be generating interactive HTML S3 URLs. In cases where folks may want to look at the non-hits, they should submit a small-scale docking ticket so that we can generate the HTML and MD GIF.

hmacdope:
Fixes #626 #653

Adds ability to upload all results from a large-scale docking run to postera, not just the top N, and adds visualisation. 

Also improves ability to join with existing postera molecule set,

Adds deduplication to project support workflows. 


asapdiscovery-data/asapdiscovery/data/postera/manifold_data_tags.yaml
```diff
@@ -61,6 +61,12 @@ manifold_outputs:
       units:
       postfix: msk
 
+  - docking-hit:
```
hmacdope, line 64: @JenkeScheen  we have added tags of the form  `in-silico_SARS-CoV-2-Mpro_docking-hit_msk` as per previous discussions to mark whether something is a hit by `top-n` criteria in large-scale docking.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -275,13 +302,71 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
...
+    # rename visualisations target id column to POSIT structure tag so we can join
+    pose_visualizatons.rename(
+        columns={
+            DockingResultCols.TARGET_ID.value: DockingResultCols.DOCKING_STRUCTURE_POSIT.value
+        },
+        inplace=True,
+    )
+
+    # join the two dataframes on ligand_id, target_id and smiles
+    scores_df = scores_df.merge(
```
hmacdope, line 333: Can optimise this slight mess down the track @JenkeScheen

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -326,31 +411,43 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
         data_intermediates / "docking_scores_filtered_sorted.csv", index=False
     )
 
-    scores_df = scores_df.drop_duplicates(subset=[DockingResultCols.LIGAND_ID.value])
+    scores_df = scores_df.drop_duplicates(
+        subset=[DockingResultCols.INCHIKEY.value], keep="first"
```
hmacdope, line 415: Added `first` for clarity, but is default.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -363,21 +460,59 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
...
+        if made_new_molset:
+            logger.info(f"Made new molecule set with name: {molset_name}")
+        else:
+            molset_name = inputs.postera_molset_name
+
+        logger.info("Uploading artifacts to PostEra")
+
+        # make an uploader for the poses and upload them
+
+        artifact_columns = [
```
hmacdope, line 498: Visualisations are being made and pushed for all compounds as take < 1 sec per compound.

asapdiscovery-data/asapdiscovery/data/postera/postera_uploader.py
```diff
@@ -64,17 +69,42 @@ def push(self, df) -> None:
...
                 df_copy = self.join_with_manifold_data(df_copy, new_data)
 
-            ms_api.update_molecules_from_df_with_manifold_validation(
-                molecule_set_id=molset_id,
-                df=df_copy,
-                id_field=self.id_field,
-                smiles_field=self.smiles_field,
-                overwrite=self.overwrite,
-            )
+            # find rows with blank id, they need to be added to molset, using **add** endpoint rather than **update**
```
hmacdope, line 78: We can now append to existing molecule sets with much better fidelity.

hmacdope (commented):
Few notes @JenkeScheen

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
JenkeScheen: great!

asapdiscovery-data/asapdiscovery/data/postera/manifold_data_tags.yaml
JenkeScheen: Perfect, I'll make sure to instruct the medchemists to use this bool.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
JenkeScheen: why do we get duplicates here?

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -326,31 +411,43 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
...
-    # take top n results
-    scores_df = scores_df.head(inputs.top_n)
+    # set hit flag to False
+    scores_df[DockingResultCols.DOCKING_HIT.value] = False
 
-    n_top_n_filtered = len(scores_df)
+    # set top n hits to True
+    scores_df.loc[
+        scores_df.index[: inputs.top_n], DockingResultCols.DOCKING_HIT.value
+    ] = True  # noqa: E712
```
JenkeScheen, line 429: are we sure that `Postera` accepts `True`/`False`, and shows it correctly in the table view? If not we might be better off setting as `"True"` and `"False"`, similar to `NaN` (where we have to set it as `""`)

JenkeScheen (commented):
this is looking great! we should both do a live test on `lilac` before we merge this in.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
hmacdope: If a ligand is docked to more than one structure (often the case) we take the best scoring one (first in sorted df)

hmacdope (commented):
reminder to myself to apply same fix as in #667 here also.

hmacdope:
> reminder to myself to apply same fix as in #667 here also.

Fixed now.

hmacdope:
Merging here so we can do some combined debugging on main.

PR #658, opened 2023-11-28
Add a71 capsid

JenkeScheen:
Adds fast-follow EV A71 Capsid. The only non-standard thing about this one is that we need to Z-clip the HTML view because the ligand is encapsulated. Results in this view:

image

I'll let CI run and then merge myself. 


hmacdope (approved):
Amazing work @JenkeScheen

Issue #671, opened 2023-12-01
Re-add logic for truncating GIFs to be the <x> period of time and reduce GIF size. 


hmacdope:
the small-scale docking workflow v2 doesn’t have this: [https://github.com/choderalab/asapdiscovery/blob/test_full_prep_setup/asapdiscover[…]ng/asapdiscovery/docking/scripts/dock_small_scale_end_to_end.py](https://github.com/choderalab/asapdiscovery/blob/test_full_prep_setup/asapdiscovery-docking/asapdiscovery/docking/scripts/dock_small_scale_end_to_end.py#L1047-L1051) so now generated GIFs are very large, and contain whole simulation.

PR #675, opened 2023-12-04
fix alignment and long GIF

JenkeScheen:
Fixes two issues:

- ligand was not being aligned together with the protein, so was placed way outside of pocket by GIF viz. 
    -> done by jumping alignment step to just after loading trajectory instead of after application of centering etc
- hotfix for https://github.com/choderalab/asapdiscovery/issues/671, still need to make a more modular solution.
    -> done by adding a default value of `start` based on the default MD protocol


JenkeScheen:
should be ready to go - testing on `lilac` now.

hmacdope (approved):
LGTM, I think this is a good  enough fix for now.  Thanks for sorting this out.

PR #692, opened 2023-12-13
Add MultiPose Processing for Ligands

apayne97:
## Description
We'd like to have a mechanism to check if the problems with docking are a failure to pose the molecule correctly or a failure to score the docked poses well. I'm running an experiment to keep 50 docked poses for every ligand + protein pair. Currently this isn't handled well by our ligand schema (I'm going to add some tests to show how exactly).

## Todos
- [x] add a multi pose ligand to AWS testing suite
- [x] add test showing why current implementation doesn't work with this
- [x] update functions to read / write multiconformer ligands
- [x] make sure SD data is handled correctly
- [x] add roundtrip test with multiconf ligand
- [x] add `to_oemols` method
- [x] add some rdkit backend and testing
- [x] add a `.sort_confs` method
- [x] add test for multipose ligand roundtrip to json
- [x] rename sort_confs to `sort_confs_by_sd_tag_value` and make it clear in the docs that it is edited in place
- [x] combine the getter and setter openeye methods to handle both multi and single confs
- [x] combine the getter and setter rdkit methods to handle both multi and single confs
- [x] change rdkit docs to reference rdkit molecules
- [x] test rdkit set_SD_data exception
- [x] remove warnings and update docs to make the behaviour clear
- [x] add docs

## Questions
- [x] need to decide if we want Ligand to handle multiple poses natively or have explicit Ligand and MultiConformerLigand classes that only read / write those kinds of results.

## Status
- [x] Ready to go


kaminow:
I actually think we might not need a new loading function. I think if we just add the `ifs.SetConfTest` line to our current `load_openeye_sdf` function, and replace all the `OEGraphMol` with just `OEMol`, everything should actually work right. this should obviously be tested first though since I'm just basing all this off the OE docs

apayne97:
Yeah I think you're right in terms of the OE functionality, but in case anything weird happens I wanted to keep separate at first instead of breaking old things

kaminow:
fair enough. in that case I think just duplicate everything we have but as multi-conformer versions (all the functions converting between SDF/OEMol/Ligand) and maybe also create a `MultiConformerLigand` schema or something

hmacdope:
@apayne97 what was the use case of this again?

hmacdope (commented):
An option is a class that subclasses `Ligand` so its very clear when working with / expecting Multiconformer but this way also is perfectly fine? What do you think?

apayne97:
We'd like to have a mechanism to check if the problems with docking are a failure to pose the molecule correctly or a failure to score the docked poses well. I'm running an experiment to keep 50 docked poses for every ligand + protein pair. Currently   this isn't handled well by our ligand schema (I'm going to add some tests to show how exactly)

apayne97:
@hmacdope @kaminow is it possible to have an overloaded version of `from_sdf` that will detect if there are multiple poses and then return a Ligand | MultiConformerLigand depending on if there are multiple poses?

kaminow:
the way I can think to do it would be to have a `from_sdf` function somewhere that doesn't belong to a specific class, that would dispatch the correct `from_sdf` class-based function. this feels a bit weird to me though and I'm not sure where the right place for that function would be. another options is to just have one `Ligand` class that assumes multiple conformations but that works identically with one conformation if that makes sense. the final option I guess is to just trust the user to use the right class for their use-case

hmacdope:
> @hmacdope @kaminow is it possible to have an overloaded version of `from_sdf` that will detect if there are multiple poses and then return a Ligand | MultiConformerLigand depending on if there are multiple poses?


Yeah for sure you can just do this as a classmethod, its a bit non-idiomatic but I think its warranted here. 

```python 
@classmethod
def from_sdf(cls):
   ...
   if  not is_multiconf:
       return cls(**args)
   else:
        return MultiConformer(**args)
```

apayne97:
@hmacdope any objections to this option?
> another options is to just have one Ligand class that assumes multiple conformations but that works identically with one conformation if that makes sense.

hmacdope:
> @hmacdope any objections to this option?
> 
> > another options is to just have one Ligand class that assumes multiple conformations but that works identically with one conformation if that makes sense.

I am just a little tentative to this as single pose-ligand has been the paradigm we have been working with for a while. I totally agree that we need multi-confomer support, but I am just a bit worried we will break something.

 Is there a way to reverse this and think about single confomer ligand as the base case and multiple as being supported as well? Not so much a code change as a way to think about it.
 
  I guess if you can demonstrate that everything works as expected there is no problem really. Just thinking out loud here. What do you think @jthorton?

apayne97:
> Is there a way to reverse this and think about single confomer ligand as the base case and multiple as being supported as well? Not so much a code change as a way to think about it.

I agree that I'm also worried about breaking things - my goal is to keep the single-pose behavior as "first class" citizen within our schema, but bc openeye treats them differently I need to be a little clever in handling that

kaminow:
this will take some rewriting I think, but my view is that it makes more sense to (internally at least) treat every `Ligand` object as if it has multiple conformers. we can still expose methods like `Ligand.to_oemol()`, but that method will be something like `return self.to_oemols()[0]` or something like that.

this way, if you are working with single-conformer ligands, your experience won't change at all, but we'll still be able to naturally support multi-conformer ligands at the same time. I just feel like there's a clear path for assumed multi-conf -> single-conf, where as assuming a single-conf when you actually need a multi-conf will involve some panicking

apayne97:
Ok, I think I agree. I'm not sure what to do about the ligand tags though. Right now we are in the habit of generating sdf files where tags like "Docking_posit_hybrid_clash_Chemgauss4" can be different for each conformer. This makes me think that we should split tags into "has to be the same for all conformers" and "can be different for each conformer". And then I think we should make it so that those different tags can be easily accessible (and maybe even add a function to sort or pull out the conformer by a certain value). Here's some mock code:
```
$ lig = Ligand.from_sdf("multiconf.sdf"
$ lig.num_confs
50
$ lig.tags
['SMILES', 'Dataset', 'Compound_ID', 'Docking_posit_hybrid_clash_RMSD', 'Docking_posit_hybrid_clash_POSIT', 'Docking_posit_hybrid_clash_POSIT_method', 'Docking_posit_hybrid_clash_Chemgauss4', 'Docking_posit_hybrid_clash_clash']
```

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
```diff
@@ -697,7 +697,8 @@ def _generate_poses(
 
         from openff.toolkit import Molecule
 
-        core_ligand = prepared_complex.ligand.to_rdkit()
+        # make sure we are not using hs placed by prep as a reference coordinate for the generated conformers
+        core_ligand = Chem.RemoveHs(prepared_complex.ligand.to_rdkit())
```
hmacdope, line 701: Is this a seperate change?

apayne97: oops this is weird mistake, thanks!

apayne97: undone

jthorton:
@apayne97 this is looking great so far, I think the `to_rdkit` function needs to be updated and tested to ensure it supports converting to multi-conformer molecules. Are you also planning on adding the sort conformers function in this PR?

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -820,12 +845,42 @@ def _set_SD_data_repr(mol: oechem.OEMol, data: dict[str, str]) -> oechem.OEMol:
     oechem.OEMol
         OpenEye OEMol with SD data set
     """
-    # NOTE: use str to ensure re-reading the SD data will give the same value
-    mol = set_SD_data(mol, {k: str(v) for k, v in data.items()})
-    return mol
+    if isinstance(mol, oechem.OEMol):
+        if mol.NumConfs() > 1:
+            warn(f"Overwriting the following tags for all conformers: {data.keys()}")
```
hmacdope, line 850: This shoudn't warn if its expected behaviour.

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -800,15 +821,19 @@ def set_SD_data(mol: oechem.OEMol, data: dict[str, str]) -> oechem.OEMol:
...
+                f"Length of data for tag '{key}' does not match number of conformers ({mol.NumConfs()}). "
+                f"Expected {mol.NumConfs()} but got {len(value)} elements."
+            )
+        for i, conf in enumerate(mol.GetConfs()):
+            oechem.OESetSDData(conf, key, value[i])
     return mol
 
 
-def _set_SD_data_repr(mol: oechem.OEMol, data: dict[str, str]) -> oechem.OEMol:
+def set_SD_data(mol: oechem.OEMol, data: dict[str, str]) -> oechem.OEMol:
```
hmacdope, line 834: What form does the dict have to be in to sucessfully write multiconformer data? Worth explaining in docstring.

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -839,10 +894,44 @@ def get_SD_data(mol: oechem.OEMol) -> dict[str, str]:
...
+        other_sd_data = _get_SD_data(mol)
+        sd_data = {tag: value[0] for tag, value in get_multiconf_SD_data(mol).items()}
+
+        if other_sd_data:
+            warn(
+                "SD data found at the base level of a multiconformer object, which will be ignored. "
+                "Use set_SD_data to set SD data correctly."
+            )
+        if mol.NumConfs() > 1:
+            warn(
```
hmacdope, line 921: Can't we just call the function instead of warning?

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -839,10 +894,44 @@ def get_SD_data(mol: oechem.OEMol) -> dict[str, str]:
...
+                f"{mol.NumConfs()} conformers found, only returning SD data from the first conformer."
+                f"Use get_multiconf_SD_data to get data from all conformers."
+            )
+        return sd_data
+
+    elif isinstance(mol, oechem.OEGraphMol) or isinstance(mol, oechem.OEConfBase):
+        return _get_SD_data(mol)
+    else:
+        raise TypeError(
```
hmacdope, line 930: Nice.

asapdiscovery-data/asapdiscovery/data/backend/rdkit.py
```diff
@@ -1,6 +1,100 @@
...
 
 
+def set_SD_data(mol: Chem.Conformer, data: dict) -> None:
+    for key, value in data.items():
+        mol.SetProp(str(key), str(value))
+
+
+def set_multiconf_SD_data(mol: Chem.Mol, data: dict[str, list]):
+    """
+    Set the SD data on an OpenEye OEMol, overwriting any existing data with the same tag
```
hmacdope, line 15:
```suggestion
    Set the SD data on an RDKit Mol, overwriting any existing data with the same tag
```

asapdiscovery-data/asapdiscovery/data/backend/rdkit.py
```diff
@@ -1,6 +1,100 @@
...
+
+    Parameters
+    ----------
+    mol: oechem.OEMol
```
hmacdope, line 19: RDKit down here also.

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -476,6 +559,20 @@ def canonical_tautomer(self) -> "Ligand":
         else:
             raise ValueError("Unable to generate canonical tautomer")
 
+    @property
+    def num_poses(self) -> int:
```
hmacdope, line 563: Nice

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -476,6 +559,20 @@ def canonical_tautomer(self) -> "Ligand":
...
+        """
+        Get the number of poses in the ligand.
+        """
+        return self.to_oemol().NumConfs()
+
+    @property
+    def has_multiple_poses(self) -> bool:
```
hmacdope, line 570: Also nice

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -617,6 +714,34 @@ def get_chemical_relationship(self, other: "Ligand") -> ChemicalRelationship:
             relationship |= ChemicalRelationship.DISTINCT
         return relationship
 
+    def sort_confs(self, by: str, ascending: bool = True) -> np.ndarray:
```
hmacdope, line 717: This will be super useful, I would say make it clear that its by SD tag by making the name `sort_confs_by_sd_tag_value`

asapdiscovery-data/asapdiscovery/data/tests/test_rdkit.py
```diff
@@ -0,0 +1,14 @@
...
+
+def test_ligand_sdf(moonshot_sdf, multipose_ligand, sdf_file):
+    single_conf = load_sdf(moonshot_sdf)
+    assert single_conf.GetNumConformers() == 1
+    multiconf = load_sdf(multipose_ligand)
+
+    assert multiconf.GetNumConformers() == 50
+
+    # this should fail if the file has multiple ligands
+    with pytest.raises(RuntimeError):
```
hmacdope, line 13: Our openeye version doesn't fail, intead it just grabs the first one silently. Perhaps we should try and make behaviour consistent.

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -839,10 +894,44 @@ def get_SD_data(mol: oechem.OEMol) -> dict[str, str]:
...
+    from collections import defaultdict
+
+    sd_data = defaultdict(list)
+    for conf in mol.GetConfs():
+        for dp in oechem.OEGetSDDataPairs(conf):
+            sd_data[dp.GetTag()].append(dp.GetValue())
     return sd_data
 
 
+def get_SD_data(mol: oechem.OEMolBase) -> dict[str, str]:
```
hmacdope, line 906: I think it should be possible to abstract away from the user  whether we need to use a `get_multiconf` or `get_` regular. Is this the case? Same for the `set` methods. 

IE it would be great if we didn't have to worry as a user to select the correct method. Does this seem feasible.

asapdiscovery-data/asapdiscovery/data/backend/rdkit.py
```diff
@@ -1,6 +1,100 @@
...
+
+    Returns
+    -------
+    oechem.OEMol
+        OpenEye OEMol with SD data set
+    """
+    num_confs = mol.GetNumConformers()
+    for key, value in data.items():
+        if not len(value) == num_confs:
+            raise ValueError(
```
hmacdope, line 30: Test this exception.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema.py
```diff
@@ -127,6 +127,144 @@ def test_ligand_from_sdf_title_used(moonshot_sdf):
...
+    lig = Ligand.from_sdf(multipose_ligand)
+    assert lig.num_poses == 50
+
+    lig.to_sdf(tmp_path / "test.sdf")
+
+    lig2 = Ligand.from_sdf(tmp_path / "test.sdf")
+
+    assert lig2.num_poses == 50
+    assert lig2 == lig
+
```
hmacdope, line 140: Test the JSON round trip also.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema.py
```diff
@@ -127,6 +127,144 @@ def test_ligand_from_sdf_title_used(moonshot_sdf):
...
+        "Docking_posit_hybrid_clash_POSIT_method": "HYBRID",
+        "Docking_posit_hybrid_clash_Chemgauss4": "-6.079061985015869",
+        "Docking_posit_hybrid_clash_clash": "0",
+        "Dataset": "Mpro-P2005_0A",
+        "Docking_posit_hybrid_clash_POSIT": "0.18000000715255737",
+        "Compound_ID": "ADA-UCB-6c2cb422-1",
+    }
+
+    for i in range(50):
+        assert ligs[i].tags == lig.get_SD_data(i)
```
hmacdope, line 177: Nice this is clever!

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -379,16 +432,44 @@ def to_sdf(self, filename: Union[str, Path], allow_append=False) -> None:
...
+        for k, v in data.items():
             if k in self.__fields__.keys():
-                raise ValueError(f"Tag name {k} is a reserved attribute name")
-        self.tags.update(data)
+                warnings.warn(f"Tag name {k} is a reserved attribute name, skipping")
+            else:
+                new_data[k] = v
+
+        # first update the conformer tags, which is simple if the data is a dict of lists
+        if all([isinstance(v, list) for v in new_data.values()]):
```
hmacdope, line 449: I don't really understand this bit (apologies), can you clarify?

hmacdope (changes requested):
Great work! Thanks for going deep on this one. Main query is to whether we can abstract away the single or multiconf versions from the user (no stress if not technically possible).

Otherwise its looking super promising.

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
hmacdope: Goes to my other point about a method that covers both cases.

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
jthorton, line 717: This is great, maybe make it super clear that in the doc string the ligand is edited in place.

examples/testing.ipynb
jthorton, line 1: I think we can remove this now we have `sort_confs`!

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
```diff
@@ -392,6 +388,13 @@ def _generate_pose(
             return_code != oeomega.OEOmegaReturnCode_Success
         ):
             # add the failure message as an SD tag, should be able to see visually if the molecule is 2D
+
+            # because of downstream inconsistencies in the way we handle SD data, here we'll just add both to the
+            # top level and conformations. There are testing errors if we don't do both.
```
jthorton, line 393: I think this should only need to be set on the molecule SD tag.

jthorton (commented):
Great job @apayne97 this looks great and should make the conformer selection stuff in the pose generator easier which I can look at in another PR!

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
apayne97: I agree that it *should* be the case, but I get errors in `test_pose_generation` if I don't do both. I *think* it's because if you make a GraphMol from an OEMol it will lose conformation-specific SD data?

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
apayne97: If you want to set multiconf data, ie. set the POSIT probability or something for all the conformers, then you should use the `set_multiconf_SD_data` function with a `dict[str, list]`

I could change it so that if you pass `dict[str, str]` than it does that this function does and `dict[str, list]` then it does what the other function does. would that make more sense?

apayne97: Or I could make this (and the `get` methods) always use a `dist[str, list]` and for setting the data, if the list has length 1 then it sets it to all the confs, and if the length of the list matches it does what the multiconf method does (and if the lengths dont match it throws an error).

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
apayne97: this could be simplified by using a single `set_SD_data` openeye function that handles this logic, and then go through an OE object, but here i'm checking to make sure that all of the data is either a dict[str, str] or that it is all dict[str, list]. I'll try implementing this differently so that its a little clearer and assumes that the dict should always be `dict[str, list]` which if the list is len=1, it will update all confs with that value.

asapdiscovery-data/asapdiscovery/data/tests/test_rdkit.py
apayne97: I've made this an issue here #909

asapdiscovery-data/asapdiscovery/data/backend/rdkit.py
apayne97: done - and added for openeye too

apayne97:
@hmacdope @jthorton this is ready for another review! thanks!

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -115,7 +115,15 @@ class Ligand(DataModelAbstractBase):
...
-    tags: dict[str, str] = Field({}, description="Dictionary of SD tags")
+    tags: dict[str, str] = Field(
+        {},
+        description="Dictionary of SD tags. "
+        "If multiple conformers are present, these tags represent the first conformer.",
+    )
+
+    conf_tags: Optional[dict[str, list]] = Field(
+        {}, description="Dictionary of SD tags for each conformer."
+    )
```
hmacdope, line 126: There should be only one API point for getting tags IMO, otherwise error prone.

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -230,22 +251,31 @@ def to_oemol(self) -> oechem.OEMol:
...
+
+        mol = set_SD_data(mol, data)
+
         return mol
 
+    def to_single_conformers(self) -> ["Ligand"]:
+        """
+        Return a Ligand object for each conformer.
+        """
+        return [self.from_oemol(conf) for conf in self.to_oemol().GetConfs()]
```
hmacdope, line 266: Nice

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
hmacdope: Unless this is a technical impossibility, or there are serious reasons not to.

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema.py
```diff
@@ -127,6 +127,156 @@ def test_ligand_from_sdf_title_used(moonshot_sdf):
...
+
+def test_multiconf_ligand_basics(multipose_ligand):
+    lig = Ligand.from_sdf(multipose_ligand)
+    lig2 = Ligand.from_sdf(multipose_ligand)
+    assert lig.num_poses == 50
+
+    ligs = lig.to_single_conformers()
+
+    mols = [lig.to_oemol() for lig in ligs]
+
```
hmacdope, line 162: test `to_rdkit` also

asapdiscovery-data/asapdiscovery/data/tests/test_ligand_schema.py
```diff
@@ -127,6 +127,156 @@ def test_ligand_from_sdf_title_used(moonshot_sdf):
...
+    }
+
+    lig.sort_confs_by_sd_tag_value(sort_by, ascending=True)
+
+    assert lig.tags == top_tags_dict
+    assert lig_unsorted.tags != lig.tags
+    assert lig.to_oemol().GetCoords() != lig_unsorted.to_oemol().GetCoords()
+
+
+def test_multiconf_lig_to_rdkit(multipose_ligand):
```
hmacdope, line 274: Ah ok its tested here.

asapdiscovery-data/asapdiscovery/data/tests/test_rdkit.py
```diff
@@ -0,0 +1,449 @@
...
+
+    multiconf = load_sdf(multipose_ligand)
+    assert get_SD_data(multiconf) == {
+        "Compound_ID": [
+            "ADA-UCB-6c2cb422-1",
+            "ADA-UCB-6c2cb422-1",
+            "ADA-UCB-6c2cb422-1",
+            "ADA-UCB-6c2cb422-1",
+            "ADA-UCB-6c2cb422-1",
+            "ADA-UCB-6c2cb422-1",
```
hmacdope, line 31: This is enormous but I will allow it.

asapdiscovery-data/asapdiscovery/data/util/data_conversion.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_first_value_of_dict_of_lists(d: dict[str, list]) -> dict[str, str]:
+    """
+    Get the first value of a dictionary of lists
+    """
+    return {k: v[0] for k, v in d.items()}
+
+
+def get_dict_of_lists_from_dict_of_str(d: dict[str, str], length=1) -> dict[str, list]:
```
hmacdope, line 11: Can you make the docstring clearer with an example.

asapdiscovery-data/asapdiscovery/data/util/data_conversion.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    new_dict = defaultdict(list)
+    _ = [new_dict[k].append(v) for k, v in d.items() for _ in range(length)]
+    return dict(new_dict)
+
+
+def get_dict_of_lists_from_list_of_dicts(d: list[dict[str, str]]) -> dict[str, list]:
+    """
+    Convert a list of dictionaries to a dictionary of lists.
+    """
```
hmacdope, line 23: Example here too plz.

hmacdope (changes requested):
Final few iterations, mostly queries and nitpicks. I will also need @jthorton to sign off on his part, but almost there.

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
hmacdope: We can discuss. :)

asapdiscovery-data/asapdiscovery/data/tests/test_rdkit.py
apayne97: Yeah this is terrible. We should probably do an audit of the tests for simplicity and ease. But it works for now, eh?

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
apayne97: This is worth a discussion. This was a choice to avoid breaking backward compatibility.

apayne97: I think the cleanest way to do this going forward would be to replace `tags` with what `conf_tags` is.

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -309,11 +310,11 @@ def load_openeye_sdf(sdf_fn: Union[str, Path]) -> oechem.OEGraphMol:
...
         oechem.OEIFlavor_SDF_Default,
     )
+    ifs.SetConfTest(oechem.OEOmegaConfTest())
     if ifs.open(str(sdf_fn)):
-        coords_mol = oechem.OEGraphMol()
-        oechem.OEReadMolecule(ifs, coords_mol)
+        for mol in ifs.GetOEMols():
+            return mol
         ifs.close()
-        return coords_mol
```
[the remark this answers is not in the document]
apayne97: fixed here: https://github.com/choderalab/asapdiscovery/pull/692/commits/6b2b1cfd4e70fd13d2a099770c7cf91e10ca038f

asapdiscovery-data/asapdiscovery/data/backend/openeye.py
```diff
@@ -659,14 +661,14 @@ def sdf_string_to_oemol(sdf_str: str) -> oechem.OEGraphMol:
...
-    ims.openstring(sdf_str)
-    # NOTE: must use GraphMol here, not OEMol, otherwise SD data will not be read
-    mol = oechem.OEGraphMol()
-    oechem.OEReadMolecule(ims, mol)
-    return mol
+    ims.SetConfTest(oechem.OEOmegaConfTest())
+    if ims.openstring(sdf_str):
+        for mol in ims.GetOEMols():
+            return mol
+        ims.close()
```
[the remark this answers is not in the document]
apayne97: fixed here: https://github.com/choderalab/asapdiscovery/pull/692/commits/6b2b1cfd4e70fd13d2a099770c7cf91e10ca038f

asapdiscovery-data/asapdiscovery/data/util/data_conversion.py
apayne97: added here: https://github.com/choderalab/asapdiscovery/pull/692/commits/5937b698832e894ca0331be347bbd116d7187cb2

asapdiscovery-data/asapdiscovery/data/util/data_conversion.py
apayne97: added here: https://github.com/choderalab/asapdiscovery/pull/692/commits/5937b698832e894ca0331be347bbd116d7187cb2

hmacdope:
@apayne97 I would be willing to just go ahead here and kick the tag discussion down the road along with #949. You have done an excellent job here to keep backwards compatibility here without breaking anything, and you have probably suffered enough with this PR I think. Shall we discuss tommorow.

apayne97:
> @apayne97 I would be willing to just go ahead here and kick the tag discussion down the road along with https://github.com/choderalab/asapdiscovery/issues/949. You have done an excellent job here to keep backwards compatibility here without breaking anything, and you have probably suffered enough with this PR I think. Shall we discuss tommorow.

Yes, let's discuss at dev sync! I would certainly appreciate moving on lol

hmacdope:
Just running some quick integration tests now @apayne97

hmacdope:
Project support workflows seem to be working fine, but there is an alchemy test failure blocking, any ideas @jthorton

asapdiscovery-data/asapdiscovery/data/schema/ligand.py
```diff
@@ -229,22 +250,31 @@ def to_oemol(self) -> oechem.OEMol:
...
-        # dump tags as separate items
-        if self.tags is not None:
-            data.update({k: v for k, v in self.tags.items()})
-        mol = _set_SD_data_repr(mol, data)
+
+        # update conf tags to data
+        data.update(self.conf_tags)
+
+        mol = set_SD_data(mol, data)
+
```
jthorton, line 258: Should we not be setting the molecule level tags as well, it looks like they are missing which I think was causing the last testing errors?

apayne97: I think not, if we follow the OpenEye convention as per: https://docs.eyesopen.com/toolkits/python/oechemtk/oemol.html#dude-where-s-my-sd-data

> Practically, it is best to never attach SD tag data to an [OEMCMolBase](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEMCMolBase.html#OEChem::OEMCMolBase). This should only be done as a space optimization when it is assured that the multi-conformer molecule will only be written to OEB.
> 
> To this end, when an [OEMol](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEMol.html#oechem-oemol) copy constructs from an [OEGraphMol](https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEGraphMol.html#oechem-oegraphmol) the SD tag data is attached to the first conformer.

jthorton: I think my wording confused things! I mean how will any of the data in `Ligand.tags` be passed to the oemol? If you run the following on this branch you find that only conformer tags are passed on not `Ligand.tags`.
```python
from asapdiscovery.data.backend.openeye import get_SD_data
from asapdiscovery.data.schema.ligand import Ligand

mol = Ligand.from_smiles("CC", compound_name="methane")
mol.tags # should be empty
mol.tags["data"] = 1
mol.conf_tags["help"] = 1
oemol = mol.to_oemol()
get_SD_data(oemol)
```
> {'compound_name': ['methane'],
 'provenance': ['{"isomeric_smiles": "CC", "inchi": "InChI=1S/C2H6/c1-2/h1-2H3", "inchi_key": "OTMSDBZUPAUEDD-UHFFFAOYSA-N", "fixed_inchi": "InChI=1/C2H6/c1-2/h1-2H3", "fixed_inchikey": "OTMSDBZUPAUEDD-UHFFFAOYNA-N"}'],
 'data_format': ['sdf'],
 'help': ['1']}

apayne97: Ah I see what you mean! My suggestion would be to avoid this issue by using `set_SD_data` instead:
```
from asapdiscovery.data.backend.openeye import get_SD_data
from asapdiscovery.data.schema.ligand import Ligand

mol = Ligand.from_smiles("CC", compound_name="methane")
mol.tags # should be empty
mol.set_SD_data({'mydata': 1})
oemol = mol.to_oemol()
get_SD_data(oemol)
```

> {'compound_name': ['methane'],
 'provenance': ['{"isomeric_smiles": "CC", "inchi": "InChI=1S/C2H6/c1-2/h1-2H3", "inchi_key": "OTMSDBZUPAUEDD-UHFFFAOYSA-N", "fixed_inchi": "InChI=1/C2H6/c1-2/h1-2H3", "fixed_inchikey": "OTMSDBZUPAUEDD-UHFFFAOYNA-N"}'],
 'data_format': ['sdf'],
 'help': ['1'],
 'mydata': ['1']}

apayne97: if you use `set_SD_data`, it will also warn you that:
`UserWarning: Tag name data is a reserved attribute name, skipping`

jthorton: okay that was a bad example using data as a tag. What about tags passed in the init like this 
```python
from asapdiscovery.data.backend.openeye import get_SD_data
from asapdiscovery.data.schema.ligand import Ligand

mol = Ligand.from_smiles("CC", compound_name="methane", random_data="save_me")
oemol = mol.to_oemol()
get_SD_data(oemol)
```
> {'compound_name': ['methane'],
 'provenance': ['{"isomeric_smiles": "CC", "inchi": "InChI=1S/C2H6/c1-2/h1-2H3", "inchi_key": "OTMSDBZUPAUEDD-UHFFFAOYSA-N", "fixed_inchi": "InChI=1/C2H6/c1-2/h1-2H3", "fixed_inchikey": "OTMSDBZUPAUEDD-UHFFFAOYNA-N"}'],
 'data_format': ['sdf']}

apayne97: Yeah ok obviously you're right, I've added 1 line that makes that happen. We'll see if that breaks everything else:
```
from asapdiscovery.data.schema.ligand import Ligand
mol = Ligand.from_smiles("CC", compound_name="methane", random_data="save_me")
oemol = mol.to_oemol()
get_SD_data(oemol)
Out[2]: 
{'compound_name': ['methane'],
 'provenance': ['{"isomeric_smiles": "CC", "inchi": "InChI=1S/C2H6/c1-2/h1-2H3", "inchi_key": "OTMSDBZUPAUEDD-UHFFFAOYSA-N", "fixed_inchi": "InChI=1/C2H6/c1-2/h1-2H3", "fixed_inchikey": "OTMSDBZUPAUEDD-UHFFFAOYNA-N"}'],
 'data_format': ['sdf'],
 'random_data': ['save_me']}
```

hmacdope:
Ping me here if resolved @apayne97 @jthorton

apayne97:
@hmacdope we're good to go!

hmacdope:
> @hmacdope we're good to go!

Congrats and thanks for all your hard work.

hmacdope (approved):

Issue #607, opened 2023-11-09
Reduce code duplication in workflows: Construct Input schema to workflows one level higher 


hmacdope:
At dev sync today we had a very fruitful discussion with @kaminow , @apayne97  and @ijpulidos  about how to avoid the following kind of duplication where a parameter goes from:

CLI -> unpacked into Inputs -> unpacked into Workflow -> unpacked into class in Workflow

```python
# cli 
def cli(blah=x):
   inputs = Inputs(blah=blah)
   
   wf(inputs)
  
# workflow
def wf(inputs):
     # use blah
     Blah(blah=inputs.blah)
 ```
 
 Where instead the `Blah` class (as they are factories) can just be instantiated directly at the CLI layer.

```python
# cli 
def cli(blah=x):
   inputs = Inputs(blah=Blah(blah=blah) # embed instantiated model directly (may not be possible for all models)
   wf(inputs)
  
# workflow
def wf(inputs):
     # use blah
     inputs.blah() # use instantiated model
 ```

I propose we make this change after #594 and #595 as will be a bit of a change of design.

apayne97:
Does this new way of doing it make this PR obsolete? #596

hmacdope:
Yes but I think we would need to do a test run to see how easy to implement. I'll put it on roadmap?

apayne97:
How easy is it to serialize nested schema? That's one concern I had - I feel like I've struggled with that before but maybe we're doing it in a smarter way this time

kaminow:
I think pydantic should handle it natively, I don't think we need to do anything special

PR #693, opened 2023-12-13
Use the overwrite flag instead of deleting the directory by default in workflows

apayne97:
## Description
In all the workflow functions we're still using `rmtree` to delete the output directory instead of actually using the overwrite flag. This code makes sure we actually use the overwrite flag.

## Todos
- [x] update all three workflows

## Questions
- [x] the shared default is False, which will change the behavior from what you've been using. if we don't want that, we can make the defaults in the small_scale and large_scale workflows to have it be True.

## Status
- [x] Ready to go


apayne97:
@JenkeScheen @hmacdope @chrisiacovella could you sign off on this?

hmacdope (changes requested):
Looks good, but seeing as `overwrite` is default `False` and not exposed at CLI level, there is no way to change it AFAK. Would you be able to add a CLI arg in `asapdiscovery.data.cli_args` and add an endpoint in the cli itself?

apayne97:
Oops. Yes, of course!

hmacdope (approved):
LGTM!

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
```diff
@@ -87,6 +87,15 @@ def output_dir(func):
     )(func)
 
 
+def overwrite(func):
+    return click.option(
+        "--overwrite",
+        is_flag=True,
+        default=False,
```
chrisiacovella, line 94: Does setting this to False reproduce the original behavior of the workflow?

chrisiacovella: I guess I should have read the initial comment more closely.  I guess it does not, but I feel as though, in general, by default you should not overwrite, so this is likely better behavior.

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
```diff
@@ -81,9 +81,11 @@ def cross_docking_workflow(inputs: CrossDockingWorkflowInputs):
...
 
     output_dir = inputs.output_dir
-    if output_dir.exists():
+    if output_dir.exists() and inputs.overwrite:
         rmtree(output_dir)
-    output_dir.mkdir()
+    else:
+        output_dir.mkdir(exist_ok=True, parents=True)
+
     logger = FileLogger(
```
chrisiacovella, line 89: Would it be better to move the logger initialization before line 83, and log whether or not the directory has been overwritten or not? Either way, I think there should be something reported as what has actually be done.

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
```diff
@@ -81,9 +81,11 @@ def cross_docking_workflow(inputs: CrossDockingWorkflowInputs):
     """
 
     output_dir = inputs.output_dir
```
chrisiacovella, line 83: Since this bit appears in all the workflows, probably better to put this as a function in the base class and add a simple test for it.

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -120,9 +120,10 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
     """
 
     output_dir = inputs.output_dir
```
chrisiacovella, line 122: add this bit as a function in the base class?

asapdiscovery-docking/asapdiscovery/docking/workflows/large_scale_docking.py
```diff
@@ -120,9 +120,10 @@ def large_scale_docking_workflow(inputs: LargeScaleDockingInputs):
...
-    if output_dir.exists():
+    if output_dir.exists() and inputs.overwrite:
```
chrisiacovella, line 123: add this bit as a function in the base class?

asapdiscovery-docking/asapdiscovery/docking/workflows/small_scale_docking.py
```diff
@@ -160,9 +160,10 @@ def small_scale_docking_workflow(inputs: SmallScaleDockingInputs):
     """
 
     output_dir = inputs.output_dir
-    if output_dir.exists():
+    if output_dir.exists() and inputs.overwrite:
```
chrisiacovella, line 163: add this bit as a function in the base class?

chrisiacovella (changes requested):
You put the same bit of code in like 4 different places. I suggest just moving it to a function it the base class and calling that function in each spot.  Also would highly recommend adding a test to said function.

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
apayne97: Yeah that's what I'm thinking. We're still in an early enough stage of development that I don't feel too bad about changing default behavior when justified!

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
apayne97: what base class are you referring to? we haven't turned these workflows into classes yet so there wouldn't be a place to put it? Oh I guess we can do that here? https://github.com/choderalab/asapdiscovery/issues/607

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
apayne97: oh that's a good point!

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
```diff
@@ -87,6 +87,15 @@ def output_dir(func):
     )(func)
 
 
+def overwrite(func):
```
hmacdope, line 90: It I think overwrite should still be the default, as a mixed run directory is prone to unexpected results. You can use one of the `--overwrite/--no-overwrite`  mixed binary flags.

hmacdope (changes requested):

asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py
chrisiacovella: Don't these workflows all insert from `PosteraDockingWorkflowInputs` which inherits from `DockingWorkflowInputsBase`? I guess I was thinking stuck in one of those.

hmacdope: This is an ongoing discussion about possibly moving to a class based system, see #596. @chrisiacovella the `Inputs` just sanitise the various attributes rather than being used to run the workflow itself.

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
hmacdope: @apayne97 I still think overwriting should be the default, as we have been relying on this behaviour for all our uploads etc. 

There are all kinds of issues with partial writes eg a number of files that get written to a specific filename each time. eg, every time the workflow runs a bunch of CSV files are written with names like `postera_uploaded_final.csv`. If the docking workflow is run twice in the same directory this file will not reflect full state of the run. Same goes for a bunch of other files. Lots of potential for confusion.

 Unless these are read and merged or some other workaround is used I think overwrriting should be the default still. I am fine if people want to enter the danger zone themselves, but I don't think all the project support workflows should not-overwrite.

hmacdope (changes requested):
See comment on why I think overwriting should be the default.

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
apayne97: yep, this makes sense, thanks for the longer explanation. I'll make overwriting be the default!

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
hmacdope, line 94: not sure if this was ready for re-review but i'll just add here anyway with some info. Still need to change the default. You can use the `--overwrite/--no-overwrite` constructor for `click` that handles these kinds of things nicely.

eg:
```python
@click.option("--allow-dask-cuda/--no-allow-dask-cuda", default=True)
```

hmacdope (changes requested):

asapdiscovery-cli/asapdiscovery/cli/cli_args.py
apayne97: It wasn't, but thank you! I will add that

hmacdope:
@apayne97 tests are failing, can you address and then I will re-review?

apayne97:
@hmacdope ok I think i'm finally ready for a re-review

hmacdope:
@apayne97 just ping here when tested 😸

hmacdope (approved):
LGTM, thanks m80 i trust you to shepard this to completion

Issue #695, opened 2023-12-14
Prep failure is causing docking workflow failure


apayne97:
From the log file:
```
2023-12-13 12:31:55,112 | root | INFO | cross_docking.py | cross_docking_workflow | Loaded 1 query ligands
2023-12-13 12:31:55,115 | root | INFO | cross_docking.py | cross_docking_workflow | Loaded 219 complexes
2023-12-13 12:31:55,117 | root | INFO | cross_docking.py | cross_docking_workflow | Prepping complexes
2023-12-13 12:31:57,084 | asapdiscovery.modeling.protein_prep_v2 | INFO | protein_prep_v2.py | prep | Loaded 218 cached structures from: /lila/data/chodera/asap-datasets/full_frag_prepped_mpro_20230122/du_cache_p_only.
2023-12-13 12:32:00,604 | asapdiscovery.modeling.protein_prep_v2 | INFO | protein_prep_v2.py | prep | Matched 216 cached structures which will be reused.
2023-12-13 12:32:00,607 | asapdiscovery.modeling.protein_prep_v2 | INFO | protein_prep_v2.py | prep | Prepping 3 complexes
```

From the error file. I've seen the `Warning: After setting primary alternate location, ' OXT ILE 43   B 0  ' and ' N   CYS 44   B 0  ' are too close 0.43 < 0.7A. This suggests unconventional altloc use, which is invalid using the 'Primary' OEAlternateLocationOption. Try again with the 'Enumerate' option.
Warning: No ligand or sites detected, please revise inputs and/or options` before but in the current workflow this just causes the whole thing to die. I could just catch the error in the workflow.
```
Warning: No BioAssembly transforms found, using input molecule as biounit: DesignUnit Components_LIG
Warning: Iridium - Structure: DesignUnit Components_LIG has no REMARK data
Warning: After setting primary alternate location, ' OXT ILE 43   B 0  ' and ' N   CYS 44   B 0  ' are too close 0.43 < 0.7A. This suggests unconventional altloc use, which is invalid using the 'Primary' OEAlternateLocationOption. Try again with the 'Enumerate' option.
Warning: No ligand or sites detected, please revise inputs and/or options
Traceback (most recent call last):
  File "/home/paynea/miniforge3/envs/asapdiscovery/bin/asap-docking", line 8, in <module>
    sys.exit(docking())
[...]

PR #697, opened 2023-12-14
Add flag to only use cached complexes to avoid failures due to repeatedly failing to prep the structures

apayne97:
## Description
This made it possible to run the cross-docking runs to avoid the issue with https://github.com/choderalab/asapdiscovery/issues/695

## Status
- [x] Ready to go


apayne97:
@jthorton could you give this a quick review?

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
```diff
@@ -122,6 +123,12 @@ def prep(
                 inputs, cached_outputs = ProteinPrepperBase._gather_new_tasks(
                     complex_to_prep=inputs, cached_complexs=cached_complexs
                 )
+                if use_only_cache:
+                    if inputs:
+                        logger.warning(
+                            f"Disregarding {len(inputs)} structures which could not be found in the cache."
```
hmacdope, line 129: Could be misunderstanding but isn't the message the reverse? Shouldn't it be something like 

```python
f"Disregarding inputs, {len(inputs)}, only loading structures prepped in cache."
```

hmacdope (changes requested):
Looks good, just one query RE logging message.

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep_v2.py
apayne97: Here is the message in context:

```
2023-12-14 11:49:03,683 | root | INFO | cross_docking.py | cross_docking_workflow | Loaded 219 complexes
2023-12-14 11:49:03,685 | root | INFO | cross_docking.py | cross_docking_workflow | Prepping complexes
2023-12-14 11:49:06,685 | asapdiscovery.modeling.protein_prep_v2 | INFO | protein_prep_v2.py | prep | Loaded 218 cached structures from: /lila/data/chodera/asap-datasets/full_frag_prepped_mpro_20230122/du_cache_p_only.
2023-12-14 11:49:09,801 | asapdiscovery.modeling.protein_prep_v2 | WARNING | protein_prep_v2.py | prep | Disregarding 3 structures which could not be found in the cache.
2023-12-14 11:49:09,808 | asapdiscovery.modeling.protein_prep_v2 | INFO | protein_prep_v2.py | prep | Matched 216 cached structures which will be reused.
2023-12-14 11:49:09,811 | root | INFO | cross_docking.py | cross_docking_workflow | Prepped 216 complexes
```

apayne97: I think this makes sense to me?

hmacdope: Yep makes sense in context, my confusion.

hmacdope (approved):
Thanks @apayne97

jthorton (approved):
This seems like a good workaround for now, I guess we could also have an option like `ignore-fails` for when you still want to try to prep everything first but that could be added separately.

PR #703, opened 2023-12-15
Track the POSIT Method that was actually used during the docking

apayne97:
## Description
Adds back the functionality from 
https://github.com/choderalab/asapdiscovery/blob/f83d4d206dcb96fbeca385b7f17b817db9a33734/asapdiscovery-docking/asapdiscovery/docking/docking.py#L286-L288

For instance, right now if you don't give POSIT any specifics we use the "ALL" method, and then the "ALL" method is saved in the SD tags. but it's actually more helpful if we record which sub-method was actually used (FRED / HYBRID / SHAPEFIT) etc.

## Status
- [x] Ready to go


hmacdope (approved):
Yep sounds good.

PR #718, opened 2024-01-08
Add EV-D68 Capsid

### LINKED ISSUE #717: Add D68 Capsid as a target
We are now working with EV-D68 Capsid as a target which is closely related to EV-A71 capsid in structure but needs a different reference structure (PDBID: 7TAF) and pocket definition.

hmacdope:
## Description
Fixes #717 

Adds EV-D68 Capsid as a target.

## Status
- [X] Ready to go


JenkeScheen (approved):
LGTM! Do we have a 1:1 comparison of constrained vs non-constrained of this target?

hmacdope:
> LGTM! Do we have a 1:1 comparison of constrained vs non-constrained of this target?

@JenkeScheen for the MD simulations? No I haven't had a look, but we should investigate, another option is also to restrain only a selection of CAs at to protein-protein interface if we want more dynamics, which may be a good idea as some of the  sims I ran looked pretty rock solid.

Issue #727, opened 2024-01-10
ASAP-Alchemy: guardrail/ enforce setting of ligand names


JenkeScheen:
currently if no ligand names are set in the SDF inputs for `asap-alchemy submit`, this network is submitted to `alchemiscale` and after simulations and `asap-alchemy gather`, gathered edges have empty strings as names which means we can't predict on them anymore: 
```
{
        "type": "TransformationResult",
        "ligand_a": "",
        "ligand_b": "",
        "phase": "solvent",
        "estimate": {
          "magnitude": -14.6842962125111,
          "unit": "kilocalorie / mole",
          ":is_custom:": true,
          "pint_unit_registry": "openff_units"
        },
        "uncertainty": {
          "magnitude": 0.08448160502963074,
          "unit": "kilocalorie / mole",
          ":is_custom:": true,
          "pint_unit_registry": "openff_units"
        }
      }, 
```
Instead, we should be enforcing the naming of ligands before they are sent off to `alchemiscale`, or at least add a guardrail that applies some kind of ligand name (e.g. by `gufe` key).

dotsdl:
Working on this; thanks @JenkeScheen!

PR #728, opened 2024-01-10
quick guardrail that uses SMILES in case no names are set for ligands

JenkeScheen:
## Description
adds a quick post-simulation guardrail that sets ligand SMILES as names in case no names are set to ligands in an `alchemiscale` network. 


dotsdl (approved):
This is a nice backup solution for cases in which no ligand names were provided on `AlchemicalNetwork` generation. Will address upstream guardrail in #727 to avoid this codepath being taken going forward.

PR #731, opened 2024-01-10
Add installer script

apayne97:
## Description
I realize this is dumb but as a stop gap before making everything conda installable I think this makes sense to have

## Todos
- [x] fix a README typo
- [x] add bash installer script

## Status
- [x] Ready to go


devtools/repo_installer.sh
```diff
@@ -0,0 +1,12 @@
+# Edited 2024.01.10 Alex Payne
+# I can never remember the order to install the asapdiscovery packages in,
```
hmacdope, line 2: There should now be no order, (can you make a note). But happy to approve

hmacdope (approved):

devtools/repo_installer.sh
hmacdope: Ok I changed it to just not have the comment.

PR #733, opened 2024-01-10
Add unique name for DockingInputMultiStructure

### LINKED ISSUE #732: Missing API point broke cross docking workflow
This is because I didn't define `unique_name` when I made `DockingInputMultiStructure`. Not sure if we need to write more / better to catch in the future?

```
  File "/home/paynea/miniforge3/envs/asapdiscovery/bin/asap-docking", line 8, in <module>
    sys.exit(docking())
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/lila/data/chodera/paynea/asapdiscovery/asapdiscovery-docking/asapdiscovery/docking/cli.py", line 239, in cross_docking
    cross_docking_workflow(inputs)
  File "/lila/data/chodera/paynea/asapdiscovery/asapdiscovery-docking/asapdiscovery/docking/workflows/cross_docking.py", line 222, in cross_docking_workflow
    results = docker.dock(
  File "/lila/data/chodera/paynea/asapdiscovery/asapdiscovery-docking/asapdiscovery/docking/docking_v2.py", line 133, in dock
    outputs = self._dock(
  File "/lila/data/chodera/paynea/asapdiscovery/asapdiscovery-docking/asapdiscovery/docking/openeye.py", line 217, in _dock
    Path(output_dir) / set.unique_name() / "docking_result.json"
AttributeError: 'DockingInputMultiStructure' object has no attribute 'unique_name'
```

### LINKED ISSUE #759: Convert `unique_name` to be a property value in all the classes that use it
This is more stylistic than anything else (I think) but it seems weird to me that this isn't a property

apayne97:
## Description
Fixes #732 
Fixes #759 

## Todos
- [x] Add unique name for DockingInputMultiStructure
- [x] make it more useful?

## Questions
- [x] Any ideas for a better name? Or a better way to use the hashes?

## Status
- [x] Ready to go


asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
```diff
@@ -71,6 +71,9 @@ def from_pairs(
     def to_design_units(self) -> list[oechem.OEDesignUnit]:
         return [protein_complex.target.to_oedu() for protein_complex in self.complexes]
 
+    def unique_name(self):
+        return f"{self.ligand.compound_name}-{self.ligand.fixed_inchikey}_to_{len(self.complexes)}_targets"
```
hmacdope, line 75: This also needs to capture the uniqueness of the complexes as well otherwise we will get collisions. You could hash the hash of all the complexes and that should do the trick.

hmacdope (changes requested):

asapdiscovery-docking/asapdiscovery/docking/docking_v2.py
chrisiacovella: What is the unique name being used for?  That is, are you going to be using this for like a file name, or will these just show up in like a pandas data frame?  hashing the collective hashes is probably a good approach to getting a unique identifier, but that isn't very descriptive to a user, so you might want to still have some sort of descriptive string, even if not fully unique.

hmacdope: Its for a directory and potentially a cache, so needs to be unique id, but I take your point a bit ugly for a user. For the 1:1 mapping we use: 
```python
    def unique_name(self):
        return f"{self.complex.unique_name()}_{self.ligand.compound_name}-{self.ligand.fixed_inchikey}"
```
Which is semi-human readable but will become very ugly and long with multiple complexes, but perhaps there is no super neat solution.

hmacdope:
```python 
data = ""
for c in complexes:
    data += c.hash()
hashlib.sha256(data.encode()).hexdigest()
```

I think this should work @apayne97

apayne97:
@hmacdope @chrisiacovella all I did was add Hugo's code but this should be good to go now

apayne97:
@hmacdope added a bunch of changes to make the logic a little simpler, also fixed #759

apayne97:
@jthorton @hmacdope @kaminow 
so in trying to simplify some of the schema in order to make `unique_name` a little more standardized I discovered that it seems like while `Complex` has a required `ligand_chain` property, `PreppedComplex` does not. Since I just made `PreppedComplex` subclass `Complex`, now some of the tests are failing. I think we can make `ligand_chain` optional? It's mainly necessary sometimes for prep

hmacdope (approved):
LGTM! Thanks for doing this.

PR #754, opened 2024-01-19
RDKit pose generator

jthorton:
## Description
This PR adds a new ASAP-Alchemy prep constrained pose generator based on RDKit.

## Todos
- [x] Check structures
- [x] add tests

## Questions
- [ ] Question 1 ..

## Status
- [ ] Ready to go


jthorton:
@JenkeScheen this should be ready for review and testing now. The last thing is to decide before we merge is the default pose generator settings do we want to swap to rdkit or keep it as openeye?

JenkeScheen:
Nice! I'll review later today. We can keep RDKit as default behavior because it's more directed towards productization of `asap-alchemy`. Looks like CI is failing.

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
```diff
@@ -224,61 +346,98 @@ def _generate_omega_instance(
...
         # build and configure omega
         omega_generator = self._generate_omega_instance(
             core_fragment=core_fragment, use_mcs=use_mcs
         )
 
+        # run omega
+        return_code = omega_generator.Build(target_ligand)
+
+        # deal with strange hydrogen DO NOT REMOVE
+        oechem.OESuppressHydrogens(target_ligand)
```
JenkeScheen, line 386: roundtripping hydrogens like this may not be great if we're dealing with specific, pre-specified protonation states on ligands, e.g. when we (somehow) know a protonation state in the binding pocket requires it. Thoughts? Not sure how else to deal with this particular issue though.

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
```diff
@@ -293,115 +452,280 @@ def _generate_poses(
...
 
-    def _prune_clashes(self, receptor: oechem.OEMol, ligands: list[oechem.OEMol]):
+
+class RDKitConstrainedPoseGenerator(_BasicConstrainedPoseGenerator):
+    """Use RDKit to embed multiple conformers of the molecule while constraining it to the template ligand."""
+
+    type: Literal["RDKitConstrainedPoseGenerator"] = "RDKitConstrainedPoseGenerator"
+
+    max_confs: PositiveInt = Field(
+        300, description="The maximum number of conformers to try and generate."
```
JenkeScheen, line 462: is this something we could reduce by default to speed things up? where did you find 300?

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
```diff
@@ -293,115 +452,280 @@ def _generate_poses(
...
+    )
+    rms_thresh: PositiveFloat = Field(
+        0.2,
+        description="Retain only the conformations out of 'numConfs' after embedding that are at least this far apart from each other. RMSD is computed on the heavy atoms.",
+    )
+    mcs_timeout: PositiveFloat = Field(
```
JenkeScheen, line 468: I'm just checking with OpenFE folks if they have a recommended default value here.

JenkeScheen (commented):
Mostly some general comments, let's discuss those before I do a more technical dive (although at first glance this all looked very good!). I compared docking using RDKit and OE, and RDKit seems to work much better, albeit a bit slow. I had 32/32 go without fail, whereas with OE only 23/32 docked succesfully.

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
jthorton: Yeah I have been looking into this and 300 comes from this paper looking into the rdkit conformer generator https://www.rdkit.org/UGM/2012/Ebejer_20110926_RDKit_1stUGM.pdf.

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
jthorton: Good idea 1 seconds seems enough most of the time but would be good to check!

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
```diff
@@ -224,6 +235,31 @@ def to_oemol(self) -> oechem.OEMol:
         mol = _set_SD_data_repr(mol, data)
         return mol
 
+    def to_rdkit(self) -> "Chem.Mol":
```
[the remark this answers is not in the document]
jthorton: I think our ligand schema only allows a single conformer for the molecule during creation so we will always only have a single conformer when creating an rdkit molecule. You are right that this is more difficult when doing `from_rdkit` we would have to make it clear we only take the first conformer.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/prep.py
```diff
@@ -94,6 +103,7 @@ def run(
     factory_file: The name of the JSON file with the configured AlchemyPrepWorkflow, if not supplied the default will be
         used but a core smarts must be provided.
     core_smarts: The SMARTS string used to identify the atoms in each ligand to be constrained. Required if the factory file is not supplied.
+    processors: The number of processors which can be used to run the workflow in parallel.
```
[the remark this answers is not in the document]
jthorton: Yes we can get the cpu count from the multiprocessing library and have it use them (all - 1) by default. This option could be called `auto` then we can have `all` and the `N` the specific number of processes to use.

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
jthorton: Good point I did look into this and the function just turns the explicit hydrogens into implicit ones so the parent atoms track them still so the protonation state can not change. This is mainly to just recalculate the positions of the hydrogens as they often get scrambled when doing this kind of conformer generation. The function reference https://docs.eyesopen.com/toolkits/cpp/oechemtk/OEChemFunctions/OESuppressHydrogens.html#OEChem::OESuppressHydrogens

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/prep.py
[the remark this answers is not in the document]
JenkeScheen: that would be best I think!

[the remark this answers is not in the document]
jthorton: awesome added in [9d7e61b](https://github.com/choderalab/asapdiscovery/pull/754/commits/9d7e61bd9bdb8a76d6f2b1ac4674ebf47a7e0b77)

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
JenkeScheen: ok, good to stick with that for now!

asapdiscovery-docking/asapdiscovery/docking/schema/pose_generation.py
JenkeScheen: okay, I was trying to set <1 but that fails. Looking a cryptic `_pickle.PicklingError`. Looking at the RDKit docs, seems like it can only be `int`, so 1 is the minimum I think. I remember peers using `timeout = 1` now that I think about it some more, so let's just stick with that. Would be good to at least set `PositiveFloat` -> `PositiveInt`?

jthorton: Good spot Ill update that.

apayne97:
> I compared docking using RDKit and OE, and RDKit seems to work much better, albeit a bit slow. I had 32/32 go without fail, whereas with OE only 23/32 docked succesfully.

crying

JenkeScheen:
> crying

FWIW, this is mostly dependent on the quality of the MCS match, not the docking itself.

jthorton:
@JenkeScheen I think we are ready for a full review now!

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/test_alchemy_prep.py
```diff
@@ -25,9 +28,12 @@
 def test_prep_workflow(strict_stereo, core_smarts, failed, mac1_complex):
     """Make sure the full prep workflow can be run and stereo issues can be filtered."""
 
-    # we do not have access to epik in testing so skip
+    # we do not have access to epik in testing so skip, use openeye as its faster
```
JenkeScheen, line 31:
```suggestion
    # we do not have access to epik in testing so skip, use openeye as it's faster
```

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/test_cli.py
```diff
@@ -129,18 +129,24 @@ def test_alchemy_prep_run_with_fails(tmpdir, mac1_complex):
...
+                "-f",
+                "openeye_workflow.json",
                 "-n",
                 "mac1-testing",
                 "-l",
                 ligand_file.as_posix(),
                 "-r",
                 "complex.json",
+                "-p",
+                1,
```
JenkeScheen, line 149: is the test system single core? or can we speed up the test by setting to `auto`?

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
JenkeScheen, line 238: are we also copying over the ligand name here? we need to be careful to conserve this both ways as it's become clear that not having names associated with ligands is problematic downstream

JenkeScheen (approved):
Awesome work, looks like this really strengthens the robustness of the pipeline! Just some minor comments, feel free to merge once addressed.

asapdiscovery-data/asapdiscovery/data/schema_v2/ligand.py
jthorton: Yes we copy it over as an SD tag called `compound_name`, but we can add it as the `_Name` property as well.

PR #782, opened 2024-01-29
Make `Alchemy predict` skip broken edges

JenkeScheen:
Just some quick changes to make `predict` not error out if an edge hasn't finished, but instead warn the user and continue. It looks like `OpenFE` automatically removes nodes from the network which aren't connected to the network. This enables two things:

- can compute free energies for a network at any given time while simulations are running (although early on this would result in all/most nodes being ignored)
- for larger networks, we won't have to endlessly `restart` tasks to make every edge succeed. If we run 1000 ligands, it's not a big deal to remove a handful of ligands. 


jthorton (approved):
LGTM!

Issue #788, opened 2024-01-29
Port VisNet from MTENN


fyng:
Port in [ViSNet](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.models.ViSNet.html?highlight=visnet) model from MTENN. ViSNet is an equivariant message passing neural net for molecular property prediction.

Requires PyG >= 2.5.0

hmacdope:
Fixed by #789

PR #789, opened 2024-01-29
Add ml visnet

fyng:
## Description
Port in VisNet from MTENN. Requires PyG >=2.5.0 (currently PyG-nightly)
#788 

## Status
- [ ] Ready


fyng:
ml-ci needs updated MTENN to pass. Current MTENN version does not have valid reference to `ViSNetModelConfig`

fyng:
Setting `--num-layers` from CLI returns command not found error

.github/workflows/ml.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, ubuntu-nightly]
```
hmacdope, line 33: I see how you got to this solution, because the conda-envs are called from the os matrix. However that was just to avoid me typing it all out. 

This won't work because os is a reserved keyword for the runner type. You need something like the below

```yaml
matrix:
  os: [ubuntu-latest]
  conda-env: [ubuntu-latest, ubuntu-latest-pyg-nightly]


    - name: Setup Conda Environment
      uses: mamba-org/setup-micromamba@v1
      with:
        environment-file: devtools/conda-envs/asapdiscovery-${{ matrix.conda-env }}.yml
        environment-name: asapdiscovery
        cache-environment: true
        cache-downloads: true
        cache-environment-key: environment-${{ steps.date.outputs.date }}
        cache-downloads-key: downloads-${{ steps.date.outputs.date }}
        create-args: >-
          python==${{ matrix.python-version }}

```

I wouldn't bother with a macos nightly build too painful.

fyng: Thanks!

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -22,7 +22,11 @@
     ModelConfigBase,
     ModelType,
     SchNetModelConfig,
+    ViSNetModelConfig,
 )
+
+# guard for visnet import. Refer MTENN issue #42
+from mtenn.conversion_utils.visnet import HAS_VISNET
 from pydantic import BaseModel, Extra, Field, ValidationError, validator
```
fyng, line 30: The goal here is to give some feedback for users who installed `mtenn` from conda, without preventing them from using the other models. Appreciate advice on a better way to go about that

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -23,6 +28,20 @@
...
+
+    # guard for visnet import. Refer MTENN issue #42
+    from mtenn.conversion_utils.visnet import HAS_VISNET
+except ImportError as e:
+    match = re.search(r"cannot import name 'ViSNetModelConfig'*", str(e))
+    if match:
+        warnings.warn("To use VisNet, pip install MTENN from Github", ImportWarning)
+        HAS_VISNET = False
+    else:
+        raise ImportError(f"Unexpected ImportError: {e}")
```
fyng, line 43: With our current setup, we are testing against the official MTENN release (no `ViSNetModelConfig` implementation), and the github MTENN with `ViSNetModelConfig`. Ideally we want to elegantly fail on VisNet import and let the users of official MTENN carry on with a warning

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -187,6 +217,8 @@ def load_cache_files(cls, config_kwargs, field):
                     config_cls = SchNetModelConfig
                 case ModelType.e3nn:
                     config_cls = E3NNModelConfig
+                case ModelType.visnet:
+                    config_cls = ViSNetModelConfig
```
fyng, line 221: However in several places we explicitly references `ViSNetModelConfig`. This needs to be taken out and addressed differently in order for `asap-ml` to work with both `mtenn` release version and Github version

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -149,6 +168,17 @@ def output_dir_check(cls, p):
...
+    @validator("model_config")
+    def check_model_type_visnet_import(cls, v):
+        # VisNet requires PyG >=2.5.0. Currently only in PyG-nightly
+        # Refer MTENN issue #42
+        if not HAS_VISNET and isinstance(v, ViSNetModelConfig):
+            raise ImportError(
+                "Can't import ViSNetModelConfig without mtenn.conversion_utils.visnet."
+            )
+        else:
+            return v
```
fyng, line 180: This validator also explicitly references `ViSNetModelConfig`

fyng: @hmacdope any advice will be hugely appreciated!

hmacdope:
@kaminow I think we can get rid of a few of these issue by just doing a quick `mtenn` release.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: This shouldn't actually execute because HAS_VISNET is false right?

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: Just need to make sure these branches do not execute with HAS_VISNET guard.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
fyng: the reference to ViSNetModelConfig will still throw a AttributeError

kaminow:
@hmacdope @fyng I just cut a 0.5.1 release for `mtenn` and added the pin to the  `nightly` env here, so I think that should take care of some of these issues

kaminow:
@fyng `mtenn` 0.5.1 has been merged into `conda-forge`, so we should be good for that here (maybe wait a couple minutes for everything to sync). I think if you change the pins for all the env files then it's safe to reference the `ViSNetModelConfig` without any guards

hmacdope:
@fyng @kaminow the conda-forge deployment failed previously due to some runner issue, now fixed so 0.5.1 should be deployed now so good to try again here.

fyng:
[PyTorch-Geometric 2.5.0](https://github.com/pyg-team/pytorch_geometric/releases) released 4 days ago, I'll update the environment requirements to match the conda release instead.

hmacdope:
@fyng you can get rid of the nightly runner now, sorry for the mess around, wasn't expecting the release so soon.

hmacdope:
@fyng @kaminow, the CLI training tests are chewing up 50 Gb+ memory, I have added a swapfile in the runner but we should possible try and use mocking instead to avoid the expensive calls.

fyng:
@hmacdope @kaminow I think we are close to the finish line!

.github/workflows/ml.yaml
```diff
@@ -87,11 +87,12 @@ jobs:
       run: |
         # run each package test suite; append to coverage file
         # Exit immediately if a command exits with a non-zero status.
-        set -e
+        # set -e
```
kaminow, line 90: should this be uncommented?

.gitignore
```diff
@@ -123,6 +123,14 @@ cluster_scripts/log_files
...
 *.out
 *-log.txt
+model_training/*
+wandb/*
+# run scripts
+run_*.sh
+
+# configs and data structures
+*_config.json
+*_ds.pkl
```
kaminow, line 133: not sure that these need to be in the repo `.gitignore` file, but I guess there isn't really any harm in having extra stuff here (unless there is @hmacdope?)

asapdiscovery-ml/asapdiscovery/ml/cli.py
```diff
@@ -788,6 +789,258 @@ def build_e3nn(
...
+            "derivative": derivative,
+        }
+        if (es_config_cache and es_config_cache.exists()) or es_type:
+            es_config = {
+                "cache": es_config_cache,
+                "overwrite_cache": overwrite_es_config_cache,
+                "es_type": es_type,
+                "es_patience": es_patience,
+                "es_n_check": es_n_check,
+                "es_divergence": es_divergence,
```
kaminow, line 956: adjust these as in the methods for the other models (also see my note in `config.py`)

asapdiscovery-ml/asapdiscovery/ml/config.py
```diff
@@ -138,22 +138,22 @@ class EarlyStoppingConfig(BaseModel):
         ),
     )
     # Parameters for best
-    patience: int = Field(
+    es_patience: int = Field(
```
kaminow, line 141: I'd rather leave these names as they are, since I think it makes more sense to not have the `es_` prefix when we're already in the `EarlyStoppingConfig`. I think my preferred solution would just be to change the `es_config` dicts in `cli.py` to match the ones in the methods for the other models (see my note in  `cli.py` as well)

kaminow (changes requested):
I think this mostly looks good! a couple questions and some small changes

.gitignore
hmacdope: No harm, but you shouldn't be in the habit of generating anything inside the source tree anyway.

.github/workflows/ml.yaml
fyng: woops, forgot to undo that when I was poking around

asapdiscovery-ml/asapdiscovery/ml/config.py
fyng: Should we change `es_type` to `type` for consistency? It was `es_type` before I edited

kaminow: I think leave the `es_type` as is, since that's what we have in the other `Config` classes. even though it's technically fine, it feels weird to me to have a field that has a protected name (`type`) which is why I like the `..._type` pattern

.gitignore
fyng: stripped out the generated file strings

hmacdope (approved):
All looks good to me, great work @fyng

kaminow (approved):
looks good, thanks for bearing with me and my nitpicks!

Issue #769, opened 2024-01-25
Move `fitness` from `data` to its own subpackage


hmacdope:
We should move `fitness` from `data` to its own subpackage.

PR #807, opened 2024-01-31
Adding code for sequence alignment

mariacm12:
## Description
Created a new folder for the project _Automated sequence/structure alignment_. The goal is to build an an automated software pipeline for cross-target sequence and structural alignment.  
The folder includes a test case where the protein sequence of MERS-CoV Mpro and SARS-CoV-2 Mpro are aligned and the result is saved as a PNG file. The Jupyter notebook contains the solution of the test case on three ways:
- Using the Bokeh, Panel and BioPython packages, which requires the installation of panel and bokeh.
- A solution using Matplotlib and BioPython. Requires installation of Matplotlib.
- A solution using Matplotlib only.  
I'm not sure which solution is better so I included all in the notebook.

**No need to merge to the main yet, this is juts a test.**

## Todos
- [ ] Solving the problem of sequence alignment.
- [ ] This is not a full package yet

## Questions
- [ ] Question 1 ..

## Status
- [ ] Not ready to go

- [ ] I certify that this contribution is covered by the MIT license as defined in our [LICENSE](https://github.com/choderalab/asapdiscovery/blob/main/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).


hmacdope:
@mariacm12 awesome start! I can build out package structure around what has been done so far, 

@JenkeScheen scope question do we want this to be part of a combined `seq/fitness` subpackage? Or do we want them to be separate? We could combine this with the start of  #769 if we want them combined.

JenkeScheen:
> @JenkeScheen scope question do we want this to be part of a combined seq/fitness subpackage? Or do we want them to be separate? We could combine this with the start of https://github.com/choderalab/asapdiscovery/issues/769 if we want them combined.

Yes, let's make this part of `fitness`. We can keep this PR as a sandbox for now while we build `fitness` separately, then merge some of this science into `fitness` once it's ready to be shipped.

hmacdope:
@mariacm12 I think there is nothing useful left in this PR? I will close, but feel free to re-open if not the case.

PR #811, opened 2024-02-01
Add MTENN version check

### LINKED ISSUE #532: Add more dramatic stop if using version of mtenn incompatible with ML model
We should throw some kind of exception rather than allowing continuation with rubbish numbers.

kaminow:
`mtenn` models will only work with certain versions of the package, so add pins in `asapdiscovery.ml.models.MLModelBase` to check and make sure versions are compatible. Fixes #532.


kaminow:
@hmacdope I think this is ready for you to take a look whenever you get a chance. not blocking anything currently but should probably be merged as soon as is convenient just to avoid shooting ourselves in the foot later at some point

hmacdope (approved):
Awesome work!

PR #831, opened 2024-02-06
Fix propogation of timing info to GIFVisualizer

### LINKED ISSUE #766: Fix GIF timestamps when full simulation is not run.
The GIF timestamps are not correct if you don't run for 10 ns, we should fix this by adding in the option to specify total time or parameters to reconstruct it to the class.

hmacdope:
## Description
Fixes #766 

### DIFF
--- a/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
+++ b/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
@@ -74,8 +74,7 @@ class SimulatorBase(BaseModel):
     debug: bool = Field(False, description="Debug mode of the simulation")
 
     @abc.abstractmethod
-    def _simulate(self) -> list["SimulationResult"]:
-        ...
+    def _simulate(self) -> list["SimulationResult"]: ...
 
     def simulate(
         self,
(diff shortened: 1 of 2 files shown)


hmacdope:
@JenkeScheen @jthorton could I snag a review here please? :+1:

asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
```diff
@@ -235,6 +235,20 @@ def check_steps(cls, values):
             )
         return values
 
+    @property
+    def n_frames(self) -> int:
+        return self.num_steps // self.reporting_interval
+
+    @property
+    def total_simulation_time(self) -> openmm.unit.quantity.Quantity:
+        return self.num_steps * self.timestep * unit.femtoseconds
```
jthorton, line 244: Would be great to have units but until then we should update the field strings on the `VanillaMDSimulator` class to detail the intended units.

jthorton (dismissed):
This looks great, would be good to detail the intended units of the inputs though till we use units in the schema.

hmacdope:
Thanks for the review @jthorton. Could I snag a re-review?

kaminow (approved):
looks good!

PR #835, opened 2024-02-07
Add sars cov 2 n protein

JenkeScheen:
Adds a new target (`SARS-CoV-2 N protein`) and enables it in fitness view too. Notes on implementing the fitness view:
- there wasn't any need for alignment, presumably because the base gene in phylo corresponds to the one used in crystallography
- because `N protein` has chains A and C, and because they overlap/ clasp into each other around the area where ligands are bound we need to show both chains' fitness. enabled this by just duplicating the fitness data. 


hmacdope (approved):
LGTM!

PR #844, opened 2024-02-12
Fix issues around passing of loglevels in dask code and improve testing

### LINKED ISSUE #843: Type enforcement is causing an issue in log level parsing
Sometimes strings such as "INFO" are passed to loglevels rather than numbers, we need to be able to support both.

hmacdope:
Fixes #843 


hmacdope (commented):
This has an urgent patch for something broken on `main` so I am just gonna merge.

Issue #720, opened 2024-01-09
adjust weight for running networks in `alchemiscale`


JenkeScheen:
Dotson:

> I'm cutting a 0.2.2 release of Alchemiscale that features the ability to set these things client-side, so after today (if deployment goes well) you should be able to set relative weights and priorities for your AlchemicalNetworks and Tasks however you like, whenever you like.

This means we can make an `asap-alchemy prioritize` which will bump the selected (running) network to the highest priority.

PR #846, opened 2024-02-13
Alchemy submit tweaks

JenkeScheen:
## Description
Add the following features:
1. enforce setting `-p` to `public` or `confidential` on `asap-alchemy submit` when submitting under org `asap`, which will help `alchemiscale` decide what hardware to run tasks on. 

2. `submit` with priority by setting `prioritize` (bool). If `True`, the submitted network weight will be set to be 0.01 higher than the highest weight associated with a network in `alchemiscale`. Fixes https://github.com/choderalab/asapdiscovery/issues/720

3. add `asap-alchemy stop` CLI command that allows users to `stop` a network that has running/waiting tasks by querying by `network_key`. 
        -> caveat: this currently needs clarification by the `alchemiscale` team because it seems we're not able to set task status to `error`, but only to `deleted` which means we can't revive the tasks. see https://github.com/openforcefield/alchemiscale/issues/236


JenkeScheen:
@jthorton can you check where this should be tested / write some tests + give a review?

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
```diff
@@ -103,8 +136,16 @@ def action_network(
                 self._client.create_tasks(tf_sk, count=planned_network.n_repeats + 1)
             )
 
+        # set the network weight based on the found network weights that are currently running while making sure the weight doesn't fall outside 0-1
+        if prioritize is None:
+            weight = 0.5
+        elif prioritize:
```
hmacdope, line 142: This will accepty any `truthy` value eg "a", may want to check `if prioritize is True` 

In fact I might restructure this a bit so that the branches are 


```python

if prioritize is True: 
    ... # prioritize
elif prioritize is False:
    ... #deprioritize 
 else:
     # use 0.5 
 ```
 
 That way the base case is not de-prioritizing.

File: asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py [hunk 2]
```diff
@@ -81,14 +82,46 @@ def create_network(
...
+            if finished_counter == 5:
+                break
+
+            active_network_weights.append(self._client.get_network_weight(network=key))
+
+        return active_network_weights
+
     def action_network(
-        self, planned_network: FreeEnergyCalculationNetwork
+        self, planned_network: FreeEnergyCalculationNetwork, prioritize: Optional[bool]
```
hmacdope, line 117:
```suggestion
        self, planned_network: FreeEnergyCalculationNetwork, prioritize: Optional[bool]=None
```
with current setup need to specify None if we want default case to be  weight=0.5 (default). See other comment as well though.

hmacdope (changes requested):
Looking great, some super quick comments.

asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py
jthorton: Would it make our life easier to just use priority categories like `normal=0.5`, `medium=0.75` and `high=1`?  It would save us from having to get all other actioned weights when setting the priority as well which should make it faster. We can expose a way to change the priority of a network via the CLI as well.

jthorton: Would it make our life easier to just use priority categories like `normal=0.5`, `medium=0.75` and `high=1`?  It would save us from having to get all other actioned weights when setting the priority as well which should make it faster. We can expose a way to change the priority of a network via the CLI as well.

JenkeScheen: @jthorton that could still quite easily lead to a situation where we have a network running at `1`, which would block us from submitting any networks of even higher priority. Having it bump by small increments like in the above would essentially prevent that situation from ever happening

jthorton: Good point I have kept the current method and changed it slightly to ensure no networks get a weight of 0.0 as this stops all compute on them.

jthorton (approved):
Great job @JenkeScheen this should be ready to go once tests pass!

hmacdope (approved):

PR #883, opened 2024-02-27
Cast UUIDs to str rather than UUID in identifiers

hmacdope:
## Description
UUIDs cannot be serialised in Alchemy schema, so we should downcast to str


jthorton:
Nice workaround the other option is that we could extend the [custom json encoder/decoder](https://github.com/OpenFreeEnergy/gufe/blob/544262ca31f72ebd67b8ba6b411a57abdc405dff/gufe/tokenization.py#L29-L39) to accept this type if its important that we have a uuid?

hmacdope:
Good idea, I might try and push upstream

PR #887, opened 2024-02-27
Add filter in split_fragment_screen.py

SiyangPeng:
## Description
Recently found that there are multiple ligands that are not able to be recombined with the original protein. There were no error messages but the combined file of both the ligand and the protein only contains the protein. Adding a function that would run after the split to try to catch the ligands that were not able to be combined with the original protein and print out error messages if that is the case so could determine what to do with the protein and the ligand better. 

## Status
- [ ] Ready to go


SiyangPeng:
Hello, some of the split ligands are having some trouble being put back into the protein, but I do not currently know how to fix that. So I am putting on a quick bandaid of printing out the files with fragments that are having the issues so could back-track and find those fragments to see what is going on.

SiyangPeng:
> @SiyangPeng the diff here includes 150+ test and work files along with the changes to your proposed code. You need to unstage / delete these from the diff.
> 
> In general it is best to **never** work inside the source tree of the repo itself, instead make a folder outside of the `asapdiscovery` repo and work from there. This is to clearly separate work from code and avoid a situation like this. This is fine, but can be very bad if working with confidential data.

Sorry about that. I will delete those and get a better work station.

hmacdope:
> > @SiyangPeng the diff here includes 150+ test and work files along with the changes to your proposed code. You need to unstage / delete these from the diff.
> 
> > 
> 
> > In general it is best to **never** work inside the source tree of the repo itself, instead make a folder outside of the `asapdiscovery` repo and work from there. This is to clearly separate work from code and avoid a situation like this. This is fine, but can be very bad if working with confidential data.
> 
> 
> 
> Sorry about that. I will delete those and get a better work station. 

All good, it's no stress 😎

asapdiscovery-data/asapdiscovery/data/scripts/split_fragment_screen.py
```diff
@@ -142,16 +144,34 @@ def split_ligands(directory, input_file, structure_name):
         output_file = str(lig_pdb_file.stem) + "_protein.pdb"
         save_openeye_pdb(lig_protein, output_path / output_file)
         # Delete the ligand pdb since no longer useful
-        os.remove(lig_pdb_file)
+        # os.remove(lig_pdb_file)
```
apayne97, line 147: You could make this a flag (--save-intermediates). or uncomment.

apayne97 (changes requested):
Minor edit, also @hmacdope is it overkill to add the zika fragment structure to AWS and add a test for this script? I think it would make sense to do that if we actually think we might use this again

apayne97 (commented):
I think lines 117-123 is where you can use `Ligand.from_oemol()`, keep only non-redundant ligands, and then write out separate sdf files

apayne97 (approved):
LGTM!

asapdiscovery-data/asapdiscovery/data/scripts/split_fragment_screen.py
hmacdope: Happy to push testing to another PR, lets go ahead here in meantime?

hmacdope (approved):

PR #889, opened 2024-02-28
ASAP-Alchemy: Prep updates

jthorton:
## Description
Minor changes to the rdkit pose generation to improve reliability, we now remove the hydrogens from the reference compound when generating the alchemy poses. We can also now use the `dataset_name` from prep during the planning stage.

## Todos
- [x] allow dataset name to be picked up from prep
- [x] remove hydrogens from reference pose when using rdkit pose generator

## Status
- [ ] Ready to go


jthorton:
Yes it was matching the hydrogens which was fine for the MCS but we get strange geometries where an r-group atom is constrained to the hydrogen position.

PR #898, opened 2024-02-29
Fix multi-pose model training.

### LINKED ISSUE #882: New ML workflow doesn't work as-is for multi-pose models
There is some code in the new ML setup that is no longer compatible between single- and multi-pose models. One such example is accessing compound experimental data (eg https://github.com/choderalab/asapdiscovery/blob/main/asapdiscovery-ml/asapdiscovery/ml/trainer.py#L539-L541).

kaminow:
Fix issues in new ML pipeline that weren't compatible with multipose models. Closes #882.


hmacdope (approved):
LGTM

PR #908, opened 2024-03-15
Implement PatientConvergedEarlyStopping.

kaminow:
Implement new early stopping method (`patient_converged`) that checks for convergence and then continues with the training loop to ensure that the model has actually converged and not just reached a local minimum.

### DIFF
--- a/asapdiscovery-ml/asapdiscovery/ml/config.py
+++ b/asapdiscovery-ml/asapdiscovery/ml/config.py
@@ -170,23 +170,22 @@ class EarlyStoppingConfig(BaseModel):
     def check_args(cls, values):
         match values["es_type"]:
             case EarlyStoppingType.best:
-                assert (
-                    values["patience"] is not None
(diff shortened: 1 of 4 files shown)


asapdiscovery-ml/asapdiscovery/ml/config.py
```diff
@@ -175,17 +180,30 @@ def check_args(cls, values):
                     "Values required for n_check and divergence when using "
                     "ConvergedEarlyStopping."
                 )
+            case EarlyStoppingType.patient_converged:
+                assert (values["n_check"] is not None) and (
```
hmacdope, line 184: ValueError rather than assert.

hmacdope (approved):
LGTM, just one nitpick

kaminow:
@hmacdope I'm pretty sure not, but is this failing test something I broke? or can we just push ahead?

hmacdope:
Ill fix the test #910

PR #915, opened 2024-03-20
Add handling of extra_config option in Trainer

### LINKED ISSUE #753: Add handling of extra_config option in Trainer
I think I forgot to add this when I was implementing the `Trainer` class, but  there is currently no parsing the <key>,<value> pair strings being passed to `extra_configs`. Should be a quick fix of just adding a `validator` for that `Field`.

kaminow:
Add handling of `extra_config` in `Trainer` so that a list of key:value pair strings can be passed from the CLI. Closes #753.


kaminow:
@hmacdope quick question: do you think it makes sense to have this logic in `Trainer`, as I have it now, or should I push it up to the CLI level?

hmacdope (approved):
LGTM!

hmacdope:
@kaminow logic seems fine as you have check for `isinstance(dict)`

PR #924, opened 2024-03-27
[WIP] Add W&B sweep functionality

kaminow:
W&B sweeps got lost in the transition to the new `Trainer` setup, so need to add it back in.


asapdiscovery-ml/asapdiscovery/ml/config.py
```diff
@@ -21,6 +23,29 @@
 from pydantic import BaseModel, Field, root_validator
 
 
+class ConfigBase(BaseModel):
+    """
+    Base class to provide update functionality.
+    """
+
+    def update(self, config_updates={}) -> ConfigBase:
```
hmacdope, line 31: Misleading name as it doesn't update in place no? Returns a new updated object as far as I can see?

hmacdope: Im not exactly sure what I would replace it with though, ignore if you can't think of something else.

hmacdope (approved):
Only one quibble but non blocking. I wonder if there is a way to avoid the repetition of so many parameters (problem elsewhere also),  but overall but nothing springs to my mind.

asapdiscovery-ml/asapdiscovery/ml/config.py
kaminow: you definitely have a point, I was stuck on a bug for a while because I forgot that it wasn't an in-place update and assumed it was. I'll think on a replacement (or maybe just make it in place?)

kaminow:
also, which parameters are you referring to:

>  I wonder if there is a way to avoid the repetition of so many parameters (problem elsewhere also), but overall but nothing springs to my mind.

asapdiscovery-ml/asapdiscovery/ml/config.py
kaminow: ok @hmacdope I think I'm going to leave it as is for now since it matches how I have it in `mtenn`, but I'll make an issue to change this in both repos. I think `new_from_update` instead of just `update` might make sense as the new name, but let me know what you  think

hmacdope: Sounds good.

hmacdope:
> also, which parameters are you referring to:
> 
> > I wonder if there is a way to avoid the repetition of so many parameters (problem elsewhere also), but overall but nothing springs to my mind.

The big dicts that are floating around for each of the CLI opts, but we can think on it #943

PR #932, opened 2024-03-29
Fix batch training math

kaminow:
Realized that I left out a scaling term in the loss when using batches.

### DIFF
--- a/asapdiscovery-ml/asapdiscovery/ml/trainer.py
+++ b/asapdiscovery-ml/asapdiscovery/ml/trainer.py
@@ -76,7 +76,11 @@ class Trainer(BaseModel):
         ),
     )
     batch_size: int = Field(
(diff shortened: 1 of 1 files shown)


hmacdope (approved):

kaminow:
@hmacdope I've been seeing this issue with the `docking` CI failing pretty frequently now. it always seems to be the same thing of different tests being collected. do you know what that is/what's causing it? it seems to go away if I re-run the CI workflow

hmacdope:
@kaminow, not really I need to do some digging. Lets move to an issue as I agree annoying to always be re-running.

kaminow:
gotcha. is there a way to push these PRs through without that check?

hmacdope:
Lets try again otherwise can just force merge.

PR #933, opened 2024-03-29
Rework Trainer building in ML CLI

kaminow:
There were a couple of things that didn't work the way I wanted them to in the ML CLI that are fixed in this PR:
* Passing a `Trainer` config file ignore all other CLI args -> args from  config file are loaded, and then any CLI args that are specified override those in the file
* Continuing an existing run didn't do anything -> now works as expected


asapdiscovery-ml/asapdiscovery/ml/cli.py
```diff
@@ -1392,151 +1462,165 @@ def build_and_train_schnet(
...
+                    {
+                        k: v
+                        for k, v in trainer_kwargs[config_name].items()
+                        if v is not None
+                    }
+                )
+            else:
+                config_trainer_kwargs[config_name] = trainer_kwargs[config_name]
+
+        trainer_kwargs = config_trainer_kwargs
```
hmacdope, line 1600: This has been repeated a lot, you can just make it a function that lives in this module so you only have to update one spot

asapdiscovery-ml/asapdiscovery/ml/cli.py
```diff
@@ -1392,151 +1462,165 @@ def build_and_train_schnet(
...
-        # Save Trainer
-        if trainer_config_cache and (
-            (not trainer_config_cache.exists()) or overwrite_trainer_config_cache
-        ):
-            trainer_config_cache.write_text(t.json())
+    # Save Trainer
+    if trainer_config_cache and (
+        (not trainer_config_cache.exists()) or overwrite_trainer_config_cache
+    ):
+        trainer_config_cache.write_text(t.json())
```
hmacdope, line 1623: Same here, just mash into a function that handles this (if possible)

hmacdope (changes requested):
Just make some utility functions to avoid some repetition, other than that looks good.

kaminow:
@hmacdope good call, should be good to go now

hmacdope (approved):
LGTM, thanks for making the change.

hmacdope:
Ah @kaminow will need to fix conflicts

kaminow:
oop yeah I will fix

PR #919, opened 2024-03-22
`ASAP-Alchemy`: Inject experimental ligands in to networks

jthorton:
## Description
This PR extends the `ASAP-Alchemy` prep workflow to automatically add experimental ligands into a network from the provided CDD experimental protocol.

closes #897 

## Workflow:
- Download all ligands with valid experimental data in the given CCD protocol
- Extract all ligands with fully defined stereochemistry
- Sort the ligands by MCS overlap with the crystal ligand (to ensure a valid pose can be generated)
- Remove any experimental ligands which are in the target list (ensure no self-transforms)
- Iteratively generate poses for the experimental ligands until we have the user-provided N (default =3) experimental ligands posed in the dataset
- Mark the ligands as experimental to ensure experimental edges are added during the prediction stage to strengthen the network. 

## Todos
Notable points that this PR has either accomplished or will accomplish.
- [x] Inject experimental ligands
- [ ] identify experimental ligands during prediction and add edges, this is now being moved to another PR
- [x] Add tests of full workflow

## Questions
- [ ] Question 1 ..

## Status
- [x] Ready to go

## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT license as defined in our [LICENSE](https://github.com/choderalab/asapdiscovery/blob/main/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).

Issue #918, opened 2024-03-22
`asap-alchemy` find best crystal reference for a folder with PDB files


JenkeScheen:
a current bottleneck in the pipeline is finding a good crystal reference. Trawling the google drive is time-consuming mostly because there's no easy way of seeing the ligand's structure there.

we need to build some code using @apayne97's fragalysis code (that matches by MCS) to do this automatedly instead; for now finding the reference in a folder should be good enough (can just keep it local and download periodically from google drive).

jthorton:
Good idea, I guess we would need to find the crystal reference that has the largest MCS match to all of the ligands we want to run on assuming we are using a single reference for the whole set.

JenkeScheen:
yep, code should be pretty much there already I think. I'll make a first pass to see where it gets us

hmacdope:
This should help yall get started, picks the best complex for each ligand.

```python 
from asapdiscovery.data.operators.selectors.mcs_selector import MCSSelector
from asapdiscovery.data.readers.structure_dir import StructureDirFactory

ligands = ...
structure_factory = StructureDirFactory.from_dir(self.structure_dir)
complexes = structure_factory.load( )
selector = MCSSelector()
    pairs = selector.select(
        ligands,
        complexes,
        n_select=1,
    )

hmacdope:
Although thinking about this more this is a 1:Many comparison, not a 1:1 comparison, so you probably want to compute the ligand that has the minimum distance to all other ligands in MCS space, rather than just the best match for each ligand and do some aggregation. 

Something like: https://gist.github.com/jhjensen2/6450138cda3ab796a30850610843cfff will probably do the trick, and then some metric on the resulting graph?

JenkeScheen:
thanks @hmacdope that's helpful. Thinking about this it's actually non-trivial because it's not strictly about chemical similarity to the reference, it's also about whether the found reference is a good fit MCS-wise, i.e. if there are ring systems to be grown the reference should ideally not contain a ring on that site.

hmacdope:
Right, so this basically boils down to the "is this a good transformation" question.

JenkeScheen:
actually this might be easier than I thought because we're not actually using the crystal reference pose for calculations, just for alignment. So if we take a reference pose that spans the space of R groups we should generally be able to pose our prospective compounds in a way that agrees with the reference pose. 

So a good strategy would probably be to take the largest compound in the prospective set and find its closest neighbor in the reference crystals

[...]

PR #938, opened 2024-04-03
Automated ref finder for `asap-alchemy`

JenkeScheen:
## Description
For `asap-alchemy` we don't want to have to manually find a crystal structure to use as reference for docking every time we build a network. Instead, we want to find this automatically (https://github.com/choderalab/asapdiscovery/issues/918).

Caveats:
- we typically `prep` before we `dock`, but to find the correct reference we need the ligands, so this jumbles up the order a bit. I opted for loading the ligands during `prep`, then just re-loading them during `dock`. This isn't very elegant but has minimal overhead.
- references may/may not be compatible with `OpenMM` for a variety of reasons. I built in a check that makes sure we're moving forward with a compatible reference.
- I'm a horrible person so the code is a bit monolithic in its current state; @hmacdope should advise on how to move some things around, particularly in [this section](https://github.com/choderalab/asapdiscovery/pull/938/files#diff-e0af4b3e05dcd7e52116cd7e7fc0f91862414d4a9e628074511f00e732737315R389) and [this one](https://github.com/choderalab/asapdiscovery/pull/938/files#diff-e0af4b3e05dcd7e52116cd7e7fc0f91862414d4a9e628074511f00e732737315R371).
- I accidentally pushed to another branch (https://github.com/choderalab/asapdiscovery/pull/919) so I had to fix all that which resulted in this PR having a large bulk commit rather than incremental ones. 
- @jthorton is adding a `AlchemyPrepWorkflow` that the workflow in this PR could potentially migrated into once https://github.com/choderalab/asapdiscovery/pull/919 is merged


JenkeScheen:
@hmacdope can I grab a quick review? Feel free to just give some pointers on where things should be dropped if you think the current placing is off.

asapdiscovery-workflows/asapdiscovery/workflows/prep_workflows/protein_prep.py
```diff
@@ -276,3 +322,139 @@ def protein_prep_workflow(inputs: ProteinPrepInputs):
...
+from openff.toolkit import ForceField, Topology
+from openmm import unit as openmm_unit
+
+
+### SOMETHING GOING WRONG HERE - NO BOX?!
+def create_protein_only_system(pdb_path):
+    # attempt to make an OpenMM system with the prepped protein.
+    # follows https://docs.openforcefield.org/en/latest/examples/openforcefield/openff-toolkit/toolkit_showcase/toolkit_showcase.html
+    top = Topology.from_pdb(pdb_path)
+    sage_ff14sb = ForceField("openff-2.1.0.offxml", "ff14sb_off_impropers_0.0.3.offxml")
```
JenkeScheen, line 430: @jthorton notes we should use the FFs we use in production forr `asap-alchemy`

jthorton: The default force fields can be found [here](https://github.com/OpenFreeEnergy/gufe/blob/29f98ec0227ed9f30ea6b84096dc13818a9ac8d3/gufe/settings/models.py#L161-L166). I think we just need to make an openMM system using the topology and the force fields listed and provided we don't hit an exception the pdb should be valid. Something like this should work 
```python
forcefield = openmm.ForceField([
        "amber/ff14SB.xml",  # ff14SB protein force field
        "amber/tip3p_standard.xml",  # TIP3P and recommended monovalent ion parameters
        "amber/tip3p_HFE_multivalent.xml",  # for divalent ions
        "amber/phosaa10.xml",  # Handles THE TPO
    ])
pdb = openmm.app.PDBFile(pdb_path)
try:
    _ = forcefield.createSystem(pdb.topology)
except Exception as e:
    raise InvalidStructureError from e
```

hmacdope:
Closing in favour of #966

PR #947, opened 2024-04-04
Stopgap for no UUID in postera artifact upload

### LINKED ISSUE #946: Stopgap if no uuid compound slips through in docking workflows
Add a stopgap that prevents compounds not assigned a UUID from slipping though when uploading to Postera. We should also try and get to the bottom of why this is happening (only seen once.)

hmacdope:
FIxes #946 


hmacdope:
@ijpulidos @JenkeScheen @apayne97 could I grab a review here.

hmacdope:
> would assigning some random string work, rather than dropping the row? maybe I misunderstand what you're doing here.

No because to append will need a UUID already assigned on the postera end.

JenkeScheen (approved):

PR #966, opened 2024-04-10
`ASAP-Alchemy` reference finder

jthorton:
## Description
This is an alternative to #938 which uses the same method but moves the logic to `ASAP-Alchemy prep run` and allows users to provide a folder of prepared structures rather than a single complex and the most suitable structure will be selected for the constrained pose generation. The main benefit is that the logic of prepping is kept separate from selection which should simplify things. 

The current selection method is based on the maximum MCS overlap between the largest ligand of the set and the reference ligand which should ensure the binding site can accommodate all of the ligands. However, there is some evidence that choosing the structure with the highest affinity complex can lead to more accurate results https://pubs.acs.org/doi/10.1021/acs.jctc.8b01290 which might be worth considering. Maybe we could add a selection method option as well?

## Todos
- [x] expose a way to have a reference automatically selected 

## Questions
- [ ] Do we want to have another separate CLI entry point which just allows you to select the receptor?

## Status
- [ ] Ready to go


asapdiscovery-alchemy/asapdiscovery/alchemy/cli/prep.py
```diff
@@ -79,6 +79,12 @@ def create(filename: str, core_smarts: str):
     type=click.Path(resolve_path=True, exists=True, file_okay=True, dir_okay=False),
     help="The name of the JSON file which contains the prepared receptor complex including the crystal ligand.",
 )
+@click.option(
```
hmacdope, line 82: We use `--structure-dir` for this argument in `docking` and `wfs`, may be good to keep consistent.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/prep.py
```diff
@@ -128,6 +135,7 @@ def run(
     dataset_name: The name which should be given to the AlchemyDataset all results will be saved in a folder with the same name.
     ligands: The name of the local file which contains the input ligands to be prepared in the workflow.
     receptor_complex: The name of the local file which contains the prepared complex including the crystal ligand.
+    receptor_folder: The name of the folder which contains the prepared complexs and we should select the best reference from
```
JenkeScheen, line 138:
```suggestion
    receptor_folder: The name of the folder which contains the prepared complexes that we should select the best reference from
```

JenkeScheen:
@hmacdope yes that's the plan

@jthorton:
- are we sure that we want a folder with prepped structures be the input, rather than a folder with unprepped structures (and then prep a structure once it's been selected)? As we don't have an automated way of pulling new crystal structures (because of Fragalysis being broken) maintaining an up-to-date local folder with prepped structures would likely require some manual work.
- I currently don't see a check whether a structure is compatible with simulation (as in https://github.com/choderalab/asapdiscovery/pull/938#discussion_r1557585454). Would it make sense to build that in here?  I get that you might want to have that be a prep step in general but this would potentially exclude a bunch of structures that we could've otherwise used for e.g. docking 
- In your workflow here there is no fallback on when a structure isn't suitable for simulation. We'd need to build that in (by e.g. finding the next most similar reference) if we're going to build in something like I describe in the above bullet point

jthorton:
@JenkeScheen hopefully I have convinced you now that the improvements to the protein prep workflow should make it very easy to keep your prepped structure folder up to date (another example of why we need docs!) which should handle point one. For the other two I have now added your check I agree it should be done here to make sure the structures are suitable for simulation only and let the other stages such as docking decide if they should use a given structure or not.

I think this is now ready for another review, one thing to note is that we don't do any stereochemistry checking when matching to the reference structure which we might want to do in future.

hmacdope (approved):
LGTM @jthorton

PR #984, opened 2024-04-16
`ASAP-Alchemy`: Prediction no overlapping experimental data

jthorton:
## Description
This PR fixes a bug where absolute predictions become `nan` when experimental data is provided but does not overlap with the ligands in the network. We now only shift the data if at least one ligand in the network has an experimental data point this also means interactive reports are not generated unless there is overlapping experimental data. 

We also now extract the mbar error as our uncertainty value if only a single repeat is performed rather than the stdv which is supplied by gufe.

## Todos
- [x] make sure the predicted absolute values are valid even with no exp overlap
- [x] always have a non-zero error for FEC estimation.

## Questions
- [ ] Question 1 ..

## Status
- [x] Ready to go


jthorton:
looks like https://github.com/choderalab/asapdiscovery/pull/983 might have broken some of the imports used in testing.

jthorton:
@hmacdope I have reversed a change made to one of the init files in the repo, let me know if you want me to change the imports instead and have the init blank? Also the live testing of postera seems to be broken.

jthorton:
Just noticed the init is fixed on main so I have reversed the patch I added.

JenkeScheen (approved):
LGTM, great work!

Issue #927, opened 2024-03-28
asap-prep does not support --oe-active-site-residue:


SiyangPeng:
I have been trying to run command line but : 
asap-prep --target ZIKV-NS2B-NS3pro --pdb-file zikv_ns3.pdb --output-dir orig_prepped --align zikv_ns3.pdb --oe-active-site-residue TYR:161::B --ref-chain B --active-site-chain B

The error log indicates: 
Warning: Failed, no pocket found with site residue predicate
Warning: Unable to parse and utilize specified site residue TYR:161::B to define site in _chainB_ASP-MET-TYR-ILE-GLU-ARG-ALA-GLY-ASP-ILE-THR-TRP-GLU-LYS-ASP-ALA-GLU-VAL-THR-GLY- - skipping
ValueError: Failed to make design unit.

I identified the active site residue by running this code block using the get_interactions_plip function from asapdiscovery plip (https://github.com/choderalab/asapdiscovery/blob/main/asapdiscovery-data/asapdiscovery/data/backend/plip.py#L214):
protein_orig = load_openeye_pdb('/home/pengs/fold_zika/test_prep/zikv_ns3.pdb')
pose_orig = load_openeye_sdf('/home/pengs/fold_zika/lig_sdfs/lig_0.sdf')
color_method = 'fitness'
target = "ZIKV-NS2B-NS3pro"
get_interactions_plip(protein_orig, pose_orig, color_method, target)

I think the reason why this is failing is because I am not able to identify the residue identified by plip with the oeGraphMol object constructed during prepping and so it could not identify a binding pocket to dock the ligand even though there is a clear pocket there.

hmacdope:
@SiyangPeng can I suggest you try an make a minimal example to replicate this that uses **just** the raw underlying openeye functionality ie `OEMakeDesignUnits`  oespruce.OEMakeDesignUnits(initial_prot, metadata, opts, site_residue)

Forget about PLIP for now, is there a tyrosine at position 161 in the input protein? 

Also check carefully that you are using the right OpenEye syntax for the residue. This issue happens most often when you get that very slightly wrong (this has happened to me a lot).

SiyangPeng:
I have tried to do that. It did not make a designunit. The error message was 
`Warning: No BioAssembly transforms found, using input molecule as biounit: 
Warning: Iridium - Structure:  has no REMARK data
Processing BU # 1 with title: , chains AB
[...]

PR #1013, opened 2024-04-25
Add variant transfer prepper v2

apayne97:
## Description
copied over from #972 
We want a way to transfer molecules from a crystal structure to an alphafold (or otherwise generated) structure and then re-dock them with OpenEye / etc.

Based on *very* old code from ben.

This is also a way around #927 

## Todos
- [x] re-created `make_du_from_new_lig`
- [x] created LigandTransferProteinPrepper
- [x] created ligand_transfer_docking_workflow
- [x] moved some inputs options to WorkflowInputsBase so that I can re-use them without needing to use everything else
- [x] created cli
- [x] incorporate maria's changes
- [ ] speed up prep with parallelization - use just spruce (i.e. don't make a DU) to prep the structures, then pass that structure 1 time with all the ligands
- [ ] make tests
- [ ] make how-to

## Questions
- [ ] Question 1 ..

## Status
- [ ] Ready to go

- [ ] I certify that this contribution is covered by the MIT license as defined in our [LICENSE](https://github.com/choderalab/asapdiscovery/blob/main/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).


asapdiscovery-workflows/asapdiscovery/workflows/docking_workflows/ligand_transfer_docking.py
```diff
@@ -0,0 +1,314 @@
...
+        dask_client=dask_client,
+        return_df=False,
+        dask_failure_mode=inputs.dask_failure_mode,
+    )
+    del results
+    import csv
+    with open(data_intermediates / "docking_scores_raw.csv", 'w', newline='') as myfile:
+        wr = csv.writer(myfile, quoting=csv.QUOTE_ALL)
+        for item in scores_list:
+            wr.writerow([item])
```
apayne97: @mariacm12 could you add info on the error you got? is it the target-related stuff?

mariacm12: It's because the dataframe has duplicate entries I think. Although I couldn't see the duplicates myself when I generated the csv file the alternative way (from DataFrame.pivot() docs: _Raises ValueError when there are any index, columns combinations with multiple values_)

asapdiscovery-modeling/asapdiscovery/modeling/protein_prep.py
mariacm12, line 1: Here I had to use the hash of the reference complex, but the target name of the actual target. This is a quick fix because the `hash` and `unique_name` attributes  depend on the ligand's InChIKey, which in this case the target doesn't have. Previously the target name of the reference was used so it was re-writting all of the docking results with the same name (the reference's).
The hash string still needs to be fixed though, I just don't know how to do it without breaking the other workflows. because the current fix still leads to problems when importing files from cache.

asapdiscovery-workflows/asapdiscovery/workflows/docking_workflows/ligand_transfer_docking.py
```diff
@@ -0,0 +1,314 @@
...
+        logger.info("Writing final docked poses to SDF file")
+        write_ligands_to_multi_sdf(
+            output_dir / "docking_results.sdf", [r.posed_ligand for r in results]
+        )
+    # NOTE: The return_df option doesn't work, the DataFrame.pivot() function fails. Fix is pending.
+    scores_list = scorer.score(
+        results,
+        use_dask=inputs.use_dask,
+        dask_client=dask_client,
+        return_df=False,
```
hmacdope, line 308: If you return_df=True here will be a lot easier to wrangle, including CSV using `df.save_csv` below.

apayne97: @hmacdope see Maria's comments, `return_df` isn't working. ultimately we'd want to use that though

hmacdope:
@apayne97 has this been superseded now? If so can you close? :smile:

apayne97:
Yes, I think Maria's branch #1110 took what was needed from this!

PR #997, opened 2024-04-18
`asap-docking` don't return None in POSITDocker

apayne97:
## Description
fixes #989 

## Todos
Notable points that this PR has either accomplished or will accomplish.
- [ ] change POSITDocker to not return anything in the case of a skipped failure
- [ ] write a test to catch this bug

## Questions
- [ ] in terms of test writing - it seems overkill to go through docking that we know fails just to test what happens when it fails, but I'm not sure how else to test it?

## Status
- [ ] Ready to go

## Developers certificate of origin
- [ ] I certify that this contribution is covered by the MIT license as defined in our [LICENSE](https://github.com/choderalab/asapdiscovery/blob/main/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).

Issue #989, opened 2024-04-17
`asap-docking` Missing docking result causes a bug


apayne97:
I think I've found the bug, but we need a solution.

Appending `None` to this list of docking results here:
https://github.com/choderalab/asapdiscovery/blob/9f73bb2d99b60176a7b7ff754aa08fc78bdb40e9/asapdiscovery-docking/asapdiscovery/docking/openeye.py#L336

causes an error here in `small_scale` and `cross_docking`
https://github.com/choderalab/asapdiscovery/blob/9f73bb2d99b60176a7b7ff754aa08fc78bdb40e9/asapdiscovery-workflows/asapdiscovery/workflows/docking_workflows/cross_docking.py#L224-L228

https://github.com/choderalab/asapdiscovery/blob/9f73bb2d99b60176a7b7ff754aa08fc78bdb40e9/asapdiscovery-workflows/asapdiscovery/workflows/docking_workflows/small_scale_docking.py#L323-L327

and I haven't tested it but I think it should cause an error in `large_scale_docking` here:
https://github.com/choderalab/asapdiscovery/blob/9f73bb2d99b60176a7b7ff754aa08fc78bdb40e9/asapdiscovery-workflows/asapdiscovery/workflows/docking_workflows/large_scale_docking.py#L286-L293

here's my error:
```
Traceback (most recent call last):
  File "/home/paynea/miniforge3/envs/asapdiscovery/bin/asap-docking", line 8, in <module>
    sys.exit(docking())
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/lila/home/paynea/miniforge3/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
[...]

PR #1018, opened 2024-04-26
Add resilience to non-dask computation

### LINKED ISSUE #1017: Add resilience for non-dask mode
We need better resilence for running in a non-dask mode.

hmacdope:
FIxes #1017 

Needs some wrangling around the `failure-mode` argument, but also fixes a few issues, including incorporating the changed in #997 (tagging @apayne97 ) which had a subtle interaction with the way dask backend handled array shaping, leading to #989 

### DIFF
--- a/asapdiscovery-cli/asapdiscovery/cli/cli_args.py
+++ b/asapdiscovery-cli/asapdiscovery/cli/cli_args.py
@@ -1,6 +1,6 @@
 import click
 from asapdiscovery.data.services.postera.manifold_data_validation import TargetTags
-from asapdiscovery.data.util.dask_utils import DaskFailureMode, DaskType
+from asapdiscovery.data.util.dask_utils import FailureMode, DaskType
 from asapdiscovery.ml.models import ASAPMLModelRegistry
 from asapdiscovery.simulation.simulate import OpenMMPlatform
 
@@ -54,18 +54,18 @@ def dask_type(func):
     )(func)
 
 
-def dask_failure_mode(func):
+def failure_mode(func):
     return click.option(
-        "--dask-failure-mode",
-        type=click.Choice(DaskFailureMode.get_values(), case_sensitive=False),
-        default=DaskFailureMode.SKIP,
+        "--failure-mode",
+        type=click.Choice(FailureMode.get_values(), case_sensitive=False),
+        default=FailureMode.SKIP,
         help="The failure mode for dask. Can be 'raise' or 'skip'.",
         show_default=True,
     )(func)
 
 
 def dask_args(func):
-    return use_dask(dask_type(dask_failure_mode(func)))
+    return use_dask(dask_type(failure_mode(func)))
 
 
 def target(func):
--- a/asapdiscovery-cli/asapdiscovery/cli/cli_args.py
+++ b/asapdiscovery-cli/asapdiscovery/cli/cli_args.py
@@ -1,6 +1,6 @@
 import click
 from asapdiscovery.data.services.postera.manifold_data_validation import TargetTags
-from asapdiscovery.data.util.dask_utils import FailureMode, DaskType
+from asapdiscovery.data.util.dask_utils import DaskType, FailureMode
 from asapdiscovery.ml.models import ASAPMLModelRegistry
 from asapdiscovery.simulation.simulate import OpenMMPlatform
 
--- a/asapdiscovery-data/asapdiscovery/data/operators/selectors/selector.py
+++ b/asapdiscovery-data/asapdiscovery/data/operators/selectors/selector.py
@@ -6,7 +6,7 @@ from asapdiscovery.data.schema.complex import Complex, PreppedComplex
 from asapdiscovery.data.schema.ligand import Ligand
 from asapdiscovery.data.schema.pairs import CompoundStructurePair
 from asapdiscovery.data.util.dask_utils import (
-    DaskFailureMode,
+    FailureMode,
     actualise_dask_delayed_iterable,
 )
 from asapdiscovery.docking.docking import DockingInputPair  # TODO: move to backend
@@ -31,7 +31,7 @@ class SelectorBase(abc.ABC, BaseModel):
         complexes: list[Union[Complex, PreppedComplex]],
         use_dask: bool = False,
         dask_client=None,
-        dask_failure_mode=DaskFailureMode.SKIP,
+        failure_mode=FailureMode.SKIP,
         **kwargs,
     ) -> list[Union[CompoundStructurePair, DockingInputPair]]:
         if use_dask:
@@ -43,7 +43,7 @@ class SelectorBase(abc.ABC, BaseModel):
                 # see # 560
                 delayed_outputs.append(out)
             outputs = actualise_dask_delayed_iterable(
-                delayed_outputs, dask_client, errors=dask_failure_mode
+                delayed_outputs, dask_client, errors=failure_mode
             )
             outputs = [
                 item for sublist in outputs for item in sublist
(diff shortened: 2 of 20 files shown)


apayne97:
@hmacdope I appreciate this, but surely adding a bunch of `try, except` clauses is not an appropriate solution? I think we should move towards a case where dask doesn't do *any* error processing, otherwise we will constantly be fighting with dask and no-dask processes getting out of date? Why can't we instead rewrite our functions to not return unreadable outputs?

hmacdope:
There is a few layers to this and I will unpack them one by one. I probably should have done this in a more atomised way so that the changes and intent were clearer.

> @hmacdope I appreciate this, but surely adding a bunch of `try, except` clauses is not an appropriate solution?

Firstly, our first priority should be to have resilience to compute failure in any given sub-task. Each workflow needs **some** way to be able to skip failed attempts to dock, prep, score etc as we are dealing with large numbers of ligands, structures and scores, if we don't have resilience for failed compute, most workflows are never going to finish as one `ValueError` or something will be the end of it all. 

Currently we let dask handle this, but then we have no resilience in non-dask mode, which has tripped people up a fair bit. This PR is mainly trying to fix the lack of non-dask resilience, which also led to some other stuff. 

The only really robust code construct to catch failed attempts  is a `try: except: clause. Indeed this is what dask does under the hood when it runs compute on its backend. So I would argue there is no real way around having a try:except in the function body that does the work, you can change what level the exceptions are caught at by moving the try:except , but catching them still must be done at some point. 

In summary the `try: except` stuff is orthogonal  to dask dask really, it is more adding the ability for **non-dask** computation to fail gracefully without killing the workflow. 


 > I think we should move towards a case where dask doesn't do _any_ error processing, otherwise we will constantly be fighting with dask and no-dask processes getting out of synce? Why can't we instead rewrite our functions to not return unreadable outputs?
 
 Dask has to do error **resilience** but the key question here is what is the idiomatic thing to return when there is unsuccessful compute. I agree that we should try and have dask and non-dask behaviour as close together as possible, (hence this PR trying to make non-dask failure resistent)
 
 A value is always returned to the calling function, even if you don't have a `return` statement (see below).  So we can't just let control flow reach end or we will have same issue. 
 
 ```python
 def func_with_no_return()
     a = 1
 b = func_with_no_return()
 print(b)
 >>> None
 ```
  
 So the question is mostly what should the idiomatic return value be given our setup? I agree with you that explicitly returning `None` is bad and we should get rid of it. (thanks for prod in #989 )
 
  
> [EDIT] this reads more combative than I meant it. I think my concern is that if we are changing the behavior by adding a bunch of boiler-plate code instead of fixing the core issue, we are guaranteeing that we'll have frustrating-to-debug errors later. To my point, you ended up using my main change from #989 - i.e. changing the function that is returning the unreadable result - instead of relying on catching it in the workflows. 

Not too combative at all, good to discuss opinions, frank feedback is often most constructive. Yes removing explicit returns of `None` seems to be the sensible thing to do. I think the one addressed in #989 is the only one. 

> I think for any function where you think it's necessary for dask to catch `None`, we should be doing something else in the function return. Silently getting rid of those in a dask mapper, which no would expect to be doing anything except for parallelization, is a nightmare for debugging.

I will lay out my case that the idiomatic return value here should be `[ ]` if there is no successful compute and why dask still needs to do something with this, but overall I think we are very closely aligned. 

From the top level the call always looks the same 


```python 

def dock(inputs: list[InputType], ):
     _dock(inputs=inputs)

# lower level
def _dock(inputs):
    results = [ ]
    for input in inputs:
         try:
             # do something
             results.append(blah) # it worked 
         except:
             # fail or skip 
    return results
```

There are two ways these functions are called 

# Non dask mode

 ```python
 # lower level, 5 inputs go in 
def _dock(inputs):
    results = [ ]
    len(inputs)
    >>> 5
    for input in inputs:
         try:
             # do something
             results.append(blah) # it worked 
         except:
             # fail or skip, for this example, lets say one failed
             
     len(results)
     >>> 4
     return results # seems ok
```

Returning a list of 4 results here feels idiomatic, as there were 4 successful runs. 
 
# Dask mode
remembering that the dask wrapper parallelised over inputs 

 ```python
 # lower level
def _dock(inputs):
    results = [ ]
    len(inputs) # always one as dask parallelised over the inputs 
    >>> 1
    for input in inputs:
         try:
             # do something
             results.append(blah) # it worked 
         except:
             # fail or skip, for this example, lets say this one failed
             
     len(results)
     >>> 0 
    print(results)
    >>> [] # possible to return empty list here, needs to be handled in context of all other parallel computations
    return results
    
    
 #  dask backend then gets a list of the _dock calls 
 def dask_backend(inputs, func):
   collated_results = []
   for inp in inputs: # parallelised over inputs
       collated_results.append(func(inp))
  print(collated_results)
  >>>  [[result1], [result2], [], [result4], [result5]]
  # need to remove empties and flatten 
  collated_results = [r for r in collated_results if r] # empty list is `Falsy`, so is `None`
  results = np.ravel(collated_results).to_list()
  len(results)
  >>> 4
   # same as non dask mode
```

Long story short, dask wrapper has to handle the empty list case regularly. Previously this returned `None` sometimes, but we should have eliminated this now including #989. Happy to link to discuss also.

apayne97:
@hmacdope thank you so much for the thorough explanation. as you predicted, I hadn't put together all the points you made which makes error handling a requirement for dask (namely, that each function is going to return *something* when you call it, and it will return `None` if you don't give it a better plan, which means you will inevitably end up having to deal with those `None`s). 

We'll have to go case by case for the different functions to make sure they are failure tolerant and that what they returns makes sense. The case for returning `[ ]` in this case makes sense to me!

I guess a problem to solve in the future is how to cleanly record failed results. Right now you would just grep the log files, which is *ok* but not great. I could see an argument for a POSITDockingResult object that can handle failed results but that seems annoying to implement. We could mimic what @jthorton did in the pose generation code with a `POSITDockingResults` object instead of a list with `failed` and `posed` properties that are each lists of, i.e. POSITDockingResult and POSITDockingResultFailed or something (maybe the DockingInputBase). 

Another concern is that from a code-writing perspective, we'll have to reimplement the `try: except` structure in every function that will use dask, which may not be obvious, otherwise we'll have this silent error handling problem again?
Is it silly to have an abstract class that forces you to implement this? or some other structure? maybe it is. just trying to keep the checklist of things you need to add to a function you might want to write to a minimum.

hmacdope:
> @hmacdope thank you so much for the thorough explanation. as you predicted, I hadn't put together all the points you made which makes error handling a requirement for dask (namely, that each function is going to return _something_ when you call it, and it will return `None` if you don't give it a better plan, which means you will inevitably end up having to deal with those `None`s).
> 
> We'll have to go case by case for the different functions to make sure they are failure tolerant and that what they returns makes sense. The case for returning `[ ]` in this case makes sense to me!
> 
> I guess a problem to solve in the future is how to cleanly record failed results. Right now you would just grep the log files, which is _ok_ but not great. I could see an argument for a POSITDockingResult object that can handle failed results but that seems annoying to implement. We could mimic what @jthorton did in the pose generation code with a `POSITDockingResults` object instead of a list with `failed` and `posed` properties that are each lists of, i.e. POSITDockingResult and POSITDockingResultFailed or something (maybe the DockingInputBase).

Lets move this to an issue, fair point to discuss.

> Another concern is that from a code-writing perspective, we'll have to reimplement the `try: except` structure in every function that will use dask, which may not be obvious, otherwise we'll have this silent error handling problem again? Is it silly to have an abstract class that forces you to implement this? or some other structure? maybe it is. just trying to keep the checklist of things you need to add to a function you might want to write to a minimum.

Just to clarify something, the `try: except` structure is not needed **for** dask, it is needed for these functions to be failure tolerant **outside** of dask, which has a `try: except` wrapped around the `.compute()` call.  You are right in that we should add some documentation showing what needs to be done.

hmacdope:
I am going to merge here so I can move on to docs, forgive my lack of review process.

apayne97:
> Just to clarify something, the try: except structure is not needed for dask, it is needed for these functions to be failure tolerant outside of dask,

yes, i know! my point is just that if we want the behavior to be the same we'll need to include try: except

PR #1019, opened 2024-04-26
Fill in ML guides

kaminow:
Fill in notebooks and guides:

- [x] [docs/guides/using_ml_cli.md](https://github.com/choderalab/asapdiscovery/blob/main/docs/guides/using_ml_cli.md)
- [x] [examples/training_ml_models_on_asap_data.ipynb](https://github.com/choderalab/asapdiscovery/blob/main/examples/training_ml_models_on_asap_data.ipynb)
- [x] [examples/running_ml_inference.ipynb](https://github.com/choderalab/asapdiscovery/blob/main/examples/running_ml_inference.ipynb)


kaminow:
running into an issue where the ML notebooks reference results from other sections. added a line in `docs/conf.py` to ignore `nbsphinx` errors for now, but that's probably not a good actual solution. would it make sense to upload some test files to an S3 bucket and point to those instead of the imaginary files I have currently @hmacdope?

kaminow:
or alternatively is there a way to just have Sphinx render the notebooks but not run them? I don't think we want to do a full training run every time the docs are built

hmacdope:
Can you use the files from the testsuite? That is often a good solution to use files that are useful and public. You can just use `get_test_file` inline

kaminow:
oh yeah duh, that should work. thanks!

hmacdope (changes requested):
Looks great, just a bit of introduction and citations for model types would be great in notebook and also CLI guide.

kaminow:
@hmacdope addressed I think, lmk if this is what you were looking for or if I should add a bit more!

hmacdope (approved):
LGTM @kaminow! Thanks for pushing on this.

PR #1032, opened 2024-05-03
Viz tutorial!

### LINKED ISSUE #955: [Documentation]: Visualisations and how to make them
Document the cool types of visualisations we can make

hmacdope:
Fixes #955


hmacdope:
@apayne97 @kaminow could you have a quick run through this one please? I will find a solution for vendoring the HTML files during build ASAP, but just focus on content for now.

kaminow:
very nice! I fixed a couple small things that I saw as I was going through, but I think on the whole this is good. two outstanding things that I noticed:
* missing DOI for Mac1 in the DMS section
* there isn't really any discussion in the DMS section on how to incorporate your DMS data into the structure. not sure if there's some standard accepted way to do this, but otherwise it might be worth walking through how you go from the CSV file or whatever that you get from a DMS experiment to a PDB file that you can color with `color_method="fitness"`

hmacdope:
@kaminow addressed comments. :)

kaminow:
sorry I guess my main thing with the second point is, what kind of data needs to be in `results` in order to be able to use `color_method="fitness"`? like if I just pass any old docked structure what will happen?

kaminow:
also, one other thing with the `.gitignore` file, do we want to be including extra paths and stuff (eg `examples/html`) in there? or should we just leave it up to people to not commit the files that they're generating when testing their notebooks? alternatively, is there a way to just blanket ignore any file that's not `*.ipynb` in `examples/`?

hmacdope:
> sorry I guess my main thing with the second point is, what kind of data needs to be in `results` in order to be able to use `color_method="fitness"`? like if I just pass any old docked structure what will happen?

Yep any old structure will work in theory. It always overlays the same data on the 3D structure so not much in way of requirements.

hmacdope:
> also, one other thing with the `.gitignore` file, do we want to be including extra paths and stuff (eg `examples/html`) in there? or should we just leave it up to people to not commit the files that they're generating when testing their notebooks? alternatively, is there a way to just blanket ignore any file that's not `*.ipynb` in `examples/`?

We may need to commit some but not all of these files explicitly to have the visualizations render in the notebook, I just wanted to avoid committing things I didn't mean to. 

There is syntax to achieve your second point IIRC which is a good idea.

kaminow:
> Yep any old structure will work in theory. It always overlays the same data on the 3D structure so not much in way of requirements.

ah ok, I think I wasn't understanding that the fitness data is based on the target and built into the package, rather than something that the user should provide. maybe adding a note in that section would be helpful, but other than that I think everything makes sense

PR #1037, opened 2024-05-03
Rework sweep CLI logic

kaminow:
We updated the way the training CLI handles combining config caches and CLI args, but never updated the sweep CLI in the same way. This PR addresses that and makes sure the sweep CLI reconciles everything in a sensible manner.


hmacdope (approved):
LGTM @kaminow

PR #1046, opened 2024-05-09
`ASAP-Simulate`: update small molecule FF in `VanillaMDSimulator`

jthorton:
## Description

This PR exposed the small molecule OpenFF force field used in the `VanillaMDSimulator` as an option and updates the default to `openff-2.2.0.offxml` to take advantage of the improved sulfamide modelled geometries! 

## Todos
- [x] update the default small molecule force field
- [x] expose setting the force field as an option in the schema

## Questions
- [ ] Should we also expose the protein force field? 

## Status
- [x] Ready to go

### DIFF
--- a/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
+++ b/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
@@ -164,6 +164,11 @@ class VanillaMDSimulator(SimulatorBase):
         description="Whether to truncate num_steps to multiple of reporting interval, used mostly for testing",
     )
 
+    small_molecule_force_field: str = Field(
+        "openff-2.2.0.offxml",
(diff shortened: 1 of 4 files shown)


JenkeScheen:
should the new version of openff be pinned in the env yamls?

jthorton:
@JenkeScheen we don't have any pins around the openff-toolkit so it should always install the newest version but we could pin it just to ensure the default always works!

JenkeScheen (approved):
yep go ahead and pin a minimal version and then merge!

JenkeScheen:
no need to expose yet imo

hmacdope:
Thanks for doing this!

hmacdope:
> @JenkeScheen we don't have any pins around the openff-toolkit so it should always install the newest version but we could pin it just to ensure the default always works!

We should pin just to be clear.

PR #1061, opened 2024-05-28
switch to pIC50 view in HTML reports

JenkeScheen:
@jthorton we should switch to pIC50 plotting by default. I know we want to have a toggled view between pIC50 and DG but this is a workable form for now, albeit not very elegantly written.


JenkeScheen:
this is a much more useful view for debugging because we deal with pIC50 on all other levels (moleculesets, medchem decision-making, ML predictions etc)

hmacdope (approved):
LGTM

jthorton (approved):
Looks great!

PR #1096, opened 2024-06-06
Add option for adjusting weights saving frequency

### LINKED ISSUE #1063: Add configurable option to not write .th files for every training step, or clean them up
As we scale up ML training we should only write out final models OR try and clean them up post training.

kaminow:
Current training implementation saves a new weights file for each training epoch. This is nice for debugging but not for prod. This PR adds an option to `Trainer` (and to CLI) to control how often weights are saved. The new parameter is `Trainer.save_weights`, or `--save-weights` on the CLI. The options are:

* `"all"`: Keep every epoch's weight file (current behavior and will continue to be the default)
* `"recent"`: Only keeps the most recent epoch's weight file, saves it as `weights.th` in the output directory
* `"final"`: Only keep the final epoch, after the model has finished its entire training, saves as `final.th` (this file is included in both of the above options as well)

Note that if you select `"final"`, if your training job runs out of time or dies for any reason there will be no weights files generated. For this reason, I think the recommended option for production is `"recent"`, as this will allow continuing a run while only storing one extra weights file.

I think this closes #1063, but lmk if this works for you @JenkeScheen @hmacdope.

### DIFF
--- a/asapdiscovery-ml/asapdiscovery/ml/cli_args.py
+++ b/asapdiscovery-ml/asapdiscovery/ml/cli_args.py
@@ -28,6 +28,18 @@ def output_dir(func):
     )(func)
 
 
+def save_weights(func):
+    return click.option(
+        "--save-weights",
+        type=click.Choice(["all", "recent", "final"], case_sensitive=False),
+        help=(
+            "How often to save weights during training."
+            'Options are to keep every epoch ("all"), only keep the most recent '
+            'epoch ("recent"), or only keep the final epoch ("final").'
+        ),
+    )(func)
+
+
 def trainer_config_cache(func):
     return click.option(
         "--trainer-config-cache",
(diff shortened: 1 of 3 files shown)


JenkeScheen:
Nice, what will the behavior be with early stopping? From the `--help` it's not immediately clear. I presume `final` would loop back and take the 'best' epoch according to the early stopping setup, correct?

kaminow:
yeah I can make it more clear in the help text, but as long as training completes, the `final.th` weights file will be generated regardless of which option you choose for `--save-weights`. if you used early stopping, it will have the weights from the converged/best epoch, based on which early stopping method you're using. otherwise, it will just be the weights from the end of the training run

asapdiscovery-ml/asapdiscovery/ml/cli_args.py
```diff
@@ -28,6 +28,18 @@ def output_dir(func):
     )(func)
 
 
+def save_weights(func):
+    return click.option(
+        "--save-weights",
+        type=click.Choice(["all", "recent", "final"], case_sensitive=False),
+        help=(
```
hmacdope, line 35: Give it a default?

JenkeScheen: should be `final` imo

PR #1113, opened 2024-06-21
`asap-alchemy prep alchemize`

JenkeScheen:
We want to be able to run alchemy for P3 but it’s HtL so chemistry is much more diverse. We want to be able to find the maximum number of sub-networks we can run from a set like e.g. https://asap.postera.ai/molecule-set/7eaecca8-25b9-4e1e-9a2f-87336fc9d791?viewer=7eaecca8-25b9-4e1e-9a2f-87336fc9d791

and then generate SDF files (for now) for the clusters so that we can feed each of those into `asap-alchemy prep run`.


JenkeScheen:
this is already quite fast on an M2 macbook pro:
image
and should scale linearly. We'll need to implement multiprocessing if we ever go to sets of thousands.

JenkeScheen:
OK - now with SDF writing enabled:
image

We should discuss how we're going to actually implement this thing, currently it's way too monolithic; a few of the methods should probably be migrated elsewhere. Also still need to write tests.

JenkeScheen:
tested this on a larger set, still runs nicely on only a single processor: 
image

JenkeScheen:
parking for now, will figure out where this should live after my leave.

JenkeScheen:
@hmacdope this is quite complete now - can you have a look through and if you're feeling it perhaps add multiprocessing that works correctly with `tqdm` in `alchemize.rescue_outsiders()`?

hmacdope (changes requested):
Looking great. I just re-jigged the tests a little bit. 

Only question is why is default Mac1 structure changed? 

Re parallelisation, I think lets leave it for now. We can revisit if its a consistent annoyance / blocker.

JenkeScheen:
good catch on mac1! That's definitely not intended.

hmacdope (approved):

PR #1114, opened 2024-06-24
S3 push following  ml training and link to W&B

### LINKED ISSUE #1082: Enable optional push of ML artefacts to S3 bucket following training
For our ML scaleup work, we need to optionally be able to save the final weights and the config from training an **ensemble** of ML models  in an S3 bucket that we can then pull from later. 

The layout I was thinking of directory was 

`https://asap-discovery-ml-weights.s3.amazonaws.com/training/<endpoint>/<endpoint>_<isodate>/<endpoint>_<isodate>_<model_index>.th`

Optionally we can also add references in weights and biases: https://docs.wandb.ai/guides/artifacts/track-external-files

## MVP
- [ ] @kaminow would you be able to test adding an S3 push at the end of a training run?  Should be as simple as using this code here (https://github.com/asapdiscovery/asapdiscovery/blob/main/asapdiscovery-data/asapdiscovery/data/services/aws/s3.py) and doing something like

```python
from asapdiscovery.data.services.aws.s3 import S3
aws_s3_settings = S3Settings()
s3 = S3.from_settings(aws_s3_settings)
s3.push_file("local_path/thing.th", location=training/<endpoint>/<endpoint>_<isodate>/<endpoint>_<isodate>_<model_index>.th, content_type="application/octet-stream")
```

##Notes
If you make the bucket path a configurable option we can just pass it into the CLI, no need to do internally.

hmacdope:
Fixes #1082 


hmacdope:
@kaminow this allows push to an S3 bucket and cross linking of final model with a W&B artifact.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -994,12 +1019,32 @@ def train(self):
...
+            s3 = S3.from_settings(self.s3_settings)
+            s3.push_file(
+                final_model_path,
+                location=self.s3_path,
+                content_type="application/octet-stream",
+            )
+            if self.use_wandb:
+                # track S3 artifacts
+                print("Linking S3 artifacts to W&B")
+                model_artifact = wandb.Artifact(
```
kaminow, line 1037: maybe I'm missing something, but does this encode any S3 information anywhere/do we want that to be the case? seems like we're just separately pushing to S3 and to W&B

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -124,6 +126,11 @@ class Trainer(BaseModel):
         None, description="Any extra config options to log to W&B."
     )
 
+    # artifact tracking options
+    upload_to_s3: bool = Field(False, description="Upload artifacts to S3.")
+    s3_settings: S3Settings | None = Field(None, description="S3 settings.")
```
kaminow, line 131: I didn't see this arg in the CLI stuff, do we want to be able to specify this there?

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -518,6 +536,13 @@ def initialize(self):
                 logfile=str(self.log_file.name),
             ).getLogger()
 
+        # check S3 settings so that
+        if self.upload_to_s3:
+            try:
+                self.s3_settings = S3Settings()
```
kaminow, line 542: check if something  was specified for `self.s3_settings` before overwriting? not sure what all is stored in the `S3Settings` object

kaminow (changes requested):
looks good, thanks for taking care of this? just a couple small questions

asapdiscovery-ml/asapdiscovery/ml/cli.py
```diff
@@ -96,6 +97,7 @@ def build_ds():
 @loss_args
 @trainer_args
 @overwrite_args
+@s3_args
```
hmacdope, line 100: @kaminow s3 CLI args here. :)

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: Good point.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: Ah hang on I think I am not doing the linking properly, ill revisit.

hmacdope: Good spot.

hmacdope: Should be better now.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: Fixed.

asapdiscovery-ml/asapdiscovery/ml/trainer.py
kaminow: sorry, I meant for the `s3_settings` specifically

asapdiscovery-ml/asapdiscovery/ml/trainer.py
kaminow: yeah looks good, thanks!

asapdiscovery-ml/asapdiscovery/ml/trainer.py
hmacdope: Ahhhhhhh, I would say lets leave this out of CLI as its composed of 3 separate (sensitive) environment variables, better to store in an env file,  if we need the flexibility later we can add it in.

kaminow: ok sounds good. as long as y'all have a way of specifying everything you need when running it, I'm satistfied

kaminow (approved):
looks good!

PR #1119, opened 2024-06-28
DNM `ASAP-Alchemy`: NEQ support

### LINKED ISSUE #1116: Support NEQCycling protocol in ASAP Alchemy
We should add support for the NEQ Cycling protocol in ASAP Alchemy. 

I will leave @jthorton @ijpulidos  @dotsdl to enumerate what is required on this issue, as you have a better overview of the blockers.

jthorton:
## Description
This PR adds support for NEQ protocols in FEFlow. 
This is a test PR to check the changes required to add support for NEQ and other protocols in future. 

### Design
Following the OpenFE design we might be able to create a settings class for each of the protocols and have a single Factory and Network class which just works with these settings objects to make sure the correct protocol is used in the network. This would allow us to reuse the mapping/planning part of the factory for each protocol and have a single interface which converts our networks to an openfe network. 

In the future with living networks these settings classes might also add value as we can set a different protocol for different types of transformation like NEQ for normal, and ATM/SepTop for core changes. 

The downside of this large change is that any run networks before this change will not be compatible with the updated models without migration. On the other hand, we are still developing quickly so maybe breaking this is okay for now if we aim to have the schema settle down by version 1.0.0?

## Todos
- [ ] Point 1 ...

## Questions
- [ ] Question 1 ..

## Status
- [x] Ready to go


hmacdope:
@JenkeScheen this is where start of implementation is.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec/protocols.py
```diff
@@ -12,42 +12,16 @@
...
-        0.15 * OFFUnit.molar,
-        description="The ionic concentration required in molar units.",
-    )
-
-    def to_solvent_component(self) -> gufe.SolventComponent:
-        return gufe.SolventComponent(**self.dict(exclude={"type"}))
-
-
+# DD: I think we'll want to try and avoid duplicating the attribute structure of Protocol settings here;
+# though we don't expect them to change often, it would be cleaner to avoid duplication
```
jthorton, line 24: Great point, I would love to use the openfe/gufe models here if possible and did originally try this to track the provenance of all settings used in the calculation but hit issues with serialisation see https://github.com/OpenFreeEnergy/openfe/issues/518, or found that some settings like the `SolventSettings` class above didn't have a model in openfe.

hmacdope:
@ijpulidos would you be able to review here once complete?

jthorton:
blocked by https://github.com/OpenFreeEnergy/feflow/pull/88

dotsdl:
@jthorton I tried performing a temporary override to address the problem addressed upstream in https://github.com/OpenFreeEnergy/feflow/pull/88, but this gives very strange behavior. Calling `planned_network.to_alchemical_network()` after [this line](https://github.com/asapdiscovery/asapdiscovery/blob/df5b5f8f27f1f646e26012f80469b05d711a5d9c/asapdiscovery-alchemy/asapdiscovery/alchemy/utils.py#L102) gives an object with a different `gufe` key than that when called before that line. This is triggering the test failure shown in the [CI currently](https://github.com/asapdiscovery/asapdiscovery/actions/runs/10946784474/job/30394072783?pr=1119#step:8:406).

jthorton:
Currently running a test on tyk2 on alchemiscale.

dotsdl:
@jthorton is this working out okay?

jthorton:
@dotsdl we are not sure yet we have hit issues with live testing on alchemiscale see https://github.com/OpenFreeEnergy/feflow/issues/89

dotsdl:
@jthorton Hmmm...since the `RelativeHybridTopologyProtocol` can also handle taking a single `ComponentMapping` instead of a list of them, can we change the behavior here in `asap-alchemy` to do this for all protocols?

jthorton:
Good point this should be fixed up stream but I have also changed alchemy to provide a single mapping now as well, will re run the test network!

hmacdope:
@jthorton any news on how this went?

jthorton:
Getting closer 2 issues left to fix 1 of which can be done here:
1) results are massive https://github.com/OpenFreeEnergy/alchemiscale/issues/326 this will be fixed in alchemiscale
2) the results objects don't contain the input components so we need to change how we extract them. Dotson advises that getting the transformation from alchmiscale and checking the components is the most reliable way to do this across protocols, I'll add this in then we should be good to merge!

hmacdope:
Thanks for the update @jthorton! Lets take this off the 0.5.2 milestone until we get the new alchemiscale release that will unlock this.

dotsdl:
Closing for now; will be taken as prior art and superseded by a fresh PR reflecting current state of the library.

PR #1124, opened 2024-07-01
Add option to zoom GIFVisualizer into binding site

mariacm12:
## Description
The current implementation of `GIFVisualizer.visualize()` sometimes have issues with centering the final view of the output (which is a problem for the .gif file). I added the `zoom_view` parameter to `GIFVisualizer`, which if True, will center the .gif view to the binding site. 

I placed the centering command right before generating the _Session 6_ PyMOL output, so it only affect this and the .gif file. I could move it so it centers other outputs, but don't think it's important for a PyMOL file to be initially centered.

## Todos
- [x] Review.

## Status
- [x] Ready to go


hmacdope (approved):
LGTM, thanks @mariacm12

PR #1151, opened 2024-07-31
Add  CLI and required changes for running automated training against CDD. 

### LINKED ISSUE #1083: Allow ML model specification to pull from an S3 bucket for multiple endpoints.
The other side of #1082 

From the repo perspective we need to be able to read multiple endpoints from an S3 bucket of the form `https://asap-discovery-ml-weights.s3.amazonaws.com/training/<endpoint>/<endpoint>_<isodate>/<endpoint>_<isodate>_<model_index>.th` and the associated configuration files. 

This could look something like 

```python
ASAPMLModelRegistry = RemoteMLModelRegistry.from_yaml(asap_remote_models_yaml)


...


Where `asap_remote_models_yaml`

has the form 

```yaml 
targets:
   - Blah:
      endpoints:
        - Blah 
        - Blah2
```

Such that models can be pulled down on the repo end with something like 

```python 
model = RemoteMLModelRegistry.get_latest_from_target_and_endpoint(target, endpoint)
```

## Notes 

* Should have a way of merging models between the two types of registry?

### LINKED ISSUE #1087: Validate that ML-scaleup pipeline works end-to-end
Close this issue when it works in production!

### LINKED ISSUE #1089: Allow batch creation date to be pulled by CDD API
Adjust CDD API download so that we can pull `"Batch Created Date"` directly, very useful for ML training

hmacdope:
Fixes #1087
Fixes #1083
Fixes #1089

Will need a review from @kaminow as has touched a few points in the ML training. 


hmacdope:
@kaminow would you mind taking a first rough pass? Need to sand off some edges, but I think the basic ideas are in place.

asapdiscovery-data/asapdiscovery/data/util/utils.py
```diff
@@ -341,6 +341,126 @@ def cdd_to_schema(cdd_csv, out_json=None, out_csv=None):
...
+    if len(missing_cols) > 0:
+        raise ValueError(f"Required columns not present in CSV file: {missing_cols}. ")
+
+    # Make extra sure nothing snuck by
+    idx = df["smiles"].isna()
+    logging.debug(f"Removing {idx.sum()} entries with no SMILES", flush=True)
+    df = df.loc[~idx, :]
+
+    # Fill semi-qunatitative data with the mean of others WHAT TO DO FOR regular scalar?
+    df.loc[df["semiquant"], target_prop] = df.loc[~df["semiquant"], target_prop].mean()
```
kaminow, line 394: in the original function, this was just for the experimental error. don't need to fill anything for the actual values, since we should also be checking if they're semi-quantitative when we use them later

asapdiscovery-data/asapdiscovery/data/util/utils.py
```diff
@@ -341,6 +341,126 @@ def cdd_to_schema(cdd_csv, out_json=None, out_csv=None):
     return compounds
 
 
+def cdd_to_schema_v2(
```
kaminow, line 344: not sure if you're already planning on updating this, but as it stands this will break my pipeline. I think I'd need:
* ability to pass a list of values to `target_prop`
* boolean flags for also getting CI values, in-range values, and uncertainty values (see keys I have in `reqd_cols` in the original version of this function)

the second bullet could be handled also as lists, so some keys in `target_prop` can require the extra information while some may not need it

kaminow (commented):
a couple of things I noticed that'll cause issues, but so far this seems like it shouldn't affect me much

asapdiscovery-data/asapdiscovery/data/util/utils.py
hmacdope: I think we can just let these live side by side for now, and then we can reconcile them in the future?

asapdiscovery-ml/asapdiscovery/ml/trainer.py
```diff
@@ -729,13 +736,23 @@ def train(self):
...
-                ).float()
-                uncertainty = torch.tensor(
-                    [[pose[f"{self.target_prop}_stderr"]]],
-                    device=self.device,
-                ).float()
+
+                if self.has_range:
+                    in_range = torch.tensor(
+                        [[pose[f"{self.target_prop}_range"]]], device=self.device
+                    ).float()
```
hmacdope, line 743: This is worth noting @kaminow, under current assumptions, we require there to be a `range` and `stderr` tag bit of data alongside. Also these names are non-configurable. I have added a bypass but we can make more flexible in a seperate PR?

asapdiscovery-data/asapdiscovery/data/util/utils.py
hmacdope: So this line can be safely removed?

asapdiscovery-data/asapdiscovery/data/services/aws/s3.py
```diff
@@ -124,6 +124,31 @@ def push_file(
...
+        Parameters
+        ----------
+        path
+            Path to directory on local filesystem to push.
+        location
+            Location in the S3 bucket to place object relative to ``self.prefix``;
+            should not contain a leading ``/``.
+        """
+        if location is None:
+            location = os.path.basename(path)
```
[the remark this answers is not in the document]
hmacdope: Good point.

PR #1081, opened 2024-06-05
Add ability to have multiple loss functions

kaminow:
We want to be able to incorporate multiple loss functions into our training, eg standard affinity target loss + loss penalty for predicting values outside a normal pIC50 range. This PR makes that infrastructure change, and also improves how we're tracking loss values and model predictions over training. Also introduce some new loss functions.

- [x] Allow multiple loss functions in `Trainer`/CLI
- [x] Add range penalty loss function
- [x] Add loss function for correctly ranking poses
- [x] Add pose ranking to the dataset loading (labels for above loss function in training)
- [x] Add `asapdiscovery.ml.schema.TrainingPrediction` and `asapdiscovery.ml.schema.TrainingPredictionTracker` to organize how we track predictions
- [x] New `ml.viz` module with common visualization stuff (loss curves, prediction plots, etc)

PR #1152, opened 2024-07-31
Update mtenn pin for upcoming breaking release.

kaminow:
Going to cut an `mtenn` release soon that will break the current code, so pin to <= current version. Code is being updated to work with the new `mtenn` version in #1081.

### DIFF
--- a/devtools/conda-envs/asapdiscovery-macOS-12.yml
+++ b/devtools/conda-envs/asapdiscovery-macOS-12.yml
@@ -64,7 +64,7 @@ dependencies:
   - dgl <=2.0.0
   - dgllife
   - pooch
-  - mtenn >=0.5.1,<=0.5.2
+  - mtenn =0.5.2
   - wandb
 
   # md
--- a/devtools/conda-envs/asapdiscovery-macOS-12.yml
+++ b/devtools/conda-envs/asapdiscovery-macOS-12.yml
@@ -64,7 +64,7 @@ dependencies:
   - dgl <=2.0.0
   - dgllife
   - pooch
-  - mtenn >=0.5.1
+  - mtenn >=0.5.1,<=0.5.2
   - wandb
 
   # md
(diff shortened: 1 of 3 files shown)


devtools/conda-envs/asapdiscovery-macOS-12.yml
```diff
@@ -64,7 +64,7 @@ dependencies:
   - dgl <=2.0.0
   - dgllife
   - pooch
-  - mtenn >=0.5.1
+  - mtenn >=0.5.1,<=0.5.2
```
ijpulidos, line 67:
```suggestion
  - mtenn =0.5.1
```

devtools/conda-envs/asapdiscovery-macOS-latest.yml
```diff
@@ -63,7 +63,7 @@ dependencies:
   - dgl <=2.0.0
   - dgllife
   - pooch
-  - mtenn >=0.5.1
+  - mtenn >=0.5.1,<=0.5.2
```
ijpulidos, line 66:
```suggestion
  - mtenn =0.5.1
```

devtools/conda-envs/asapdiscovery-ubuntu-latest.yml
```diff
@@ -64,7 +64,7 @@ dependencies:
   - dgl <=2.0.0
   - dgllife
   - pooch
-  - mtenn >=0.5.1
+  - mtenn >=0.5.1,<=0.5.2
```
ijpulidos, line 67:
```suggestion
  - mtenn =0.5.1
```

ijpulidos (commented):
Made some minor suggestions. Wouldn't it be better in terms of versioning for the new `mtenn` release to be `0.6.x`? I know we are in the `0.x` branch so it is not that important, but I think it would make it a little easier for the users to expect that `0.6` breaks compatibility with `0.5`. If that makes any sense.

mikemhenry:
+1 to doing a release where the next version of mtenn is 0.6 and doing a `=` pin since we know there are no more 0.5.x releases and with the changes in mtenn we will need to do a code update

kaminow:
that makes sense, thanks! updated `mtenn` pin to `=0.5.2` (most recent version on `conda-forge`)

mikemhenry (approved):

Issue #560, opened 2023-10-27
Improve Dask backend in MCS selector by breaking up computation further 


hmacdope:
Currently we don't have parallelism over complexes, this is possible by computing their MCS match score in a seperate step that can also be split up over dask workers.

hmacdope:
Ok I have done some testing and benchmarking on this one with the result that using a dask-based parallelism scheme is not beneficial

This is because the  inner loop invariant `mcss` SWIG object is not pickelable meaning it has to be reconstructed in each dask work unit.  Compare the two approaches below. 

## 
```python
  for ligand in ligands:
            pattern_query = oechem.OEQMol(ligand.to_oemol())
            pattern_query.BuildExpressions(atomexpr, bondexpr)
            mcss = oechem.OEMCSSearch(pattern_query)
            mcss.SetMCSFunc(oechem.OEMCSMaxAtomsCompleteCycles()) # MCSS object is loop invariant in inner loop

            sort_args = []
            for complex in complexes:
                complex_mol = complex.ligand.to_oemol()
                # MCS search
                try:
                    mcs = next(iter(mcss.Match(complex_mol, True)))
                    sort_args.append((mcs.NumBonds(), mcs.NumAtoms()))
                except StopIteration:  # no match found
                    sort_args.append((0, 0))
            ...
   ```
   
   ```python 
  for ligand in ligands:
            for complex in complexes:
                res.append(dask.delayed(mcs_search(lig, complex)) # MCSS object reconstructed each time 
            ...
  
   ```
   
   Both are M x N scaling but the overhead involved in reconstructing the GIL holding MCSS object each time outweighs the benefits.  At  least that was  the case in the benchmarking I did. 
   
  Instead  we should remove the Dask API as an option to the MCS Selector altogether.

PR #1157, opened 2024-08-01
Improve MCS performance

hmacdope:
Improves the MCS performance by allowing the use of a faster approximate method. 

We couldn't multiprocess or use dask due to limitations outlined in #560. 

We may need to investigate RDKit based MCS searching for a parallel component, but may run into the same issue there. 

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/operators/selectors/mcs_selector.py
+++ b/asapdiscovery-data/asapdiscovery/data/operators/selectors/mcs_selector.py
@@ -214,7 +214,14 @@ class MCSSelector(SelectorBase):
                 pairs.append(pair_cls(ligand=ligand, complex=complexes[0]))
                 continue
 
-
+            pattern_query = oechem.OEQMol(ligand.to_oemol())
+            pattern_query.BuildExpressions(atomexpr, bondexpr)
+            if self.approximate:
+                mcs_stype = oechem.OEMCSType_Approximate
+            else:
+                mcs_stype = oechem.OEMCSType_Exhaustive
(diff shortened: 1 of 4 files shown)


hmacdope:
This is about a 5x improvement from my benchmarking but still not ideal.

hmacdope:
@JenkeScheen can I grab a quick review?

asapdiscovery-data/asapdiscovery/data/operators/selectors/mcs_selector.py
```diff
@@ -208,7 +212,11 @@ def _select(
 
             pattern_query = oechem.OEQMol(ligand.to_oemol())
             pattern_query.BuildExpressions(atomexpr, bondexpr)
-            mcss = oechem.OEMCSSearch(pattern_query)
+            if self.approximate:
+                mcs_stype = oechem.OEMCSType_Approximate
```
JenkeScheen, line 216: should log this as `warning`, possibly referring to docs: https://docs.eyesopen.com/toolkits/python/oechemtk/patternmatch.html

JenkeScheen (approved):
Looks good, thanks for the improvement! I think this will help for most cases, but we'll definitely still need multithreading. Currently 3000 compounds takes ~5 days on 32 jobs, with this PR that would be ~1 day. We'd like to be able to large-dock 30-50k compounds at the same time.

Thanks!!

asapdiscovery-data/asapdiscovery/data/operators/selectors/mcs_selector.py
hmacdope: Good idea.

PR #1177, opened 2024-08-22
Add outpaths to lig transfer docking

### LINKED ISSUE #1153: The current saving format of asapdiscovery-simulation needs to be more general
I want to implement a minimization and re-docking step for the `ligand-transfer-docking` pipeline, but the current `outpath` format wouldn't allow for me to do a second docking step. Specifically, right now the output format looks like `out_dir/input.unique_name/minimized.pdb`, but I would need to save the files in a single "raw" directory to be able to read them back into the workflow. 

Can we implement an option to save the outputs in a raw directory?

hmacdope:
Fixes #1153 

I think we could already do this so I added implementation @mariacm12 

### DIFF
--- a/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
+++ b/asapdiscovery-simulation/asapdiscovery/simulation/simulate.py
@@ -34,6 +34,7 @@ from pydantic import (
 )
 from rdkit import Chem
 from tqdm import tqdm
+import shutil
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


hmacdope:
@mariacm12 would you be able to test that this works for you?

mariacm12:
Is the fix supposed to save all MD results in a single folder? I tested it and the files are still saved in the `md/<unique_name>/minimized.pdb format`.

hmacdope:
@mariacm12 this now collects all the MD results in a flat structure if requested. It currently saves all the files but you can easily add a step in workflow to delete ones you don't want.

hmacdope:
Should be able to combine with https://github.com/asapdiscovery/asapdiscovery/pull/1181 to just pull out the ones you want at the next step.

mariacm12 (approved):
Awesome!! The last edit works well.

PR #1181, opened 2024-08-28
Add custom glob to structure dir

hmacdope:
## Description
Allows specifying a custom glob for `StructureDirFactory`


hmacdope:
@mariacm12 can you review here?

mariacm12:
Looks good! This would be to use `StructureDirFactory` instead of `MetaStructureFactory`?

hmacdope:
This just might be useful if you want to pull structures of a certain type out of the `StructureDir` eg only `minimized_blah.pdb`. But I realise now that as you are only minimizing there will only ever be the PDB files you want. Regardless I think this could be useful.

mariacm12 (approved):
Ok!

PR #1170, opened 2024-08-15
[DNM] OpenFE 1.0 + Alchemiscale 0.5 compat

JenkeScheen:
## Developers certificate of origin
- [x] I certify that this contribution is covered by the MIT license as defined in our [LICENSE](https://github.com/choderalab/asapdiscovery/blob/main/LICENSE) and adheres to the [**Developer Certificate of Origin**](https://developercertificate.org/).

PR #1184, opened 2024-09-02
[DNM] `Alchemy predict` chuck out disconnected

JenkeScheen:
## Description
In cases where networks are disconnected (sometimes, especially at scale, some ligands just don't want to work) we need to be able to make predictions anyway while discarding the failed ligands. In `cinnabar` this is currently not possible - this PR implements a workaround.

## Todos
- [ ] finalize tests - blocked by #1170 

## Status
- [ ] Ready to go

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
@@ -5,8 +5,8 @@ import pandas as pd
 import rich
 
 if TYPE_CHECKING:
-    from asapdiscovery.data.schema.ligand import Ligand
     from asapdiscovery.alchemy.schema.fec import FreeEnergyCalculationNetwork
+    from asapdiscovery.data.schema.ligand import Ligand
     from cinnabar import FEMap
 
 


asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
```diff
@@ -259,3 +259,59 @@ def report_alchemize_clusters(alchemical_clusters, outsiders):
...
+    from asapdiscovery.alchemy.schema.fec import (
+        AlchemiscaleResults,
+        FreeEnergyCalculationNetwork,
+    )
+    from rich.padding import Padding
+
+    fe_map_nx = fe_map.graph
+    subnetworks_nodenames = sorted(  # split the network into subnetworks
+        nx.strongly_connected_components(fe_map_nx), key=len, reverse=True
+    )
```
jthorton, line 287: If we could just extract the largest subnetwork from the graph we might be able to convert directly back to an FEmap using `from_networkx` which would make this a lot more simple. https://github.com/OpenFreeEnergy/cinnabar/blob/c140fea77d4019119ed40acd1a699b92ed6bbf10/cinnabar/femap.py#L174-L184

JenkeScheen: the commit that added that feature isn't in the conda-installable conda-forge release of `cinnabar`. is there a reasonable way of pinning our `cinnabar` dep to that most recent commit? if it's too hacky I think we might be better off sticking with this approach as it's a bit more controllable than a magic 'from_networkx()` method imo.

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
```diff
@@ -259,3 +259,59 @@ def report_alchemize_clusters(alchemical_clusters, outsiders):
         ]
     )
     return alchemical_cluster_sizes, outsider_cluster_sizes, alchemical_num_in_clusters
+
+
+def cinnabar_femap_is_connected(fe_map):
```
jthorton, line 264: Nice the naming is much better! Can we add type hits though please!

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
```diff
@@ -259,3 +259,59 @@ def report_alchemize_clusters(alchemical_clusters, outsiders):
...
+    """Checks whether the provided femap is connected. Convenience function to make function
+    naming clearer compared to cinnabar nomenclature."""
+    return fe_map.check_weakly_connected()
+
+
+def cinnabar_femap_get_largest_subnetwork(fe_map, result_network, console):
```
jthorton, line 270: type hints here as well please!

asapdiscovery-alchemy/asapdiscovery/alchemy/cli/utils.py
jthorton: Good point lets stick with the current method until cinnabar gets updated!

jthorton (approved):
Great job!

PR #1243, opened 2024-10-11
Add filter on fluorescence experimental errors

### LINKED ISSUE #1234: When IC50 stderrors are very large pIC50 and IC50 values can be clipped to very large numbers.
Using stderr based rounding for IC50s and pIC50s here: https://github.com/asapdiscovery/asapdiscovery/blob/main/asapdiscovery-data/asapdiscovery/data/util/utils.py#L951-L956

Can make values very large if stderrs are large. We should work on some kind of fix.

kaminow:
Add filter on fluorescence experimental errors to prevent overly large errors from rounding the readout values to 0/10. Closes #1234.

### DIFF
--- a/asapdiscovery-data/asapdiscovery/data/util/utils.py
+++ b/asapdiscovery-data/asapdiscovery/data/util/utils.py
@@ -831,6 +831,7 @@ def parse_fluorescence_data_cdd(
     assay_name="ProteaseAssay_Fluorescence_Dose-Response_Weizmann",
     dG_T=298.0,
     cp_values=None,
+    pic50_stderr_filt=10.0,
 ):
     """
     Filter a dataframe of molecules to retain those specified. Required columns are:
@@ -874,6 +875,9 @@ def parse_fluorescence_data_cdd(
         Cheng-Prussoff equation. These values are assumed to be in the same
         concentration units. If no values are passed for this, pIC50 values
         will be used as an approximation of the Ki
+    pic50_stderr_filt : float, default=10.0
+        Max allowable standard error in pIC50 units. Overly large errors lead to rounded
+        values that don't make sense, so set anything that will cause issues to NaN
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


kaminow:
@hmacdope I think this is ready, lmk if you think there's anything else we want to do. rather than removing stuff here I just set the IC50 and pIC50 values to NaNs since they should be removed later in the pipeline anyway, but we can change if you think we should just remove the offending rows

asapdiscovery-data/asapdiscovery/data/util/utils.py
```diff
@@ -952,21 +956,28 @@ def parse_fluorescence_data_cdd(
...
-                    pIC50, uncertainty=pIC50_stderr, sep=tuple, output_type=str
-                )  # strings
-            except ModuleNotFoundError:
-                # Don't round
-                pass
+            # Check error for filtering
+            if pIC50_stderr > pic50_stderr_filt:
+                IC50 = "nan"
+                pIC50 = "nan"
+                # @HMO do we want to leave the stderr values as they are for posterity
```
hmacdope, line 963: Log them as a warning I think would be good.

asapdiscovery-data/asapdiscovery/data/util/utils.py
```diff
@@ -952,21 +956,28 @@ def parse_fluorescence_data_cdd(
...
+                #  or set them to NaN as well?
```
hmacdope, line 964: Set them to `nan` I think is the right move.

asapdiscovery-ml/asapdiscovery/ml/cli_mlops.py
```diff
@@ -774,13 +781,20 @@ def mlops():
     is_flag=True,
     help="Run in test mode, no S3 push or WandB logging to main project",
 )
+@click.option(
+    "--pic50-stderr-filt",
```
hmacdope, line 785: Nice, thanks.

hmacdope (changes requested):
Great work! I say we leave in the backup clipping L488-498, what do you think?

hmacdope:
Also @kaminow you will need to merge with `upstream/main`

kaminow:
@hmacdope I think ready for another review!

also, which file are you referring to here?
> Great work! I say we leave in the backup clipping L488-498, what do you think?

hmacdope:
> @hmacdope I think ready for another review!
> 
> also, which file are you referring to here?
> 
> > Great work! I say we leave in the backup clipping L488-498, what do you think?

asapdiscovery-ml/asapdiscovery/ml/cli_mlops.py , but should be all good.

hmacdope (approved):

PR #1253, opened 2024-10-17
Adaptive settings: adaptive sampling time and padding size

JenkeScheen:
fixes https://github.com/asapdiscovery/asapdiscovery/issues/1227

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
@@ -348,7 +348,7 @@ class _FreeEnergyBase(_SchemaBase):
         description="Settings for simulation control, including lengths and writing to disk.",
     )
     adaptive_settings: Optional[AdaptiveSettings] = Field(
-        None,
+        AdaptiveSettings(),
         description="Run adaptive settings depending on e.g. expected edge reliability or system phase.",
     )
     protocol: Literal["RelativeHybridTopologyProtocol"] = Field(
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
@@ -408,7 +408,8 @@ class FreeEnergyCalculationNetwork(_FreeEnergyBase):
                     sys_b_dict, name=f"{mapping.componentB.name}_{leg}"
                 )
 
-                # make the OpenFE protocol for this edge; double the simulation time if requested
+                ## make the OpenFE protocol for this edge
+                # double the simulation time if requested
                 protocol_copy = self.copy(deep=True)
                 if (
                     self.adaptive_settings.adaptive_sampling
@@ -418,6 +419,17 @@ class FreeEnergyCalculationNetwork(_FreeEnergyBase):
                     protocol_copy.simulation_settings.production_length *= (
                         self.adaptive_settings.adaptive_sampling_multiplier
                     )
+                # adjust solvent padding per phase if requested
+                if self.adaptive_settings.adaptive_solvent_padding:
+                    if leg == "solvent":
+                        protocol_copy.solvation_settings.solvent_padding = (
+                            self.adaptive_settings.solvent_padding_solvated
+                        )
+                    else:
+                        protocol_copy.solvation_settings.solvent_padding = (
+                            self.adaptive_settings.solvent_padding_complex
+                        )
+                print(protocol_copy.solvation_settings.solvent_padding)
                 protocol_openfe = protocol_copy.to_openfe_protocol()
 
                 # set up the transformation
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
@@ -21,7 +21,7 @@ from openfe.protocols.openmm_utils.omm_settings import (
 from openfe.setup.atom_mapping import lomap_scorers, perses_scorers
 from openff.models.types import FloatQuantity
 from openff.units import unit as OFFUnit
-from pydantic import BaseSettings, Field
+from pydantic import Field
 
 from ._util import check_ligand_series_uniqueness_and_names
 from .base import _SchemaBase, _SchemaBaseFrozen
@@ -29,23 +29,7 @@ from .network import NetworkPlanner, PlannedNetwork
 
 if TYPE_CHECKING:
     from asapdiscovery.data.schema.ligand import Ligand
-
-
-class AlchemiscaleSettings(BaseSettings):
-    """
-    General settings class to capture Alchemiscale credentials from the environment.
-    """
-
-    ALCHEMISCALE_ID: str = Field(
-        ..., description="Your personal alchemiscale ID used to login."
-    )
-    ALCHEMISCALE_KEY: str = Field(
-        ..., description="Your personal alchemiscale Key used to login."
-    )
-    ALCHEMISCALE_URL: str = Field(
-        "https://api.alchemiscale.org",
-        description="The address of the alchemiscale instance to connect to.",
-    )
+    from gufe.mapping import LigandAtomMapping
 
 
 class SolventSettings(_SchemaBase):
@@ -284,8 +268,7 @@ class _FreeEnergyBase(_SchemaBase):
     )
     forcefield_settings: settings.OpenMMSystemGeneratorFFSettings = Field(
         settings.OpenMMSystemGeneratorFFSettings(
-            small_molecule_forcefield="openff-2.2.0",
-            nonbonded_cutoff=0.9 * OFFUnit.nanometer,
+        small_molecule_forcefield="openff-2.2.0.offxml"
         ),
         description="The force field settings used to parameterize the systems.",
     )
@@ -413,6 +396,8 @@ class FreeEnergyCalculationNetwork(_FreeEnergyBase):
         Returns:
             An openfe.AlchemicalNetwork created from the schema.
         """
+        import copy
+
         transformations = []
         # do all openfe conversions
         ligand_network = self.network.to_ligand_network()
@@ -421,6 +406,16 @@ class FreeEnergyCalculationNetwork(_FreeEnergyBase):
 
         # build the network
         for mapping in ligand_network.edges:
+            # returns the name of the ff if we have no bespoke parameters
+            ff_string = self._inject_bespoke_parameters(edge=mapping)
+            # make a copy of the protocol and add the bespoke force field
+            edge_protocol = copy.deepcopy(protocol)
+            # make the settings editable
+            edge_protocol._settings = edge_protocol._settings.unfrozen_copy()
+            edge_protocol._settings.forcefield_settings.small_molecule_forcefield = (
+                ff_string
+            )
+
             for leg in ["solvent", "complex"]:
                 sys_a_dict = {"ligand": mapping.componentA, "solvent": solvent}
                 sys_b_dict = {"ligand": mapping.componentB, "solvent": solvent}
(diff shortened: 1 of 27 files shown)


JenkeScheen:
also added in some speed improvement suggestions by OpenFE, and added a class for adaptive settings that is a draft version of https://github.com/asapdiscovery/asapdiscovery/issues/1254, but good enough for now. @jthorton can you have a look over these changes as it's mostly pydantic finaggling?

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -359,11 +406,35 @@ def to_alchemical_network(self) -> openfe.AlchemicalNetwork:
...
+                        protocol_copy.solvation_settings.solvent_padding = (
+                            self.adaptive_settings.solvent_padding_solvated
+                        )
+                    else:
+                        protocol_copy.solvation_settings.solvent_padding = (
+                            self.adaptive_settings.solvent_padding_complex
+                        )
+                protocol_openfe = protocol_copy.to_openfe_protocol()
+
+                # set up the transformation
```
jthorton, line 432: Lets move all of this logic to a function on the `AdaptiveSettings` class that will make this section easier to read as this will get more complicated as the class evolves.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -340,14 +379,22 @@ def to_alchemical_network(self) -> openfe.AlchemicalNetwork:
...
+
+        # get the atom mapping scorer in case we need to double simulation time for potentially challenging edges
+        if self.network.scorer == "default_lomap":
+            scorer = lomap_scorers.default_lomap_score
+        elif self.network.scorer == "default_perses":
+            scorer = perses_scorers.default_perses_scorer
+        else:
+            raise ValueError(
+                f"Atom mapping scorer {self.network.scorer} not recognized; use one of `default_lomap`, `default_perses`."
+            )
```
jthorton, line 391: Move to a function on the `AdaptiveSettings` class.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -60,6 +61,105 @@ def to_solvent_component(self) -> gufe.SolventComponent:
         return gufe.SolventComponent(**self.dict(exclude={"type"}))
 
 
+class AdaptiveSettings(_SchemaBase):
+    """
+    A settings class to encode settings for adaptive settings. These were recommended by OpenFE.
+    """
+
+    type: Literal["AdaptiveSettings"] = "AdaptiveSettings"
+    adaptive_sampling: bool = Field(
```
hmacdope, line 70: This should go away, instantiating the class and attatching to `FreeEnergyCalculationFactory` should signal use of adaptive sampling.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -225,9 +325,14 @@ class _FreeEnergyBase(_SchemaBase):
         MultiStateSimulationSettings(
             equilibration_length=1.0 * OFFUnit.nanoseconds,
             production_length=5.0 * OFFUnit.nanoseconds,
+            time_per_iteration=2.5 * OFFUnit.picoseconds,
         ),
         description="Settings for simulation control, including lengths and writing to disk.",
     )
+    adaptive_settings: AdaptiveSettings = Field(
+        AdaptiveSettings(),
```
hmacdope, line 333: field should be optional with default None, and then the adaptive_sampling field on AdaptiveSampling class should go away.

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/test_fec_schema.py
```diff
@@ -229,6 +229,102 @@ def test_fec_to_openfe_protocol():
...
+            edge.protocol.settings.simulation_settings.production_length.magnitude
+        )
+    # test while sorting because edge order is scrambled randomly
+    assert sorted(reference_adaptive_sampling_times) == sorted(sampling_lengths)
+
+    ## now repeat but with disabling adaptive settings. This should make the sampling time
+    # the same for all edges.
+    factory_2 = FreeEnergyCalculationFactory()
+    factory_2.adaptive_settings.adaptive_sampling = False
+    reference_adaptive_sampling_times = [5.0, 5.0, 5.0, 5.0, 5.0, 5.0]
```
hmacdope, line 264: currently FreeEnergyCalculationFactory is created it is instantiated with AdaptiveSampling with defaults which has adaptive_sampling  set to true, and it is applying the modifier. Then you are setting to false but the modifier has already been applied, which is why the sampling has already been modified.

hmacdope (changes requested):
Copy over from slack for posterity.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -60,6 +61,115 @@ def to_solvent_component(self) -> gufe.SolventComponent:
...
+        return protocol
+
+    def apply_settings(
+        self,
+        edge_protocol: openfe.protocols.openmm_rfe.RelativeHybridTopologyProtocol,
+        network_scorer: str,
+        mapping: "LigandAtomMapping",
+        leg: str,
+        base_protocol: openfe.protocols.openmm_rfe.RelativeHybridTopologyProtocol,
+    ):
```
jthorton, line 148: The list of things to pass in is probably going to grow in time should we just pass in the `FreeEnergyCalculationNetwork` and then we can extract everything we need?

jthorton, line 148: Type hint for the return.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -225,9 +335,14 @@ class _FreeEnergyBase(_SchemaBase):
         MultiStateSimulationSettings(
             equilibration_length=1.0 * OFFUnit.nanoseconds,
             production_length=5.0 * OFFUnit.nanoseconds,
+            time_per_iteration=2.5 * OFFUnit.picoseconds,
```
jthorton, line 338: I would keep this at the default value while we look into large differences from the test you ran.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -60,6 +61,115 @@ def to_solvent_component(self) -> gufe.SolventComponent:
...
+            raise ValueError(
+                f"Atom mapping scorer {scorer_method} not recognized; use one of `default_lomap`, `default_perses`."
+            )
+        if scorer(mapping) < self.adaptive_sampling_threshold:
+            protocol._settings.simulation_settings.production_length = (
+                base_sampling_length * self.adaptive_sampling_multiplier
+            )
+        return protocol
+
+    def get_adapted_solvent_protocol(self, leg, protocol):
```
jthorton, line 122: type hints.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -60,6 +61,115 @@ def to_solvent_component(self) -> gufe.SolventComponent:
...
+        1.0 * OFFUnit.nanometer,
+        description="The solvent padding (in nm) to use for the complex phase of each edge.",
+    )
+    solvent_padding_solvated: FloatQuantity["nanometer"] = Field(
+        1.2 * OFFUnit.nanometer,
+        description="The solvent padding (in nm) to use for the solvated phase of each edge.",
+    )
+
+    def get_adapted_sampling_protocol(
+        self, scorer_method, mapping, protocol, base_sampling_length
```
jthorton, line 96: type hints

jthorton (changes requested):
Great job @JenkeScheen, just some very minor changes needed!

asapdiscovery-alchemy/asapdiscovery/alchemy/tests/test_fec_schema.py
```diff
@@ -229,6 +230,117 @@ def test_fec_to_openfe_protocol():
     )
 
 
+def test_fec_adaptive_sampling(tyk2_ligands, tyk2_protein):
```
jthorton, line 233: Great job with the tests!

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -225,9 +335,14 @@ class _FreeEnergyBase(_SchemaBase):
         MultiStateSimulationSettings(
             equilibration_length=1.0 * OFFUnit.nanoseconds,
             production_length=5.0 * OFFUnit.nanoseconds,
+            time_per_iteration=2.5 * OFFUnit.picoseconds,
         ),
         description="Settings for simulation control, including lengths and writing to disk.",
     )
+    adaptive_settings: Optional[AdaptiveSettings] = Field(
```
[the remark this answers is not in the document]
JenkeScheen: good catch, I'd intended to enable by default but must have disabled during test building.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
JenkeScheen: agreed, although for now I just added a comment so that we can migrate once we're building at this spot again. It's a bit complicated to derive all those parameters just from the `FreeEnergyCalculationNetwork` alone.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
```diff
@@ -60,6 +61,123 @@ def to_solvent_component(self) -> gufe.SolventComponent:
...
+        protocol's sampling time and not the provided edge protocol sampling time as a base value.
+
+        Returns the adjusted OpenFE Protocol.
+        """
+        if scorer_method == "default_lomap":
+            scorer = lomap_scorers.default_lomap_score
+        elif scorer_method == "default_perses":
+            scorer = perses_scorers.default_perses_scorer
+        else:
+            raise ValueError(
```
hmacdope, line 117: If this is on by default, then any scoring method other than these two will cause an exception. Are we sure that is what we want? We could just return protocol with a warning, or is there anything else stopping us from using other scorers?

hmacdope (changes requested):
Just one query about scorers as this currently restricts to just these two.

asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
JenkeScheen: it's a fair point, this check might not be needed at all because we're probably checking for this upstream. There aren't any scorers available in OpenFE currently, but removing the check here would make implementing any upcoming scorers less painful.

JenkeScheen: there are only two scorers available in OpenFE - any new scorers would need to be added here. I'll have to think about whether it's worth automating this step, also because I've heard through the grapevine that OpenFE will be porting mapping scorings directly through their transformations soon anyway.

JenkeScheen: @jthorton thoughts?

jthorton: I would do this for now and then remove this once the LigandNetworks correctly stores the edge score which should be [fixed soon](https://github.com/OpenFreeEnergy/gufe/pull/369) as you will always want to use the same scorer which planned the network.

jthorton (approved):
LGTM!

hmacdope (approved):

PR #1261, opened 2024-10-23
`alchemy` add hard stop functionality

JenkeScheen:
Solves https://github.com/asapdiscovery/asapdiscovery/issues/1196. This will allow us to _actually_ stop networks that have been submitted e.g. accidentally. Up to this PR, any running tasks would have 'escaped' cancelling because the `stop` method only set `waiting` to non-actioned. With this PR we hard a hard stop extension to the CLI so that we can prevent unnecessary compute.

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
@@ -760,7 +760,12 @@ def prioritize(network_key: str, weight: float):
     help="The network key of the network to be stopped. This can be found by running e.g. `asap-alchemy status -a`.",
     required=True,
 )
-def stop(network_key: str):
+@click.option(
+    "--hard",
+    is_flag=True,
+    help="If used, all waiting and running tasks will be deleted instead of un-actioned. Warning: these tasks will not be retrievable/re-runnable.",
+)
+def stop(network_key: str, hard: bool = False):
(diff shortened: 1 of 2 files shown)


asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
```diff
@@ -773,9 +778,13 @@ def stop(network_key: str):
     print_header(console)
 
     client = AlchemiscaleHelper.from_settings()
+    if hard:
+        console.print(
+            f"Warning: deleting all running/waiting tasks on network {network_key}. These will not be retrievable/re-runnable!"
+        )
```
hmacdope, line 784: Confirm with user

```suggestion
        )
inp = input("continue? (y/n)")
if inp == "y": 
...
elif inp == "n":
    print("aborting")
    return
else: 
    raise ValueError("option not recognised")
```

JenkeScheen: nice!

JenkeScheen: this might break the test though

JenkeScheen: hmm nop!

hmacdope (changes requested):
Also needs a test

hmacdope (approved):

PR #1275, opened 2025-01-17
Add extra training options

kaminow:
Add some extra `Trainer` options to help with running experiments. Currently planned/implemented in this PR:
- [x] `weight_decay`: Weight decay loss term implemented in `Trainer` (this is probably unnecessary as the `torch` optimizers have this implemented as well)
- [x] `batch_norm`: For mini-batches, divide the final gradients by the size of the mini-batch before taking an optimizer step


hmacdope:
@kaminow failures should be resolved if you pull in upstream

PR #1297, opened 2026-04-09
Update conda envs for alchemy (and others) for parity with deployed alchemiscale instance

dotsdl:
Sets:
- openfe =1.8.0
- gufe =1.7.1
- openmm =8.2.0


dotsdl (commented):
I'm satisfied with this; should now be well-positioned to use newly-deployed `alchemiscale` instance.

Issue #1295, opened 2026-03-28
Remove vestigial `from __future__ import annotations` imports


dotsdl:
## Summary

With the pydantic v2 migration (#1282) and Python 3.12+ target, `from __future__ import annotations` is no longer necessary in these 8 files:

- `asapdiscovery-data/asapdiscovery/data/schema/schema_base.py`
- `asapdiscovery-data/asapdiscovery/data/schema/complex.py`
- `asapdiscovery-data/asapdiscovery/data/operators/selectors/selector.py`
- `asapdiscovery-data/asapdiscovery/data/operators/selectors/pairwise_selector.py`
- `asapdiscovery-data/asapdiscovery/data/operators/selectors/mcs_selector.py`
- `asapdiscovery-data/asapdiscovery/data/services/fragalysis/fragalysis_reader.py`
- `asapdiscovery-modeling/asapdiscovery/modeling/schema.py`
- `asapdiscovery-ml/asapdiscovery/ml/config.py`

The `str | None` union syntax is native in Python 3.12+, and pydantic v2 handles forward references natively. These imports are harmless but vestigial.

## Notes

Removing `from __future__ import annotations` changes how pydantic v2 receives type annotations (live types vs strings), so each file should be tested after removal to catch any evaluation-order edge cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dotsdl:
Tried to action this in a cleanup PR and found the premise doesn't hold for Python 3.12 (the supported version). Flagging so this can be re-triaged.

## The gotcha

The issue states pydantic v2 handles forward references natively, so `from __future__ import annotations` should be removable. That's only true for Pydantic `Field()` declarations in the class body. Ordinary method parameter/return annotations still evaluate eagerly on 3.12, and PEP 649's lazy evaluation only lands in 3.14.

Minimal repro on 3.12:

```python
class Bar:
    def baz(self, other: Bar) -> bool:
        return True
```

```
NameError: name 'Bar' is not defined
```

## All 8 listed files have runtime forward references

Removing the import will break import-time class construction for every one of these:

- **`schema_base.py`** — `DataModelAbstractBase` self-referenced in `full_equal`, `data_equal`, `__eq__`, `__ne__`
- **`complex.py`** — `Complex` self-referenced in `data_equal`, `from_oemol`, `from_pdb`
[...]

PR #1298, opened 2026-04-21
Bump Actions to Node.js 24 and remove unused Docker workflow

### LINKED ISSUE #1290: Update all Actions workflows to use more recent Action versions
Observed the following warning in our Actions:

```
2026-03-27T04:11:51.0362277Z ##[warning]Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v3, mamba-org/setup-micromamba@v1. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```

We should update each action to more recent versions where possible.

### LINKED ISSUE #1296: Remove unused Docker container generation Action
We do not, to my knowledge, make use of the generated Docker images using the Linux "everything" environment. We should remove generation of this if there is no need for it.

dotsdl:
## Description

Bundles three low-risk cleanup items from the v0.7.1 milestone.

### 1. Bump Actions past the deprecated Node.js 20 runtime (closes #1290)

GitHub will force Node.js 24 on June 2, 2026. This updates every Action to its latest Node.js 24-compatible major across all 13 workflows:

- `actions/checkout`: v3 → v6
- `mamba-org/setup-micromamba`: v1 → v3
- `codecov/codecov-action`: v3 → v6 — also renames the `file` input to `files` per the v4+ input rename

Left `pierotofy/set-swap-space@master` alone — unmaintained third-party, not called out in the deprecation warning.

### 2. Remove unused Docker ``everything`` container workflow (closes #1296)

`deploy_docker.yaml` is dead: its Dockerfile references `devtools/conda-envs/asapdiscovery-ubuntu-latest.yml`, which doesn't exist after the per-subpackage env split. Removes both the workflow and `devtools/containers/Dockerfile`.

### 3. Fix silently-skipped codecov uploads

While verifying the codecov-action bump, discovered the upload step has been skipped on every run since the repo was transferred from `choderalab/` to `asapdiscovery/`. The `if:` guard on each upload step still checked `github.repository == 'choderalab/asapdiscovery'`, which is now permanently false.

Confirmed across every job on this PR and on recent scheduled runs of `main` — the step shows `conclusion: "skipped"` rather than running. Updating the guard to the current repo name re-enables uploads without changing the original fork-detection behavior. Pre-existing; would have continued silently broken if not caught here.

### Deferred: #1295

#1295 looked mechanical but isn't — all 8 listed files have runtime forward references that `from __future__ import annotations` is currently absorbing, and removing the import raises `NameError` on Python 3.12. Left a [detailed comment on the issue](https://github.com/asapdiscovery/asapdiscovery/issues/1295#issuecomment-4290788288) for later triage (close / defer to 3.14 when PEP 649 lands / full refactor).

## Test plan
- [ ] All 13 updated workflows run to completion on this PR
- [ ] `Upload Code Coverage to Codecov` step now runs (not skipped) and the upload lands on codecov.io
- [ ] No references remain to the deleted Dockerfile


dotsdl (commented):
I'm happy with this. Ready for review @JenkeScheen!

JenkeScheen (approved):
LGTM

Issue #1301, opened 2026-05-22
Remove GufeKey str-coercion workaround in AlchemiscaleFailure once gufe#713 is fixed


dotsdl:
## Summary

`asap-alchemy status --errors` (and `--with-traceback`) currently crashes with a pydantic `ValidationError` when collecting errors from a network. We've landed a client-side workaround; **this issue tracks removing that workaround once the upstream gufe bug is fixed.**

## Root cause

Upstream: [OpenFreeEnergy/gufe#713](https://github.com/OpenFreeEnergy/gufe/issues/713) — `GufeKey`s embedded directly as attributes of a `GufeTokenizable` (e.g. `ProtocolUnitFailure.source_key`) lose their type on a serialization round-trip. After being pulled back from alchemiscale they come back as plain `str`, not `GufeKey` (and lose `.prefix`/`.token`).

Since the pydantic v2 migration (#1281/#1282), `AlchemiscaleFailure.unit_key` and `dag_result_key` are typed as `GufeKey`. `GufeKey` is a bare `str` subclass with no pydantic core schema, so pydantic v2 validates the field with a strict `is_instance_of` check. A plain `str` fails that check:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for AlchemiscaleFailure
unit_key
  Input should be an instance of GufeKey [type=is_instance_of, input_value='RelativeHybridTopologyPr...', input_type=str]
```

The existing unit test never caught this because the `protocol_unit_failures` fixture constructs failures in-memory with `source_key=u.key` (a real `GufeKey`), never exercising the deserialized path.

## Workaround in place

A `mode="before"` field validator on `AlchemiscaleFailure` coerces plain strings back into `GufeKey` (`asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py`), plus a regression test (`test_get_failures_str_source_key` in `test_utils.py`) that drives `collect_errors` with plain-`str` `source_key`s.

## Action when gufe#713 is resolved

- [ ] Bump to the gufe release that fixes #713
- [ ] Remove the `_coerce_gufe_key` validator from `AlchemiscaleFailure`
- [ ] Decide whether to keep `test_get_failures_str_source_key` (recommended, as a deserialization contract guard)

PR #1302, opened 2026-05-22
Fix asap-alchemy status --errors crash on str GufeKey (gufe#713 workaround)

dotsdl:
## Summary

`asap-alchemy status --errors` (and `--with-traceback`) crashes with a pydantic `ValidationError` while collecting task errors:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for AlchemiscaleFailure
unit_key
  Input should be an instance of GufeKey [type=is_instance_of, input_value='RelativeHybridTopologyPr...', input_type=str]
```

**Root cause:** `AlchemiscaleFailure.unit_key`/`dag_result_key` are typed as `GufeKey`. `GufeKey` is a bare `str` subclass with no pydantic core schema, so under pydantic v2 the field is validated with a strict `is_instance_of` check. Keys embedded directly in a `ProtocolUnitFailure` (e.g. `source_key`) lose their type on a serialization round-trip and come back from alchemiscale as plain `str` ([gufe#713](https://github.com/OpenFreeEnergy/gufe/issues/713)) — failing that check.

The existing test never caught this: the `protocol_unit_failures` fixture builds failures in-memory with `source_key=u.key` (a real `GufeKey`), never exercising the deserialized path.

## Changes

- Add a `mode="before"` field validator on `AlchemiscaleFailure` that coerces plain strings back into `GufeKey` (restoring the `.prefix`/`.token` access gufe#713 strips).
- Add `test_get_failures_str_source_key`, a regression test driving `collect_errors` with a plain-`str` `source_key`. Verified it fails without the validator and passes with it.

## Notes

This is a client-side workaround for the still-open gufe#713. Removal is tracked in #1301.

## Test plan

- [x] `asap-alchemy status --errors` runs and prints errored tasks instead of crashing
- [x] `pytest asapdiscovery/alchemy/tests/test_utils.py -k failures` passes (incl. new regression test)
- [x] New test fails when the validator is reverted

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/schema/fec.py
@@ -26,7 +26,7 @@ from openfe.protocols.openmm_utils.omm_settings import (
 )
 from openfe.setup.atom_mapping import lomap_scorers, perses_scorers
 from openff.units import unit as OFFUnit
-from pydantic import ConfigDict, Field
+from pydantic import ConfigDict, Field, field_validator
 
 from ._util import check_ligand_series_uniqueness_and_names
 from .base import _SchemaBase, _SchemaBaseFrozen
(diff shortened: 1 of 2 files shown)


dotsdl (commented):
I'm satisfied with this. Merging!

PR #1304, opened 2026-06-18
Support mixing multiple Protocols in one AlchemicalNetwork

### LINKED ISSUE #1116: Support NEQCycling protocol in ASAP Alchemy

### LINKED ISSUE #1303: Add ability to mix `Transformation`s with differing `Protocol`s using ASAP CLI
We need to refactor `asapdiscovery-alchemy` to support multiple `Protocol`s in a single `AlchemicalNetwork`. Currently it is hardcoded to support only the `RelativeHybridTopopolgyProtocol` found in OpenFE, but we also need to support the `NonEquilibriumCyclingProtocol` featured in `feflow` as well as the `FahNonEquilibriumCyclingProtocol` variant featured in `alchemiscale-fah`.

Data model notes:
- want to be able to adjust the protocols in the YAML file produced by `asap-alchemy prep create -f settings.json` and `asap-alchemy create settings-create.json`
    - want the following fields specifiable:
        - `protocol`: list of protocols, default `["RelativeHybridTopologyProtocol"]`
        - `protocol_strategy`: strategy to use for selecting which protocol is applied to each ligand transformation, default `'all'` (which creates a `Transformation` between every ligand mapping for each included `protocol`)
	- the `FreeEnergyCalculationFactory` should auto-populate default settings for the chosen list of `Protocol`s, likely requiring its own index of available `Protocol`s and which modules they live in
	    - a `protocol_settings` field should be populated with these contents, with its value being a dict keyed by protocol names given in `protocol`, with values the corresponding default `ProtocolSettings` object for that protocol

For gathering, need to be able to separate predictions by protocol:
- the `asap-alchemy predict` command should give outputs separated by protocol in some readable way
- html output gives a table with results and plots
- csv files of predictions, giving DG and DDG in kcal/mol; we don't use these much because we need to report in pIC50
- need to figure out later how we report these results to Postera for medchemist consumption

Need to investigate:
- can `cinnabar` handle MLEs for networks with multiple protocol transformations?

dotsdl:
Closes #1303.

> [!WARNING]
> **Breaking changes for downstream consumers**
> - **Prediction output filenames now always carry a protocol suffix**, e.g. `predictions-absolute-<name>-RelativeHybridTopologyProtocol.csv` / `.html` (and `top_N_posed_ligands-<protocol>.sdf`). The old un-suffixed `predictions-absolute-<name>.csv` is **no longer written**, even for single-protocol runs. Update any tooling that reads these files. (Deliberate — consistent naming across single- and multi-protocol networks.)
> - **Old on-disk network/factory files no longer load.** The flat OpenFE-RFE settings format was removed (no migration shim); loading a pre-multi-protocol file now raises a clear, actionable error telling you to regenerate it. The four `tyk2_*` test fixtures were regenerated and their SHA256s bumped in `test_files.yaml` — **these must be uploaded to the test-files server before CI can pass.**

Refactors `asapdiscovery-alchemy` so a single `AlchemicalNetwork` can mix multiple alchemical `Protocol`s instead of the hardcoded `RelativeHybridTopologyProtocol`. Adds support for `NonEquilibriumCyclingProtocol` (feflow) and `FahNonEquilibriumCyclingProtocol` (alchemiscale-fah).

## What changed
- **`schema/protocols.py` (new):** lazy-import registry of available protocols with helpers to build them, fetch default settings, and reverse-map a protocol object to its name. `feflow`/`alchemiscale-fah` stay optional — a clear `ImportError` fires only when an uninstalled protocol is selected.
- **`schema/fec.py`:** `_FreeEnergyBase` now exposes `protocol: list[str]`, `protocol_strategy: Literal["all"]`, and `protocol_settings: dict[str, Settings]` (native gufe settings, auto-populated from the registry). The flat OpenFE-RFE fields are removed. `to_alchemical_network` builds one `Transformation` per edge **per protocol** (protocol-suffixed names); bespoke force fields and adaptive sampling/padding apply per-protocol only where the settings support it.
- **Results & predictions by protocol:** `TransformationResult` gains a `protocol` tag (recovered in `collect_results` via a gufe-key → protocol map); `to_fe_map_by_protocol()` yields one cinnabar `FEMap` per protocol; `asap-alchemy predict` writes per-protocol CSV/HTML/pose files. Postera upload is skipped-with-warning for multi-protocol networks (deferred per the issue).
- **Serialization:** the heterogeneous `protocol_settings` dict round-trips through gufe `JSON_HANDLER` (a `field_serializer`/`field_validator` pair) so each protocol's exact `Settings` subclass is preserved.

## Notes / decisions
- **ASAP's tuned RFE defaults are preserved** (`openff-2.2.0`, dodecahedron box, gapsys softcore, 1ns/5ns, `protocol_repeats=1`) rather than openfe's raw `default_settings()` — avoids silently reverting `protocol_repeats` to 3.
- **No legacy migration shim:** old flat-format network files no longer load. The four `tyk2_*` test fixtures were regenerated into the new format and their SHA256s bumped in `test_files.yaml`. ⚠️ **The regenerated fixture files still need uploading to the test-files server** for CI.
- Deep cinnabar-MLE-across-mixed-protocols and Postera multi-protocol reporting are deferred, per the issue.

## Testing
Full `asapdiscovery-alchemy` suite green locally (`test_fec_schema`, `test_utils`, `test_protocols`, `test_cli` 29 passed / 5 skipped, plus the lighter modules). Adds registry, multi-protocol build, round-trip key-stability, adaptive-skip-on-NEQ, and multi-protocol result-collection tests. black/isort/flake8 clean.

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
@@ -42,6 +42,63 @@ def create(filename: str):
 
 
 @alchemy.command(
+    help_priority=9,
+    short_help="Convert a legacy (pre-multi-protocol) FreeEnergyCalculationNetwork file to the current schema.",
+)
+@click.option(
+    "-i",
+    "--input",
+    "input_file",
+    required=True,
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    help="The legacy FreeEnergyCalculationNetwork JSON file to convert.",
+)
+@click.option(
+    "-o",
+    "--output",
+    "output_file",
+    required=True,
+    type=click.Path(exists=False, file_okay=True, dir_okay=False, writable=True),
+    help="The path to write the converted (current-schema) network JSON file to.",
+)
+def convert_network(input_file: str, output_file: str):
+    """
+    Convert an old-style (flat-settings, single-protocol) FreeEnergyCalculationNetwork
+    file into the current multi-protocol schema and write it to a new JSON file.
+
+    Args:
+        input_file: The legacy network JSON file to convert.
+        output_file: The path to write the converted network to.
+    """
+    import json
+
+    import rich
+    from gufe.tokenization import JSON_HANDLER
+    from rich import pretty
+    from rich.padding import Padding
+
+    from asapdiscovery.alchemy.cli.utils import print_header
+    from asapdiscovery.alchemy.schema.fec import convert_legacy_fec_network
+
+    pretty.install()
+    console = rich.get_console()
+    print_header(console)
+
+    with open(input_file) as f:
+        data = json.load(f, cls=JSON_HANDLER.decoder)
+
+    network = convert_legacy_fec_network(data)
+    network.to_file(output_file)
+
+    message = Padding(
+        f"Converted legacy network [repr.filename]{input_file}[/repr.filename] to the "
+        f"current schema and saved to [repr.filename]{output_file}[/repr.filename]",
+        (1, 0, 1, 0),
+    )
+    console.print(message)
+
+
+@alchemy.command(
     help_priority=2,
     short_help="Plan a FreeEnergyCalculationNetwork using the given factory and inputs. The planned network will be written to file in a folder named after the dataset.",
 )
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
@@ -939,131 +939,169 @@ def predict(
     )
     console.print(message)
 
-    predict_status = console.status("Calculating absolute free energies")
-    predict_status.start()
-
     # gather all ligands needed for the prediction labels
     ligands = result_network.network.ligands
     if result_network.network.central_ligand is not None:
         ligands.append(result_network.network.central_ligand)
 
-    # convert to cinnabar fepmap to do the prediction via MLE
-    fe_map = result_network.results.to_fe_map()
-    is_connected = cinnabar_femap_is_connected(fe_map)
-
-    if is_connected:
-        try:
-            fe_map.generate_absolute_values()
-        except np.linalg.LinAlgError:
-            raise ValueError(
-                "MLE failed during absolute value generation. Does your result network contain "
-                "NaNs? You can manually remove these or run `predict -c` to remove them automatically."
-            )
-    elif not is_connected and force_largest:
-        fe_map = cinnabar_femap_get_largest_subnetwork(fe_map, result_network, console)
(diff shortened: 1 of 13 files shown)


dotsdl:
Thanks for the thorough review — addressed in 7fd65a0 and 66a7793.

**Blocking**
1. **CI fixtures** — known; the regenerated `tyk2_*` fixtures must be uploaded to the test-files server (the SHA256 bumps are already in `test_files.yaml`). Tracked in the PR description warning; CI will be red until then.
2. **Legacy file failure** — fixed. Added a `model_validator(mode="before")` that detects the old flat-settings keys and raises a clear, actionable error ("created with a pre-multi-protocol version … regenerate with `asap-alchemy create`") instead of an opaque `extra="forbid"` dump. Regression test added. (Still no migration *shim*, per the deliberate design decision.)
3. **Filename suffix break** — keeping the always-suffix behavior (deliberate decision for consistent naming), now **called out loudly** in the PR description under Breaking changes.

**Medium**
4. **Network rebuild in `collect_results`** — kept, with a clarifying comment on *why*: the bespoke-force-field injection feeds into the gufe transformation key, so a faithful key→protocol map requires rebuilding the same transformations that were submitted; the protocol-suffixed name is not available on the result keyed by ScopedKey. It is once-per-gather, not per-result.
5. **Silent tag loss** — fixed; now warns on a key miss.
6. **`key_to_protocol` outside the guard** — fixed; moved inside the `if planned_network:` block.
7. **Non-RFE gather path untested** — found a real latent bug here: `unit_estimate_error` is an OpenFE-RFE-only unit output (feflow NEQ units emit `forward_work`/`reverse_work`/…), so the `uncertainty.m == 0.0` fallback would `KeyError` on an NEQ result. Guarded the key access, and extended `test_collect_results_multi_protocol` to exercise a result shape lacking `unit_estimate_error`.

**Low/nits**
8. `to_fe_map_by_protocol()` is public, tested API; the predict loop needs per-protocol connectivity/force-largest handling before MLE so it builds maps individually — left as-is.
9/10/11. Comments clarified (idempotent double JSON_HANDLER pass; in-place dict mutation targets the frozen Network subclass; `small_molecule_forcefield` single-FF constraint vs bespoke fitting).

Also verified the whole cinnabar surface against the just-released **0.6.0** API — production usage is unchanged; only some positional `.iloc` test lookups needed to become label-based for 0.6.0's new deterministic row sorting (done, passes on both 0.5.0 and 0.6.0).

dotsdl (commented):
I'm happy with this! Ready for review @JenkeScheen!

JenkeScheen:
LGTM!

JenkeScheen (approved):

PR #1313, opened 2026-07-02
Allow `asap-alchemy restart` to take a network ScopedKey

### LINKED ISSUE #1306: For `asap-alchemy restart`, make it possible to take only network `ScopedKey` as input
Currently this endpoint requires a full `planned_network.json`, but this isn't really necessary for this operation. We should make it possible to give a network `ScopedKey` instead.

dotsdl:
Closes #1306.

`asap-alchemy restart` previously required a full `planned_network.json`, but restarting errored Tasks only needs the network's ScopedKey. This mirrors the existing `status`/`gather` dual-input pattern so the network can be identified either way.

## Changes
- Add `-nk/--network_key` to the `restart` command; `-n/--network` is now `required=False`. When both are given, the key is preferred (matching `status`). The key is validated with `network_exists` before restarting.
- Generalize `AlchemiscaleHelper.restart_tasks` to accept a `network_key` directly (in preference to `planned_network`), matching `collect_results`. Backward compatible — existing `planned_network=` callers/tests still work.
- Add `test_alchemy_restart_network_key` exercising the key-only path (no network file needed).

## Usage
```
asap-alchemy restart -nk <network-scoped-key>          # restart all errored tasks
asap-alchemy restart -nk <network-scoped-key> <task-sk>...   # restart specific tasks
asap-alchemy restart -n planned_network.json           # still works
```

## Verification
- `python -m py_compile` passes on all modified files.
- Added CLI test mocks `check_exists`/`get_network_tasks`/`set_tasks_status` and asserts the key-only invocation restarts the errored tasks. (Full test suite requires the conda env; validated by CI.)

### DIFF
--- a/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
+++ b/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py
@@ -722,6 +722,16 @@ def status(
     help="The name of the JSON file containing a planned FEC network.",
     default="planned_network.json",
     show_default=True,
+    required=False,
+)
+@click.option(
+    "-nk",
+    "--network_key",
+    type=click.STRING,
+    help="The network key of the network whose errored Tasks should be restarted. "
+    "Used in preference to --network; find it by running e.g. `asap-alchemy status -a`.",
+    default=None,
+    required=False,
 )
 @click.option(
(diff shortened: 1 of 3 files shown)


dotsdl (commented):
I'm satisfied with this.

PR #1315, opened 2026-07-02
Drop multimethod dependency in favor of isinstance dispatch

### LINKED ISSUE #1288: Drop multimethod dependency in favor of isinstance dispatch
## Summary

The `multimethod` library is used in 4 files across the codebase purely as syntactic sugar for runtime type-based dispatch on `_dispatch` methods. It conflicts with Pydantic v2's `ModelMetaclass`, which tries to reconstruct `multimethod` descriptors via `multimethod.__new__()` without the required `func` argument. This forces every Pydantic model in those files to carry `model_config = ConfigDict(ignored_types=(multimethod,))` as a workaround.

## Proposal

Replace all `@multimethod` usage with simple `isinstance` checks in a single `_dispatch` method, then remove the `multimethod` dependency and all `ignored_types` workarounds.

## Current usage

All usage follows the same pattern — multiple `_dispatch` overloads dispatching on the element type of `inputs`:

| File | Class(es) | Overload types |
|------|-----------|---------------|
| `docking/scorer.py` | `ChemGauss4Scorer`, `FINTScorer`, `MLModelScorer`, `SymClashScorer` | `list[DockingResult]`, `list[Complex]`, `list[Path]`, `list[str]`, `list[Ligand]` |
| `dataviz/html_viz.py` | `HTMLVisualizer` | `list[DockingResult]`, `list[Path]`, `list[Complex]`, `tuple[Complex, list[Ligand]]` |
| `dataviz/gif_viz.py` | `GIFVisualizer` | `list[SimulationResult]`, `list[Path]` |
| `simulation/simulate.py` | `VanillaMDSimulator` | `list[DockingResult]`, `list[Path]` |

## Why drop it

1. **Pydantic v2 incompatibility** — requires `ignored_types` on every affected model (currently 6 model_config entries)
2. **Not necessary** — all dispatch is on the first element's type, trivially replaceable with `isinstance`
3. **Reduces dependency surface** — one fewer third-party package to maintain compatibility with
4. **Fragile** — any new Pydantic model that forgets `ignored_types` will silently break

dotsdl:
Closes #1288.

The `multimethod` library was used in four files purely as syntactic sugar for runtime type-based dispatch on `_dispatch` methods. It conflicts with Pydantic v2's `ModelMetaclass` (which tries to reconstruct `multimethod` descriptors via `multimethod.__new__()` without the required `func`), forcing every affected model to carry `model_config = ConfigDict(ignored_types=(multimethod,))`. Any new model that forgot it would silently break.

All dispatch is on the first element's type, so it is trivially replaceable with `isinstance`.

## Changes
- Replace each `@multimethod` / `@_dispatch.register` overload set with a single `isinstance`-dispatching `_dispatch` method (with an explicit `NotImplementedError` for unknown types and an empty-input short-circuit):
  - `docking/scorer.py` — `ChemGauss4Scorer`, `FINTScorer`, `GATScorer`, `E3MLModelScorer`, `SymClashScorer`
  - `dataviz/html_viz.py` — `HTMLVisualizer`
  - `dataviz/gif_viz.py` — `GIFVisualizer`
  - `simulation/simulate.py` — `VanillaMDSimulator`
- Remove the six `ignored_types=(multimethod,)` workarounds, and drop now-unused `ConfigDict` / `multimethod` / `Ligand` imports.
- Drop the `multimethod` conda dependency from all env files and `docs/requirements.yaml`.

The dispatch behavior is preserved 1:1 — each branch keeps the original body, and the re-dispatch calls (e.g. `Path` → `Complex`) still work since they route through the same `isinstance` method.

## Verification
- `python -m py_compile` passes on all four modules.
- AST check: each scorer class now has exactly one, undecorated `_dispatch`.
- Repo-wide `grep` confirms **no** remaining `multimethod` references (code, env, or docs).
- No tests referenced `multimethod` / `DispatchError` / `_dispatch` directly.
- (Full test suite requires the conda env; validated by CI.)

### DIFF
--- a/asapdiscovery-dataviz/asapdiscovery/dataviz/gif_viz.py
+++ b/asapdiscovery-dataviz/asapdiscovery/dataviz/gif_viz.py
@@ -3,7 +3,6 @@ import shutil
 from pathlib import Path
 from typing import Any, Optional, Union
 
-from multimethod import multimethod
 from pydantic import ConfigDict, Field, PositiveInt
 
 from asapdiscovery.data.metadata.resources import master_structures
@@ -94,9 +93,7 @@ class GIFVisualizer(VisualizerBase):
     interval: PositiveInt = Field(1, description="Interval between frames")
     debug: bool = Field(False, description="Whether to run in debug mode")
 
-    model_config = ConfigDict(
-        arbitrary_types_allowed=True, ignored_types=(multimethod,)
-    )
+    model_config = ConfigDict(arbitrary_types_allowed=True)
 
     @dask_vmap(["inputs"], has_failure_mode=True)
     @backend_wrapper("inputs")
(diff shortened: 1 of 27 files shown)


dotsdl (commented):
I'm satisfied with this.

